PR #1, opened 2018-12-03
Add a minimal README

smarterclayton:


smarterclayton:
@derekwaynecarr look ma, no openshift

PR #2, opened 2018-12-19
Add ability to capture request headers & cookies

rsevilla87:
HAProxy allows to capture request/response headers and cookies, this PR make users able to capture these headers or cookies and print them in a log.
Also added strings.Split to template helpers functions that could be useful in a future.

more info: https://github.com/openshift/origin/pull/19704


vladimircervjakov:
Any update on this?

PR #8, opened 2019-01-22
Comply with RFC7239 for Forwarded header

alexcern:
`proto-version` may be empty when http/2 is not enabled (the default), and empty values
must be surrounded by quotes as per https://tools.ietf.org/html/rfc7239

Applications may fail to parse the header if not complying with the RFC


alexcern:
@ramr could you please take a look? Thanks!

ironcladlou:
/lgtm
/approve

Sorry for the delay!

PR #9, opened 2019-01-23
Fix broken update handling of path-based routes

ironcladlou:
Fix broken update handling of path-based routes

When an existing route's path is updated to a new value not previously observed,
the new route state is correctly activated, but the previous route is not
replaced in the `hostRules` internal active list. Instead, the updated route
state (with a different path but duplicate UID) is appended to the active list.
A `hostRules` active list with more than one route instance sharing a UID and
host is an invalid internal state. The net effect is some unintended confusing
behavior which manifests when updating the route's path, including:

  * Updating from path A->B->A preventing future transitions to B
  
  * Updating from path A->B->A->B leaves an orphaned claim on the host in the
namespace, preventing the recreation of the route with either of the previously
observed paths until the router is restarted (causing the internal state to be
rebuilt)

Routes updates which are affected by this bug will be rejected with a status
like:

    message: route foo already exposes foo-ns.os.example.com and is older
    reason: HostAlreadyClaimed

Where `foo` refers to itself. This indicates the route update was rejected due
to a claim made by the same route being updated.

To fix the problem, ensure that when a path change is detected on the same route
the prior existing state is removed from the index, eliminating the possibility
of keeping an orphaned claim. Then add the new route as usual, which handles
acceptance/rejection normally.

### DIFF
--- a/pkg/router/controller/hostindex/hostindex.go
+++ b/pkg/router/controller/hostindex/hostindex.go
@@ -101,7 +101,10 @@ func (hi *hostIndex) add(route *routev1.Route, changes *routeChanges) bool {
 			hi.remove(existing, false, changes)
 			// uid changed, which means this is
 		case existing.Spec.Path != route.Spec.Path:
-			// path changed, must check to see if we displace / are displaced by another route
+			// path changed, must check to see if we displace / are displaced by
+			// another route. Remove the existing state to avoid maintaining a
+			// duplicate claim in the index.
+			hi.remove(existing, false, changes)
 		default:
 			// if no changes have been made, we don't need to note a change
 			if existing.ResourceVersion == route.ResourceVersion {
--- a/pkg/router/controller/hostindex/hostindex_test.go
+++ b/pkg/router/controller/hostindex/hostindex_test.go
@@ -233,6 +233,64 @@ func Test_hostIndex(t *testing.T) {
 			},
 			active: map[string][]string{"test.com": {"001"}},
 		},
+		{
+			name:       "multiple changes to same path-based route",
+			activateFn: SameNamespace,
+			steps: []step{
+				{route: newRoute("test", "1", 1, 1, routev1.RouteSpec{Host: "test.com", Path: "/foo"})},
+				{route: newRoute("test", "1", 1, 2, routev1.RouteSpec{Host: "test.com", Path: "/bar"})},
+				{route: newRoute("test", "1", 1, 3, routev1.RouteSpec{Host: "test.com", Path: "/foo"})},
+				{route: newRoute("test", "1", 1, 4, routev1.RouteSpec{Host: "test.com", Path: "/bar"})},
+				{route: newRoute("test", "1", 1, 5, routev1.RouteSpec{Host: "test.com", Path: "/foo"})},
+			},
+			active:    map[string][]string{"test.com": {"001"}},
+			activates: map[string]struct{}{"001": {}},
+		},
+		{
+			name:       "remove unchanged path-based route",
+			activateFn: SameNamespace,
+			steps: []step{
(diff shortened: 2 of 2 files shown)


pkg/router/controller/hostindex/hostindex.go
```diff
@@ -265,7 +265,26 @@ func (r *hostRules) add(route *routev1.Route, fn RouteActivationFunc, changes *r
...
+	// should cause an update to the existing route in the plugin without
+	// necessitating a delete/create.
+	filtered := active[:0]
+	for _, existing := range active {
+		if existing.UID == route.UID && existing.ResourceVersion != route.ResourceVersion {
+			continue
+		}
+		filtered = append(filtered, existing)
+	}
+	r.active = filtered
```
ramr: Do you think it might be simpler to just remove the route inside `HostIndex.[aA]dd` if the paths don't match and let the add  proceed as if it was a new route?  

Aka: in here: 
https://github.com/openshift/router/blob/2b0286e8d005682382ef14f343e0903911ca7039/pkg/router/controller/hostindex/hostindex.go#L104
add something like:  
```
      hi.remove(existing, false, changes)
      newRoute = true
```

That way, if it gets re-activated back it would just be an underlying plugin route added but if it gets displaced because there is some other route holding that path, it would be a displaced route and its status would be updated to rejected as some other route holding that "host+path". 

And that would be a good test case to add: 
1. Route1 is the oldest route and holds host and path ==  www.abc.test/x/y/z  
2. Route2 is newer than Route1 and originally set to host+path == www.abc.test/foo
3. Route2 gets modified to host+path === www.abc.test/bar 
    which should let it be re-activated and admitted back in with the new/updated path.
4. Route2 gets modified to host+path == www.abc.test/x/y/z 
     which should cause Route2 to be rejected because Route1 holds that host+path.

Would that change make more sense?

Miciah: > Do you think it might be simpler to just remove the route inside `HostIndex.[aA]dd` if the paths don't match and let the add proceed as if it was a new route?

I think this is correct.  The `hi.remove` code path would remove the existing route from the inactive and active lists before the activation function ran, so the activation function would not call `changed.Displaced` on the existing route but would activate the oldest route that the existing route had been displacing.  Then the `rules.add` would re-activate the existing route iff it did not conflict with an older route.

Miciah: This is analogous to the check for host changed:
```go
	// if the host value changed, remove the old entry
	oldHost, ok := hi.routeToHost[key]
	if ok && oldHost != host {
		if existing, _, _, ok := hi.findRoute(oldHost, key); ok {
			hi.remove(existing, true, changes)
			newRoute = false
		}
	}
```
Of course we can only check for path changed after we have gotten the route using `findRoute`.  However, the cases are similar, so if using `hi.remove` and then `rules.add` is correct for the host-changed case, it makes sense to use it for the path-changed case as well.

ironcladlou: Great ideas, I think this works. One thing I think was missing from my mental model here is that by the time we're handling this route update, the new state is already persisted to storage, so it's safe to remove the old state immediately (because if the new state is rejected, we're not going to revert the persisted state). Awesome!

ironcladlou: @ramr @Miciah refactored and added new test, PTAL.

ironcladlou:
@smarterclayton PTAL, you gotta tag this one for now

pkg/router/controller/hostindex/hostindex_test.go
```diff
@@ -233,6 +233,64 @@ func Test_hostIndex(t *testing.T) {
 			},
 			active: map[string][]string{"test.com": {"001"}},
 		},
+		{
```
smarterclayton: nice catch

PR #10, opened 2019-02-05
Refactor haproxy router config manager blueprint validation code.

ramr:
Follow up to refactor the blueprint validation code.

Migrated from: https://github.com/openshift/origin/pull/20722

### DIFF
--- a/pkg/router/template/configmanager/haproxy/blueprint_plugin.go
+++ b/pkg/router/template/configmanager/haproxy/blueprint_plugin.go
@@ -24,7 +24,7 @@ func NewBlueprintPlugin(cm templaterouter.ConfigManager) *BlueprintPlugin {
 func (p *BlueprintPlugin) HandleRoute(eventType watch.EventType, route *routev1.Route) error {
 	switch eventType {
 	case watch.Added, watch.Modified:
-		p.manager.AddBlueprint(route)
+		return p.manager.AddBlueprint(route)
 	case watch.Deleted:
 		p.manager.RemoveBlueprint(route)
 	}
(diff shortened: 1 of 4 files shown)


Miciah:
What's up with the return value of `AddBlueprint`? Was that change supposed to be in this PR? Should `(p *BlueprintPlugin) HandleRoute` be changed to handle the error return?

ramr:
> What's up with the return value of `AddBlueprint`? Was that change supposed to be in this PR? 
> Should `(p *BlueprintPlugin) HandleRoute` be changed to handle the error return?
Missed that - good catch. Yeah, `HandleRoute` should propogate that error up.  So the old pr went thru' some iterations where it was generating events in here and then we dropped that, so the error handling needed to propogate errors back (on blueprint add).

Miciah:
With `(p *BlueprintPlugin) HandleRoute` propagating the error from `AddBlueprint`, `(c *RouterController) processRoute` will log it, right? Do we still need the `glog.Errorf` in `AddBlueprint`?

ramr:
> With `(p *BlueprintPlugin) HandleRoute` propagating the error from `AddBlueprint`, `(c *RouterController) processRoute` will log it, right? Do we still need the `glog.Errorf` in `AddBlueprint`?

I think it should - will check that next week and if so, remove it. Thx

ramr:
@ironcladlou  PTAL needs approval. Thx

knobunc:
@smarterclayton Do you know why this PR hates @Miciah and @ironcladlou?

/lgtm

ramr:
/retest

error was: `level=fatal msg="failed to fetch Terraform Variables: failed to fetch dependency of \"Terraform Variables\": failed to fetch dependency of \"Cluster ID\": failed to fetch dependency of \"Install Config\": failed to fetch dependency of \"Base Domain\": failed to generate asset \"Platform\": EOF"
�[0G�[2K�[1;92m? �[1;99mPlatform  [Use arrows to move, type to filter]`

PR #14, opened 2019-02-20
Fix issue where wildcard routes get 503s intermittently.

ramr:
Pull from origin PR: https://github.com/openshift/origin/pull/22087

As part of the resync processing, wildcard routes get a "HostAlreadyClaimed"
error and then later as part of the same event processing flow, the route
gets re-added. This results in a temporary outage (route returns 503s).

fixes #1660647 (https://bugzilla.redhat.com/show_bug.cgi?id=1660647)

@openshift/sig-network-edge  PTAL Thx


ramr:
@ironcladlou  needs approval. Thx

pweil-:
/lgtm
/approve

Approving based on slack convo with @ironcladlou

PR #17, opened 2019-02-27
owners: clean up OWNERS

ironcladlou:
Remove unnecessary sub-OWNERS files and revise the root file.


ironcladlou:
@smarterclayton @knobunc something is still up with OWNERS (https://github.com/openshift/router/pull/10), while we figure it out I wanted to see if we could do some cleanup (I don't know we need sub-OWNER granularity here at the moment)

PR #19, opened 2019-03-13
WIP: Support subdomain field when host is unset and we have a canonical host

smarterclayton:
The subdomain is a hint to a router to default the route name. It may be
ignored if the router is not configured in that mode.


pkg/cmd/infra/router/router.go
```diff
@@ -94,6 +94,10 @@ func (o *RouterSelection) Bind(flag *pflag.FlagSet) {
 
 // RouteUpdate updates the route before it is seen by the cache.
 func (o *RouterSelection) RouteUpdate(route *routev1.Route) {
+	// if a subdomain default is observed and we have a canonical hostname, set the host field
+	if len(route.Spec.Subdomain) > 0 && len(route.Spec.Host) == 0 && len(o.RouterCanonicalHostname) {
```
ironcladlou: `len()` missing a comparison — does this compile?

PR #25, opened 2019-05-07
Bug 1701410: Rewrite default certificate when secret is updated

Miciah:
* `pkg/router/template/router.go` (`watchSecretDir`): New method.  Establish a watch on the specified path, and rewrite the default certificate when the watch detects changes.
(`writeDefaultCert`): Call `watchSecretDir` to watch the directory with the default certificate secret.

---

@openshift/sig-network-edge

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -17,6 +17,7 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/golang/glog"
 	"github.com/prometheus/client_golang/prometheus"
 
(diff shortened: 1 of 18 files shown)


pkg/router/template/router.go
```diff
@@ -276,6 +277,79 @@ func secretToPem(secPath, outName string) error {
 	return ioutil.WriteFile(outName, pemBlock, 0444)
 }
 
+// watchSecretDir adds a watcher on the input directory that updates the output
+// PEM file when a change is detected.
+func (r *templateRouter) watchSecretDir(secPath, outName string) error {
```
ironcladlou: Throughout: instead of logs in the main loop path, how about capturing metrics?

pkg/router/template/router.go
```diff
@@ -276,6 +277,79 @@ func secretToPem(secPath, outName string) error {
...
+				currentRealPath = newRealPath
+
+				glog.Infof("Got watch event %v; updating %q", event, outName)
+				os.Remove(outName)
+				if err := secretToPem(secPath, outName); err != nil {
+					glog.Warningf("Failed to update %q: %v", outName, err)
+				}
+
+				r.stateChanged = true
+				r.dynamicallyConfigured = false
```
ironcladlou: Is this safe to do without acquiring `lock`?

Miciah: Probably not.  Fixed.  Thanks!

pkg/router/template/router.go
Miciah: Do you want a new metric in addition to the `metricReload` and `metricWriteConfig` metrics?

ironcladlou: Seems like having more metrics is generally useful. Is there a disadvantage to adding more metric types here?

ironcladlou:
Is there a way to test this in e2e? Would observing metrics counts be useful for test assertions?

Miciah:
> Is there a way to test this in e2e? Would observing metrics counts be useful for test assertions?

I wrote out the steps that I used to test this change here: https://bugzilla.redhat.com/show_bug.cgi?id=1701410#c8.  We could translate the `openssl` and `oc` commands into Go.  I don't think metrics would be particularly useful—I think it's easier to send a request and see what certificate is returned than it is to check a metric.

pkg/router/template/router.go
Miciah: It seems strange to have a metric for this specific case.  Do we want a metric for each reload trigger?

ironcladlou: For a basis of comparison, what questions do the existing logging here answer, and who's asking them?

A generic benefit of metrics is they doesn't presuppose the questions — metrics provide the ability to formulate ad-hoc lines of questioning.

Here are some questions:

1. How often are we reloading the router over <timespan>
2. How often are we refreshing the certificate secret over <timespan>
3. Are we refreshing the certificate abnormally (imagine an fsnotify issue — notice the lack of rate limiting in this code)

More metrics points also enable new alerts to be imagined. For example, we could even automate question 3 by defining an alert to trigger when the secret refreshes more than twice a minute.

Events would also have advantages over logs here, but I'm not sure if they're worth it?

Miciah: > For a basis of comparison, what questions do the existing logging here answer, and who's asking them?

"Did the router notice that I updated the default certificate secret?", or, "Why isn't the router using the default certificate that I just configured?".

> A generic benefit of metrics is they doesn't presuppose the questions — metrics provide the ability to formulate ad-hoc lines of questioning.
> 
> Here are some questions:
> 
> 1. How often are we reloading the router over _timespan_
> 2. How often are we refreshing the certificate secret over _timespan_
> 3. Are we refreshing the certificate abnormally (imagine an fsnotify issue — notice the lack of rate limiting in this code)
> 
> More metrics points also enable new alerts to be imagined. For example, we could even automate question 3 by defining an alert to trigger when the secret refreshes more than twice a minute.

I think the two interesting case are (1) the secret was changed or (2) the watch broke, for the following scenarios:

* The administrator changed the default certificate and expects feedback that the change has been recognized—case (1).
* The administrator did not change the default certificate and should be alerted if a change has been recognized—absence of case (1).
* The watch broke—case (2)—which would could indicate that the platform is unstable.

> Events would also have advantages over logs here, but I'm not sure if they're worth it?

An event seems appropriate for (1) above.  Can one configure an alert for an event?

For case (2), it almost seems reasonable to crashloop... But maybe we could emit an event and recreate the watch.

ironcladlou:
@Miciah is this ready to go assuming the intent is to implement the minimum useful behavior? The metrics and other stuff can be considerations for future improvements.

Miciah:
> is this ready to go assuming the intent is to implement the minimum useful behavior? The metrics and other stuff can be considerations for future improvements.

Yeah, using metrics (or some alternative, such as events) is the only outstanding work.

PR #28, opened 2019-05-17
Add shell script for graceful shutdown

Miciah:
Add a `shutdown-haproxy` shell script, which can be used in as the pre-stop handler for a router deployment in order to provide graceful shutdown.

Whereas a hard shutdown simply sends a SIGTERM signal to all haproxy processes to make them immediately shut down, a graceful shutdown does the following:

1. Create a marker file that tells the `reload-haproxy` script to exit immediately.

2. Sleep for 30 seconds to allow other proxies such as kube-proxy and any cloud load-balancers to drain connections.

3. Send a SIGUSR1 signal to all HAProxy processes to make them stop accepting new connections.

4. Wait up to 30 seconds for all HAProxy processes to terminate.

5. Perform a hard shutdown of any remaining HAProxy processes.

This commit is related to bug 1709958.

https://bugzilla.redhat.com/show_bug.cgi?id=1709958

* `images/router/haproxy/reload-haproxy`: Exit immediately if the file `/var/lib/haproxy/run/terminating` is present.
* `images/router/haproxy/shutdown-haproxy`: New file.  If the file `/var/lib/haproxy/run/terminating` is present, perform a hard shutdown, or else perform a graceful shutdown as described above.

### DIFF
--- a/images/router/haproxy/shutdown-haproxy
+++ b/images/router/haproxy/shutdown-haproxy
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# DRAIN_MARKER is the path to a file that indicates that proxies or load
+# balancers should be draining connections to the router.  The router's health
+# check for load balancers checks for this file, and if the file is present, the
+# health check fails in order to cause load balancers to drain connections to
+# the router.
+: ${DRAIN_MARKER:='/var/lib/haproxy/run/draining'}
+
+# DRAIN_PERIOD is the number of seconds that this script will wait for
+# connections to drain before it sends the SIGUSR1 signal to terminate HAProxy
+# processes.
+: ${DRAIN_PERIOD:=90}
+
+# TERMINATE_MARKER is the path to a file that indicates that this router is
+# terminating.  The reload-haproxy script exits immediately if this marker is
+# present.
+: ${TERMINATE_MARKER:='/var/lib/haproxy/run/terminating'}
+
+# GRACE_PERIOD is the number of seconds that this script will wait for HAProxy
+# processes to terminate after it sends the SIGUSR1 signal, before it sends
+# SIGTERM to any remaining HAProxy processes.
+: ${GRACE_PERIOD:=30}
+
+echo " - Setting drain marker..."
+: > "$DRAIN_MARKER"
+
+echo " - Sleeping $DRAIN_PERIOD seconds to let connections drain..."
+sleep "$DRAIN_PERIOD"
+
+echo " - Sending SIGUSR1 to HAProxy processes and waiting up to $GRACE_PERIOD seconds for processes to terminate..."
+stop=$((SECONDS + GRACE_PERIOD))
+while pkill -USR1 haproxy; rc=$?; [[ $rc -ne 1 ]] && ((SECONDS < stop)); do
+    sleep 1
+done
+
+if [[ "$rc" -eq 1 ]]; then
+    echo ' - Done.  All processes have exited.'
+    exit 0
+fi
+
+: > "$TERMINATE_MARKER"
+echo ' - Sending SIGTERM to HAProxy processes...'
+while pkill -TERM haproxy; [[ $? -ne 1 ]]; do
+    sleep 1
+done
+echo ' - Done.  All processes have been terminated.'
--- a/images/router/haproxy/reload-haproxy
+++ b/images/router/haproxy/reload-haproxy
@@ -66,6 +66,10 @@ function haproxyHealthCheck() {
 }
 
 
+if [[ -e ${TERMINATE_MARKER:-'/var/lib/haproxy/run/terminating'} ]]; then
+  exit 0
+fi
+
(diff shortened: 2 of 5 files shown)


lihongan:
If long-lived connection like websocket exists in current HAProxy process, how to handle it? Wait 30s then perform a hard shutdown? But that sounds not a graceful shutdown. 
```
  # Long timeout for WebSocket connections.
  timeout tunnel 1h
```

images/router/haproxy/shutdown-haproxy
```diff
@@ -0,0 +1,48 @@
...
+graceful_shutdown='false'
+if ! [[ -e "$TERMINATION_MARKER" ]]; then
+    : > "$TERMINATION_MARKER"
+    graceful_shutdown='true'
+fi
+
+if [[ "$graceful_shutdown" = 'true' ]]; then
+    echo 'Performing a graceful shutdown:'
+    echo " - Sleeping $DRAIN_PERIOD seconds to let connections drain..."
+    sleep "$DRAIN_PERIOD"
```
ironcladlou: Granted the reload script will be no-op'ing, if we haven't issued USR1 to haproxy it's not clear to me how connection draining is implemented with this sleep.

images/router/haproxy/shutdown-haproxy
```diff
@@ -0,0 +1,48 @@
...
+TERMINATION_MARKER='/var/lib/haproxy/run/terminating'
+
+# DRAIN_PERIOD is the number of seconds that this script will wait for
+# connections to drain before it sends the SIGUSR1 signal to HAProxy processes.
+DRAIN_PERIOD=30
+
+# GRACE_PERIOD is the number of seconds that this script will wait for HAProxy
+# processes to terminate after it sends the SIGUSR1 signal, before it sends
+# SIGTERM to any remaining HAProxy processes.
+GRACE_PERIOD=30
```
ironcladlou: Should GRACE_PERIOD and DRAIN_PERIOD be configurable?

Miciah: They easily could be, but I wasn't planning to add any API for these parameters to the ingress controller resource, so I don't know that it would be much use to make them configurable here.

ironcladlou: The ingress controller is necessarily aware of how to configure the haproxy-router, I'm not sure that implies we have to expose this as public API. I was just suggesting that if we ever need to tweak these values, we might want to do it through an operator release rather than an haproxy-router release. Especially if we end up tweaking these values dynamically in the operator...

ironcladlou: The stop hook script might even be possible to define in the operator itself rather than here, but that could be a little _too_ coupled?

Miciah: > I was just suggesting that if we ever need to tweak these values, we might want to do it through an operator release rather than an haproxy-router release. Especially if we end up tweaking these values dynamically in the operator...

Hm, that's a good point—the operator might know that one cloud provider needs a shorter or longer drain period than the typical cloud provider, in which case the operator could set a platform-appropriate value.

> The stop hook script might even be possible to define in the operator itself rather than here, but that could be a little _too_ coupled?

You mean with a configmap volume? We could do that, but (1) the reload and shutdown scripts are loosely coupled by the `/var/lib/haproxy/run/terminating` marker, and (2) it would be another piece that could break, so it adds real operational complexity to reduce hypothetical build complexity.

Miciah: Latest push makes `GRACE_PERIOD` and `DRAIN_PERIOD` configurable.

smarterclayton:
What’s the status of this?

Miciah:
> What’s the status of this?

It does not work as is.  As soon as the pod is deleted, the pod's endpoints are deleted, even if the pod has a pre-stop lifecycle handler.  The service proxy observes that the node has no local endpoints for the LB service, and it starts blackholing traffic to the node.

To make draining work, we must initiate it _before_ the pod is deleted.  To do this, we need to make the load balancer's health check start failing before deleting the pod.  Right now, the health check for LB services is serviced by the service proxy, so we would need to make the health check configurable (proof of concept: https://github.com/Miciah/origin/commit/108a036a24007b4a40bab16ead5d7101c044c333).

Alternatively, we have explored using surge during rollout, using affinity rules to schedule new replicas on the same nodes as old replicas to ensure that local endpoints never disappear during a rollout (proof of concept: https://github.com/Miciah/cluster-ingress-operator/commit/93517fd4c9a1d9abde97378b98928bad73475614).  This approach has other challenges, the most salient being that we need specific behavior when the old replicas are descheduled, namely to prefer descheduling old replicas on nodes where the new replicas are ready.  (Someone suggested a change along these lines here: https://github.com/kubernetes/kubernetes/issues/4301#issuecomment-73812243.)  My next step is to create a proof-of-concept change to the replicaset controller to give us the desired behavior.

Miciah:
Superseded by openshift/cluster-ingress-operator#280.
/close

PR #34, opened 2019-08-12
Add ability to customize global queue timeout

yann-soubeyrand:
Maybe the queue timeout should also be customizable from route annotation.


yann-soubeyrand:
Hi @ironcladlou, do you have any comment on this?

yann-soubeyrand:
ping @ironcladlou

PR #40, opened 2019-10-03
Update to go 1.12 and kube 1.16

ironcladlou:
Need I say more?


cmd/openshift-router/main.go
```diff
@@ -10,16 +10,14 @@ import (
...
-	"k8s.io/apiserver/pkg/util/logs"
 	"k8s.io/client-go/pkg/version"
 
 	"github.com/openshift/library-go/pkg/serviceability"
+
 	"github.com/openshift/router/pkg/cmd/infra/router"
 )
 
 func main() {
-	logs.InitLogs()
```
frobware: How come we don't need this (where "this" is logging) anymore?

Miciah: Good point—should we also be switching glog out for klog?

ironcladlou: I found the moved package an restored it. https://github.com/kubernetes/kubernetes/commit/ba81a5409affb6b861d5994c1e2f7c74a05826e5

I don't want to re-do log handling in this PR unless absolutely necessary

frobware: > Good point—should we also be switching glog out for klog?

I would say so. Looks like everything that is in vendor is using klog.

frobware:
The unit test are presumably failing because CI is pegged to Go 1.10 - looks like the TLS symbol requires Go 1.12:


- https://github.com/openshift/release/blob/master/ci-operator/config/openshift/router/openshift-router-master.yaml#L10
- https://github.com/openshift/release/blob/master/ci-operator/config/openshift/router/openshift-router-release-4.3.yaml#L10

cmd/openshift-router/main.go
frobware: > I don't want to re-do log handling in this PR unless absolutely necessary

Agreed. Should be a separate PR.

ironcladlou:
Depends on https://github.com/openshift/release/pull/5289

ironcladlou:
Looks like the log init needs to go away after all, as some other package is already calling it.

/usr/bin/openshift-router flag redefined: log_dir
panic: /usr/bin/openshift-router flag redefined: log_dir

```
goroutine 1 [running]:
flag.(*FlagSet).Var(0xc0000c0120, 0x192a3a0, 0x26624d0, 0x1695d41, 0x7, 0x16c9296, 0x2f)
	/usr/local/go/src/flag/flag.go:850 +0x4af
flag.(*FlagSet).StringVar(...)
	/usr/local/go/src/flag/flag.go:753
k8s.io/klog.InitFlags(0xc0000c0120)
	/go/src/github.com/openshift/router/vendor/k8s.io/klog/klog.go:428 +0x9d
k8s.io/component-base/logs.init.0()
	/go/src/github.com/openshift/router/vendor/k8s.io/component-base/logs/logs.go:35 +0x2d
```

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_router/40/pull-ci-openshift-router-master-e2e-aws/118/artifacts/e2e-aws/pods/openshift-ingress_router-default-5454cfd6bb-ssxxs_router.log

We can untangle logging in another PR. This code predates several iterations of log library migrations.

ironcladlou:
Okay, I've done some more local testing and logging is screwed up enough that I need to fix it here after all.

ironcladlou:
Various infra failures.

/retest

ironcladlou:
@openshift/sig-network-edge okay, I think I fixed logging "enough" to move forward — let's see the latest e2e runs to make sure things are being logged (even if we get some silly flag errors mixed amongst the output).

My goal is to get this working enough that we can do a followup to rip out glog entirely.

frobware:
If we know what the logging issues are..., then:

/lgtm

PR #45, opened 2019-10-07
Bug 1758374: fix haproxy reload crash when processing ECDSA keys

ironcladlou:
With extended validation enabled, if the router detects an ECDSA private key when processing TLS data,
the sanitized key output is rendered with PEM block headers like:

    -----BEGIN ECDSA PARAMETERS-----
    -----END ECDSA PARAMETERS-----

However, OpenSSL recognizes the following headers for ECDSA keys[1]:

    -----BEGIN EC PARAMETERS-----
    -----END EC PARAMETERS-----

Consequently, when the router provides haproxy with such PEM files through
`crt-list` option, haproxy (which uses OpenSSL) fails to load the key and reload
fails.

The impact is that a Route resource containing an ECDSA key in `.spec.tls.key`
will cause haproxy reloads to crash loop, preventing any further router
processing until the problematic Route is deleted.

This patch ensures that sanitized ECDSA key output follows the OpenSSL
conventions which indirectly apply to haproxy.

[1] https://github.com/openssl/openssl/blob/master/include/openssl/pem.h#L54

### DIFF
--- a/pkg/router/routeapihelpers/validation.go
+++ b/pkg/router/routeapihelpers/validation.go
@@ -62,7 +62,7 @@ func privateKeyBlockVerifier(block *pem.Block) (*pem.Block, error) {
 		}
 	case *ecdsa.PrivateKey:
 		block = &pem.Block{
-			Type: "ECDSA PRIVATE KEY",
+			Type: "EC PRIVATE KEY",
 		}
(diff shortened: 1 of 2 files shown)


crawford:
/retitle Bug 1758374: fix haproxy reload crash when processing ECDSA keys

PR #49, opened 2019-10-08
Add unified Dockerfiles

ironcladlou:
Add unified Dockerfiles. The base image is no longer required since the
v4 router is only supported with the haproxy template. With this commit,
the builds can be refactored to point at the new Dockerfiles, and then
in a followup the images directory can be freely refactored and pruned.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
+WORKDIR /go/src/github.com/openshift/router
+COPY . .
+RUN make
+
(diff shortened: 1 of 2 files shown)


ironcladlou:
There's kind of a chicken and egg problem with testing this one. After merge, I can open a PR to release which will exercise the files in a rehearse job, and that would be the real test. Can do any necessary followups in response to the rehearse outcome.

ironcladlou:
And, since the builds don't know about these files now, committing them should be harmless.

ironcladlou:
@Miciah PTAL

PR #50, opened 2019-10-08
Delete tracking of peer endpoints

Miciah:
Before this commit, the router used the endpoints of a "peer" service to discover peer router instances and made a list of these peer endpoints available to the configuration template.  However, the default template does not use peer endpoints, and the feature is not documented.

The feature is left over from an early implementation of sticky sessions (https://github.com/openshift/origin/pull/2105/commits/4c52fd188166cb5788f0bd249385b49cd5d9a7ff), which used peer endpoints to configure peers in `haproxy`.config.  However, the implementation of sticky sessions that was merged did not use peers, and no other users exist.

* `pkg/cmd/infra/router/template.go` (`Complete`, `Run`): Delete the router peer service from the template-router options.
* `pkg/router/template/plugin.go` (`TemplatePluginConfig`, `NewTemplatePlugin`): Delete logic to get the peer endpoints key from the peer service.
(`peerEndpointsKey`): Delete function.
* `pkg/router/template/router.go` (`templateRouter`, `templateRouterCfg`, `newTemplateRouter`): Delete the peer endpoints and peer endpoints key from the template router state.
(`templateData`, `writeConfig`): Delete peer endpoints from the template data.
(`DeleteEndpoints`, `AddEndpoints`): Delete logic to update the list of peer endpoints.

### DIFF
--- a/pkg/cmd/infra/router/template.go
+++ b/pkg/cmd/infra/router/template.go
@@ -17,7 +17,6 @@ import (
 	"github.com/spf13/pflag"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	ktypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/authentication/authenticatorfactory"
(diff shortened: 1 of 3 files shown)


ironcladlou:
Nice, thanks!

/lgtm

PR #52, opened 2019-10-12
Bug 1820400: Watch default certificate secret even if initial secret is invalid

miheer:
Set up a watch on the default certificate secret even if it does not contain a valid certificate when openshift-router starts.

https://jira.coreos.com/browse/NE-228

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -326,6 +326,7 @@ func (r *templateRouter) watchSecretDir(secPath, outName string) error {
 				r.lock.Lock()
 				r.stateChanged = true
 				r.dynamicallyConfigured = false
+				r.defaultCertificatePath = outName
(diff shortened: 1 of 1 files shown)


ironcladlou:
@Miciah This sounds like a bug rather than a story in Jira, WDYT? If so I would prefer to open a bz so we can track back ports and such if required.

Miciah:
> This sounds like a bug rather than a story in Jira, WDYT? If so I would prefer to open a bz so we can track back ports and such if required.

I found the deficiency myself, and no one has complained let alone requested a backport, but I don't object to opening a Bugzilla report to track it as a defect.

miheer:
@Miciah @ironcladlou filed https://bugzilla.redhat.com/show_bug.cgi?id=1820400

PR #53, opened 2019-10-15
Stop base64-encoding "router reloaded" log output

Miciah:
Convert the reload script's output from a byte slice to a string before logging it so that it is displayed properly.  Before https://github.com/openshift/router/pull/47/commits/dc68ebfcbd9fdc600391ebe050228534dd2b07c3, log messages looked like the following:

    Router reloaded:
     - Proxy protocol on, checking http://localhost:80 ...
     - Health check ok : 0 retry attempt(s).

Since that change, the reload script's output is base64-encoded in logs:

    router reloaded	{"output": "IC0gUHJveHkgcHJvdG9jb2wgb24sIGNoZWNraW5nIGh0dHA6Ly9sb2NhbGhvc3Q6ODAgLi4uCiAtIEhlYWx0aCBjaGVjayBvayA6IDAgcmV0cnkgYXR0ZW1wdChzKS4K"}

This PR restores the previous, non-encoded log messages.

* `pkg/router/template/router.go` (`reloadRouter`): Convert `out` to string.

----

- [ ] Check CI output to verify that it looks as expected.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -545,7 +545,7 @@ func (r *templateRouter) reloadRouter() error {
 	if err != nil {
 		return fmt.Errorf("error reloading router: %v\n%s", err, string(out))
 	}
-	log.V(0).Info("router reloaded", "output", out)
+	log.V(0).Info("router reloaded", "output", string(out))
 	return nil
 }
 


ironcladlou:
Logs now look good:
```
2019-10-15T08:15:43.772Z	INFO	router.template	template/router.go:548	router reloaded	{"output": " - Proxy protocol on, checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"}
2019-10-15T08:15:51.584Z	INFO	router.template	template/router.go:548	router reloaded	{"output": " - Proxy protocol on, checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"}
2019-10-15T08:16:01.503Z	INFO	router.template	template/router.go:548	router reloaded	{"output": " - Proxy protocol on, checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"}
2
```
Thanks!

Looking forward, I think we should be publishing this reload info as metrics if we're not already, and remove the pointless logging

Miciah:
> Looking forward, I think we should be publishing this reload info as metrics if we're not already, and remove the pointless logging

We already have a metric:

https://github.com/openshift/router/blob/445320a1bfe520c14bfb338de841a94ad1085a2b/pkg/router/template/router.go#L444

We should at least bump up the log level when the command does not fail.  I'll open a PR later today.

PR #57, opened 2019-10-28
Bug 1691190: hostindex: Passthrough displaces path-based TLS routes

Miciah:
Ensure that a passthrough route displaces any path-based TLS routes with the same host, because passthrough is incompatible with path-based routing.

* `pkg/router/controller/hostindex/activation.go` (`hasExistingMatch`): Return true if both routes are TLS and the existing route is a passthrough route.
* `pkg/router/controller/hostindex/hostindex_test.go` (`Test_hostIndex`): Verify that a passthrough route displaces path-based TLS routes with the same host, but does not displace non-TLS routes.


ironcladlou:
Thanks!

/lgtm

PR #59, opened 2019-10-31
Bug 1691190: hostindex: Path-based TLS displaces passthrough

Miciah:
Just as a passthrough route displaces any path-based TLS routes with the same host, a path-based TLS route displaces any passthrough route with the same host.

Follow-up to #57.

* `pkg/router/controller/hostindex/activation.go` (`hasExistingMatch`): Return true if both routes are TLS and *either* route is passthrough.
* `pkg/router/controller/hostindex/hostindex_test.go` (`Test_hostIndex`): Verify that a path-based TLS route displaces passthrough routes with the same host.

### DIFF
--- a/pkg/router/controller/hostindex/hostindex_test.go
+++ b/pkg/router/controller/hostindex/hostindex_test.go
@@ -304,6 +304,18 @@ func Test_hostIndex(t *testing.T) {
 			displaces: map[string]struct{}{"003": {}},
 			inactive:  map[string][]string{"test.com": {"003"}},
 		},
+		{
+			name:       "passthrough route that conflicts with path-based route",
+			activateFn: SameNamespace,
+			steps: []step{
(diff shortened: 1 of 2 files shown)


pkg/router/controller/hostindex/hostindex_test.go
```diff
@@ -304,6 +304,18 @@ func Test_hostIndex(t *testing.T) {
 			displaces: map[string]struct{}{"003": {}},
 			inactive:  map[string][]string{"test.com": {"003"}},
 		},
+		{
+			name:       "passthrough route that conflicts with path-based route",
+			activateFn: SameNamespace,
+			steps: []step{
+				{route: newRoute("test", "1", 1, 1, routev1.RouteSpec{Host: "test.com", TLS: &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge}, Path: "/foo"})},
+				{route: newRoute("test", "2", 2, 1, routev1.RouteSpec{Host: "test.com", TLS: &routev1.TLSConfig{Termination: routev1.TLSTerminationPassthrough}})},
```
ironcladlou: Nit for future reference: these test case functions are really hard to understand because of the unlabeled positional arguments

PR #60, opened 2019-11-11
Add logging verbosity flag backed by klog

ironcladlou:
Add a `-v` logging verbosity flag and switch to klog for consistency
with upstream and because zapr doesn't implement V levels directly anyway.


log/log.go
```diff
@@ -2,19 +2,14 @@ package log
...
 var Logger logr.Logger
 
 func init() {
-	// Set up logging.
-	zapLogger, err := zap.NewDevelopment(zap.AddCallerSkip(1), zap.AddStacktrace(zap.FatalLevel))
-	if err != nil {
-		panic(err)
-	}
-	Logger = zapr.NewLogger(zapLogger).WithName("router")
+	Logger = klogr.New()
```
frobware: Do we need the `init()`? Can't this be:

    var Logger logr.Logger = klogr.New()

In fact, do we need the package? Can this just move to `main()`?

frobware: > In fact, do we need the package? Can this just move to `main()`?

Oh, unless you want this as `log.Logger` - which does read a lot better.

ironcladlou: init can be removed, but the package is because otherwise all the library code would have to depend on the command package.

ironcladlou: Fixed

PR #61, opened 2019-11-13
Bug 1771783: Dynamically reload metrics certificate/key file changes

ironcladlou:
The metrics certificate can be rotated. Add support to dynamically reload
the certificate and key file when they change. This enables certificate
rotation without incurring an expensive restart.

Before this change, metrics scraping can begin failing when the metrics
endpoint's certificate is invalidated and can't be fixed without restarting
the controller.

### DIFF
--- a/pkg/cmd/infra/router/template.go
+++ b/pkg/cmd/infra/router/template.go
@@ -9,12 +9,15 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/MakeNowJust/heredoc"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+
+	"github.com/fsnotify/fsnotify"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -33,6 +36,7 @@ import (
 	routelisters "github.com/openshift/client-go/route/listers/route/v1"
 	"github.com/openshift/library-go/pkg/crypto"
 	"github.com/openshift/library-go/pkg/proc"
+
 	"github.com/openshift/router/pkg/router"
 	"github.com/openshift/router/pkg/router/controller"
 	"github.com/openshift/router/pkg/router/metrics"
@@ -417,16 +421,13 @@ func (o *TemplateRouterOptions) Run() error {
 			LiveChecks:  liveChecks,
 			ReadyChecks: []healthz.HealthChecker{checkBackend, checkSync},
 		}
-		if certFile := env("ROUTER_METRICS_TLS_CERT_FILE", ""); len(certFile) > 0 {
-			certificate, err := tls.LoadX509KeyPair(certFile, env("ROUTER_METRICS_TLS_KEY_FILE", ""))
-			if err != nil {
-				return err
-			}
-			l.TLSConfig = crypto.SecureTLSConfig(&tls.Config{
-				Certificates: []tls.Certificate{certificate},
-				ClientAuth:   tls.RequestClientCert,
-			})
+
+		if tlsConfig, err := makeTLSConfig(); err != nil {
+			return err
+		} else {
+			l.TLSConfig = tlsConfig
 		}
+
 		l.Listen()
 
 		// on reload, invoke the collector to preserve whatever metrics we can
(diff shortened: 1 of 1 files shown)


ironcladlou:
@Miciah @frobware PTAL. I still need to manually test this in a cluster.

pkg/cmd/infra/router/template.go
```diff
@@ -565,3 +566,88 @@ func (o *TemplateRouterOptions) blueprintRoutes(routeclient *routeclientset.Clie
 
 	return blueprints, nil
 }
+
+// makeTLSConfig checks whether metrics TLS is configured and
+// if so returns a tls.Config whose certificate is automatically
+// reloaded every 5 minutes or in response to file changes.
```
marun: What problem is the time-based reload intended to address?

pkg/cmd/infra/router/template.go
```diff
@@ -565,3 +566,88 @@ func (o *TemplateRouterOptions) blueprintRoutes(routeclient *routeclientset.Clie
...
+	keyFile := env("ROUTER_METRICS_TLS_KEY_FILE", "")
+
+	// Make sure we have something useful to start with
+	certificate, err := tls.LoadX509KeyPair(certFile, keyFile)
+	if err != nil {
+		return nil, err
+	}
+
+	// Safely reload the certificate as requested.
+	var lock sync.Mutex
```
Miciah: Might as well use `sync.RWMutex`.

pkg/cmd/infra/router/template.go
```diff
@@ -565,3 +566,88 @@ func (o *TemplateRouterOptions) blueprintRoutes(routeclient *routeclientset.Clie
...
+	reload := make(chan bool) // value will be false for an unscheduled reload
+	go func() {
+		for {
+			select {
+			case scheduled := <-reload:
+				latest, err := tls.LoadX509KeyPair(certFile, keyFile)
+				if err != nil {
+					// TODO: add a prometheus metric? Fail a health check? This should surface somehow...
+					log.Error(err, "failed to reload certificate", "cert", certFile, "key", keyFile)
```
Miciah: The router has a lot of certificates, so using a specific phrase, such as "metrics certificate", here and below would reduce risk of confusion.

pkg/cmd/infra/router/template.go
```diff
@@ -565,3 +566,88 @@ func (o *TemplateRouterOptions) blueprintRoutes(routeclient *routeclientset.Clie
...
+		for {
+			select {
+			case <-ticker.C:
+				reload <- true
+			case _, ok := <-watcher.Events:
+				if ok {
+					// Ignoring the event type for now to avoid any portability
+					// issues at the expense of potentially spurious reloads on
+					// irrelevant events (which should be rare, this is a mounted
+					// secret).
```
Miciah: What you're like to see is a flurry of events for the same file (e.g., a `CREATE` event and a `CHMOD` event for the same update).

pkg/cmd/infra/router/template.go
```diff
@@ -565,3 +566,88 @@ func (o *TemplateRouterOptions) blueprintRoutes(routeclient *routeclientset.Clie
...
+	// Reload when the files change or when the fixed timer fires
+	ticker := time.NewTicker(5 * time.Minute)
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, err
+	}
+	if err := watcher.Add(certFile); err != nil {
+		return nil, err
+	}
+	if err := watcher.Add(keyFile); err != nil {
```
Miciah: I think you will encounter this issue:  https://github.com/openshift/router/blob/b4a6b0823a2e71080eec39c021811cd603fc0be8/pkg/router/template/router.go#L288-L302

pkg/cmd/infra/router/template.go
Miciah: We don't trust fsnotify not to lose events.

pkg/cmd/infra/router/template.go
ironcladlou: I was afraid of something like that when using fsnotify... I worked around this recently in another unrelated project by relying on one of the candidate forks which support not following symlinks (e.g. https://github.com/fsnotify/fsnotify/pull/289). 

What's worse, introducing this localized and slightly insane complexity, or patching fsnotify (i.e. depending on a fork)?

pkg/cmd/infra/router/template.go
ironcladlou: Disambiguating means trusting the event type, and I have reason to doubt they are reliable across contexts based on some previous experience with the lib. Given how infrequent the updates should be (when a cert is rotated), do we really care if we trigger a few spurious reloads in this case?

ironcladlou: And by "reload" here I just mean we'll read the disk contents of the cert/key file (which are probably mounted through a RAM volume anyway)

pkg/cmd/infra/router/template.go
ironcladlou: Bikeshed, but RWMutex has (I think) objectively slightly trickier semantics, and this seems like a low to almost-no contention code path... (contention would occur when a cert secret changed during a rotation at the instant prometheus opens a connection for its periodic scrape?)

ironcladlou: And even then, we're talking about resolving the contention while a couple of files are read from disk (likely mapped to a RAM disk)... so a really dumb/simple lock seems appropriate

pkg/cmd/infra/router/template.go
```diff
@@ -565,3 +566,88 @@ func (o *TemplateRouterOptions) blueprintRoutes(routeclient *routeclientset.Clie
...
+					log.V(4).Info("reloaded certificate on fixed schedule", "cert", certFile, "key", keyFile)
+				} else {
+					log.V(0).Info("reloaded modified certificate", "cert", certFile, "key", keyFile)
+				}
+			}
+		}
+	}()
+
+	// Reload when the files change or when the fixed timer fires
+	ticker := time.NewTicker(5 * time.Minute)
```
frobware: The construction of the ticker and the watch.Add(s) could(should?) come before starting the goroutines so that we only start them (goroutines) on success.

pkg/cmd/infra/router/template.go
Miciah: I'd re-use the workaround from `watchSecretDir`.  It is well understood and has been working for us.

pkg/cmd/infra/router/template.go
Miciah: Agreed, it's not terrible to trigger spurious reloads.

pkg/cmd/infra/router/template.go
Miciah: All right, yeah, since we have only the one reader, `sync.Mutex` is no worse than `sync.RWMutex`.  It seems like natural place to use `sync.RWMutex`, but I realize now doing so would not help anything.

pkg/cmd/infra/router/template.go
```diff
@@ -565,3 +566,77 @@ func (o *TemplateRouterOptions) blueprintRoutes(routeclient *routeclientset.Clie
...
+				latest, err := tls.X509KeyPair(latestCertBytes, latestKeyBytes)
+				if err != nil {
+					log.Error(err, "failed to reload certificate", "cert", certFile, "key", keyFile)
+					break
+				}
+				lock.Lock()
+				certificate = latest
+				certBytes = latestCertBytes
+				keyBytes = latestKeyBytes
+				lock.Unlock()
```
Miciah: We only need the lock around `certificate = latest`, right?  That is, we don't need the lock when assigning `latestCertBytes` and `latestKeyBytes`?  It's confusing to lock during access of `latestCertBytes` and `latestKeyBytes` here but not for the comparison above.

ironcladlou: You're right — `certBytes` and `keyBytes` are single reader/writer and don't require guarding.

Fixed.

pkg/cmd/infra/router/template.go
```diff
@@ -565,3 +566,78 @@ func (o *TemplateRouterOptions) blueprintRoutes(routeclient *routeclientset.Clie
...
+				if bytes.Equal(latestCertBytes, certBytes) && bytes.Equal(latestKeyBytes, keyBytes) {
+					// Nothing changed, try later.
+					break
+				}
+
+				// Something changed, reload the certificate.
+				certBytes = latestCertBytes
+				keyBytes = latestKeyBytes
+				latest, err := tls.X509KeyPair(certBytes, keyBytes)
+				if err != nil {
```
Miciah: This means the router will not keep trying to reload the certificate if `tls.X509KeyPair` fails.  Now that I think about it, that is probably a good thing—`tls.X509KeyPair` is unlikely to have a transient error, so if loading fails, it probably isn't worth retrying unless the certificate changes.

ironcladlou: Good catch, this was _not_ my intent.

In the absence of analyzing the error to establish whether it's actually transient, I would prefer to be defensive and retry. Also, I find it potentially confusing to quietly cache the state as if we had succeeded.

Objection to me predicating the certBytes and keyBytes assignment on successful loading?

Miciah: No objection.  It's a tradeoff between some log noise from retrying when the certificate is invalid, or forcing the admin to delete the secret so a new one gets regenerated in case of a transient failure.  I don't feel strongly one way or the other.

PR #64, opened 2019-11-20
Bug 1774353: fix version reporting

ironcladlou:
Fix git version reporting in the build scripting and `version` subcommand.


ironcladlou:
Implementation lifted from https://github.com/openshift/library-go/blob/master/alpha-build-machinery/make/examples/multiple-binaries/pkg/version/version.go

PR #65, opened 2019-11-22
RFE-142 - Honor X-Forwarded-For HTTP header in route-specific IP Whitelist feature

koep:
Proposal pull request to address [RFE-142](https://jira.coreos.com/browse/RFE-142)


jkroepke:
If openshift is accessible from internet how to protect agist faked `X-Forwarded-For` Header like?

`curl -H "X-Forwarded-For: 127.0.0.1" redhat.com`

PR #68, opened 2019-12-06
Bug 1780398: status: performIngressConditionUpdate: Retry on 403

Miciah:
If a status update fails with a forbidden error, retry the update.

Keep retrying failed updates as long as the route has the wrong status.

* `pkg/router/controller/status.go` (`performIngressConditionUpdate`): Retry on forbidden errors.  Never give up retrying after an error.


ironcladlou:
Makes sense, thanks!

/lgtm

PR #72, opened 2019-12-16
Default to v4v6 mode in haproxy template

russellb:
The previous default was IPv4 only.  This new default allows both IPv4
and IPv6 to work.

This should be a safe default value. The case where this
would break would be a RHEL node with IPv6 completely disabled.  I'm
not sure if that's a realistic case to worry about, though.

Otherwise, we'll have to make this driven by the ingress operator
based on the configuration of the cluster.


russellb:
After some more thought, this should just be redone to be dynamic based on cluster configuration ...

PR #73, opened 2019-12-20
[WIP] Print haproxy version when building image

frobware:
Trying to understand what version of haproxy we get when the package
to install is just `haproxy`.


frobware:
No longer needed.

PR #75, opened 2019-12-20
Enable http2 negotiation on frontend and backends

ironcladlou:


knobunc:
This looks reasonable and sane.

/lgtm

ironcladlou:
We're still relatively early into the release, and this change is so trivial to revert... let's merge it and see what problems arise and what use cases it enables.

/retitle Enable http2 negotiation on frontend and backends

PR #90, opened 2020-02-18
Bug 1803001: Removes Unneeded proto-version Header

danehans:
PR #8 incorrectly double quoted the proto-version field value, even when this value was empty.  Per [rfc7230 section-3.2.6](https://tools.ietf.org/html/rfc7230#section-3.2.6), field values must consist of 1 character (`1*tchar`). During my investigation, I could not find any references for the `proto-version` header. This header appears to be unnecessary and is removed by this PR.

/assign @ironcladlou @Miciah

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -431,11 +431,11 @@ backend {{genBackendNamePrefix $cfg.TLSTermination}}:{{$cfgIdx}}
   http-request set-header X-Forwarded-Proto http if !{ ssl_fc }
   http-request set-header X-Forwarded-Proto https if { ssl_fc }
   http-request set-header X-Forwarded-Proto-Version h2 if { ssl_fc_alpn -i h2 }
-  # Forwarded header: quote IPv6 addresses and values that may be empty as per https://tools.ietf.org/html/rfc7239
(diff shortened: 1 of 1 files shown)


danehans:
`error: could not run steps: step e2e-aws failed: template pod "e2e-aws" failed: the pod ci-op-4m01std9/e2e-aws failed after 1m27s (failed containers: setup): ContainerFailed one or more containers exited`

/test e2e-aws

ironcladlou:
/approve

@Miciah PTAL and tag?

Miciah:
Do you have a reference handy to the standard that specifies `proto-version`?  I assume it would fall under https://tools.ietf.org/html/rfc7239#section-5.4 ("Extensions"), but is it explicitly defined in any standard?

danehans:
@Miciah that appears to be the correct reference for the `proto` field. However, I can't find any references for `proto-version`. Is `proto-version` a valid header?

Miciah:
I don't think it is.  The header was added with HTTP/2 support in https://github.com/openshift/origin/pull/19968/commits/a6edc171d715c58a14c00ca6baaf20d702bd956e.  I think we should just drop it.  (We can keep `X-Forwarded-Proto-Version` though as the `X-` prefix indicates that it is non-standard.)

PR #93, opened 2020-02-25
Add toleration to allow pods to schedule to edge nodes

KodieGlosserIBM:
I would like to add the default behavior to allow for scheduling of the router pods to edge nodes. It is a way for us to restrict traffic to just these nodes https://cloud.ibm.com/docs/containers?topic=containers-edge.

### DIFF
--- a/deploy/router.yaml
+++ b/deploy/router.yaml
@@ -14,6 +14,11 @@ spec:
       labels:
         k8s-app: ingress-router
     spec:
+      tolerations:
+        - key: "dedicated"
+          operator: "Equal"
+          value: "edge"
+          effect: NoSchedule
       serviceAccountName: ingress-router
       containers:
       - env:


ironcladlou:
I had actually forgotten the files in `deploy/` exist — in the 4.x branch lines (e.g. master) I'm actually not sure we need them anymore. They should at least be clearly marked as an example not intended for any sort of serious use. They're not maintained, verified, or used in the platform in any way.

cc @openshift/sig-network-edge

ironcladlou:
@KodieGlosserIBM were you intended to propose a change in the default scheduling behavior for ingress (router) pods in an OpenShift v4 install? If so, the change would belong in the [ingress operator](https://github.com/openshift/cluster-ingress-operator) which manages the router deployments.

KodieGlosserIBM:
@ironcladlou yes I was, thanks for the pointer!

ironcladlou:
> @KodieGlosserIBM were you intended to propose a change in the default scheduling behavior for ingress (router) pods in an OpenShift v4 install? If so, the change would belong in the [ingress operator](https://github.com/openshift/cluster-ingress-operator) which manages the router deployments.

Specifically, [here](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/deployment.go).

Although, the [ingress operator API](https://github.com/openshift/api/blob/master/operator/v1/types_ingress.go#L184) already supports user-defined toleration management, so we'd need to think through whether changing the defaults makes sense.

PR #96, opened 2020-03-03
[release-4.2] Bug 1809398: Dynamically reload metrics certificate/key file changes

marun:
Cherry-pick of #61. It didn't apply cleanly, the logging needed to be switched to glog for the release-4.2 branch.

/cc @ironcladlou

### DIFF
--- a/pkg/cmd/infra/router/template.go
+++ b/pkg/cmd/infra/router/template.go
@@ -590,7 +590,20 @@ func makeTLSConfig() (*tls.Config, error) {
 		return nil, err
 	}
 
-	// Safely reload the certificate as requested.
+	// Reload when the files change or when the fixed timer fires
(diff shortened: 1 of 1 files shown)


ironcladlou:
@Miciah @frobware @danehans re: logging, not sure if I like the idea of a special patch because it's going to cause history divergence and liability if we need to pick other patches later. WDYT?

marun:
> @Miciah @frobware @danehans re: logging, not sure if I like the idea of a special patch because it's going to cause history divergence and liability if we need to pick other patches later. WDYT?

I think history divergence between release and master branches is to be expected. If that wasn't the case, there would be no possibility of selective backport. 

Also, there's nothing precluding a revert of the bespoke patch in the future should it become a liability.

ironcladlou:
Sounds like everyone's good with it, so

/approve

PR #99, opened 2020-03-04
Bug 1810181: Wildcard is not needed to copy subdirectory in dockerfile

smarterclayton:
Docker treats /* as a "collapse subdirectory" option, so we should
avoid potentially incompatible behavior across multiple builders.


ironcladlou:
/title Bug 1813507: Wildcard is not needed to copy subdirectory in dockerfile

ironcladlou:
/retitle Bug 1813507: Wildcard is not needed to copy subdirectory in dockerfile

PR #104, opened 2020-03-18
Add HAProxy annotation for http2/cleartext (h2c) routes

frobware:
If your route type is "termination: edge" then your backend may still
need http2 (e.g., gRPC servers). To support such configurations you
now need to set the following annotation on an edge route:

  "haproxy.router.openshift.io/enable-h2c: true"

Enabling http2 cleartext (h2c) is documented in the haproxy-1.9
releases notes[1]:

  The "proto h2" directive allows HAProxy to communicate using HTTP/2
  without TLS, such as HTTP/2-enabled backends like Varnish and H2O,
  and gRPC servers.

With this annotation enabled the server backend will change from:

    server server1 192.168.1.13:80 alpn h2,http/1.1

to:

    server server1 192.168.1.13:80 proto h2

[1] https://www.haproxy.com/blog/haproxy-1-9-has-arrived/

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -454,7 +454,11 @@ backend {{genBackendNamePrefix $cfg.TLSTermination}}:{{$cfgIdx}}
     {{- if ge $weight 0 }}{{/* weight=0 is reasonable to keep existing connections to backends with cookies as we can see the HTTP headers */}}
       {{- with $serviceUnit := index $.ServiceUnits $serviceUnitName }}
         {{- range $idx, $endpoint := processEndpointsForAlias $cfg $serviceUnit (env "ROUTER_BACKEND_PROCESS_ENDPOINTS" "") }}
+	  {{- if isTrue (index $cfg.Annotations "haproxy.router.openshift.io/enable-h2c") }}
+  server {{$endpoint.ID}} {{$endpoint.IP}}:{{$endpoint.Port}} cookie {{$endpoint.IdHash}} weight {{$weight}} proto h2
+          {{- else }}
   server {{$endpoint.ID}} {{$endpoint.IP}}:{{$endpoint.Port}} cookie {{$endpoint.IdHash}} weight {{$weight}} alpn h2,http/1.1
+          {{- end }}
           {{- if (eq $cfg.TLSTermination "reencrypt") }} ssl
             {{- if $cfg.VerifyServiceHostname }} verifyhost {{ $serviceUnit.Hostname }}
             {{- end }}
--- a/pkg/router/template/configmanager/haproxy/manager.go
+++ b/pkg/router/template/configmanager/haproxy/manager.go
@@ -1094,6 +1094,7 @@ func modAnnotationsList(termination routev1.TLSTerminationType) []string {
 		"haproxy.router.openshift.io/rate-limit-connections.rate-tcp",
 		"haproxy.router.openshift.io/rate-limit-connections.rate-http",
 		"haproxy.router.openshift.io/pod-concurrent-connections",
+		"haproxy.router.openshift.io/enable-h2c",
 		"router.openshift.io/haproxy.health.check.interval",
 	}
 


frobware:
/hold

Want to hear more feedback before we commit to this.

ironcladlou:
A question I'm hoping someone more knowledgable can answer:  is it possible to automatically make the correct server entry based on the Route itself without a new annotation? If not, why?

frobware:
> A question I'm hoping someone more knowledgable can answer: is it possible to automatically make the correct server entry based on the Route itself without a new annotation? If not, why?

I believe it should be possible to do this from inspecting `Content-Type` and for cases that begin with `application/grpc` it would automatically dispatch to this new backend. That backend would always uses `proto h2` for the server.

I'll follow up with a new PR.

frobware:
The edge termination looks like:

```
backend be_no_sni
  server fe_no_sni 127.0.0.1:10443 weight 1

frontend fe_no_sni
  # terminate ssl on edge
  bind 127.0.0.1:10443 ssl crt /home/aim/localhost-bundle.pem alpn h2,http/1.1
  mode http
  # Strip off Proxy headers to prevent HTTpoxy (https://httpoxy.org/)
  http-request del-header Proxy
  # DNS labels are case insensitive (RFC 4343), we need to convert the hostname into lowercase
  # before matching, or any requests containing uppercase characters will never match.
  http-request set-header Host %[req.hdr(Host),lower]
  # map to backend
  # Search from most specific to general path (host case).
  # Note: If no match, haproxy uses the default_backend, no other
  #       use_backend directives below this will be processed.

##### ***** NEW AUTO DETECTION OF GRPC ***** #####
  acl is_grpc hdr_beg(Content-Type) -i application/grpc
  use_backend %[base,map_reg(/home/aim/haproxy-grpc-sample/sample/haproxy/os_edge_reencrypt_be.map)]-grpc if is_grpc

  use_backend %[base,map_reg(/home/aim/haproxy-grpc-sample/sample/haproxy/os_edge_reencrypt_be.map)]
  default_backend openshift_default
```

In there we can detect if the `Content-Type` begins with `application/grpc` and if so we dispatch to a backend that looks like:

```
backend be_edge_http:default:grpc-interop-edge-grpc
  mode http
  option redispatch
  option forwardfor
  balance leastconn
  timeout check 5000ms
  http-request set-header X-Forwarded-Host %[req.hdr(host)]
  http-request set-header X-Forwarded-Port %[dst_port]
  http-request set-header X-Forwarded-Proto http if !{ ssl_fc }
  http-request set-header X-Forwarded-Proto https if { ssl_fc }
  http-request set-header X-Forwarded-Proto-Version h2 if { ssl_fc_alpn -i h2 }
  server pod:grpc-interop:grpc-interop:127.0.0.1:1110 127.0.0.1:1110 cookie 094c1c2fba73c5173a67efa50555c79f weight 256 alpn h2
```

as opposed to:

```
# Plain http backend or backend with TLS terminated at the edge or a
# secure backend with re-encryption.
backend be_edge_http:default:grpc-interop-edge
  mode http
  option redispatch
  option forwardfor
  balance leastconn
  timeout check 5000ms
  http-request set-header X-Forwarded-Host %[req.hdr(host)]
  http-request set-header X-Forwarded-Port %[dst_port]
  http-request set-header X-Forwarded-Proto http if !{ ssl_fc }
  http-request set-header X-Forwarded-Proto https if { ssl_fc }
  http-request set-header X-Forwarded-Proto-Version h2 if { ssl_fc_alpn -i h2 }
  server pod:grpc-interop:grpc-interop:127.0.0.1:1110 127.0.0.1:1110 cookie 094c1c2fba73c5173a67efa50555c79f weight 256 alpn h2,http/1.1
```

The difference between the two is `proto h2` versus `alpn h2,http/1.1`. 

Perhaps there's a way of encoding this in a `set-var` and do the interpolation on the `server` line.

@openshift/sig-network-edge

frobware:
> Perhaps there's a way of encoding this in a `set-var` and do the interpolation on the `server` line.

I think we need to find a way to do this...

frobware:
> Perhaps there's a way of encoding this in a `set-var` and do the interpolation on the `server` line.

Stepping through the parsing of this line there is no variable interpolation done. It just matches against static strings:

```c
static struct mux_proto_list mux_proto_h2_htx =
	{ .token = IST("h2"), .mode = PROTO_MODE_HTX, .side = PROTO_SIDE_BOTH, .mux = &h2_htx_ops };
```

frobware:
There is a `use-server` directive which I think will do what we want:

```
  use-server pod:grpc-interop:grpc-interop:127.0.0.1:1110-h2c if { hdr_beg(Content-Type) -i application/grpc }
  server pod:grpc-interop:grpc-interop:127.0.0.1:1110-h2c 127.0.0.1:1110 cookie 094c1c2fba73c5173a67efa50555c79f weight 256 proto h2
```
and then the final line is as before:

```
server pod:grpc-interop:grpc-interop:127.0.0.1:1110 127.0.0.1:1110 cookie 094c1c2fba73c5173a67efa50555c79f weight 256 alpn h2,http/1.1
```

which is the default if we're not `application/grpc`.

The `use-server` docs state:

```
use-server <server> if <condition>
use-server <server> unless <condition>
  Only use a specific server if/unless an ACL-based condition is matched.
  May be used in sections :   defaults | frontend | listen | backend
                                  no   |    no    |   yes  |   yes
  Arguments :
    <server>    is the name of a valid server in the same backend section.

    <condition> is a condition composed of ACLs, as described in section 7.

  By default, connections which arrive to a backend are load-balanced across
  the available servers according to the configured algorithm, unless a
  persistence mechanism such as a cookie is used and found in the request.

  Sometimes it is desirable to forward a particular request to a specific
  server without having to declare a dedicated backend for this server. This
  can be achieved using the "use-server" rules. These rules are evaluated after
  the "redirect" rules and before evaluating cookies, and they have precedence
  on them. There may be as many "use-server" rules as desired. All of these
  rules are evaluated in their declaration order, and the first one which
  matches will assign the server.

  If a rule designates a server which is down, and "option persist" is not used
  and no force-persist rule was validated, it is ignored and evaluation goes on
  with the next rules until one matches.

  In the first form, the server will be used if the condition is met. In the
  second form, the server will be used if the condition is not met. If no
  condition is valid, the processing continues and the server will be assigned
  according to other persistence mechanisms.

  Note that even if a rule is matched, cookie processing is still performed but
  does not assign the server. This allows prefixed cookies to have their prefix
  stripped.

  The "use-server" statement works both in HTTP and TCP mode. This makes it
  suitable for use with content-based inspection. For instance, a server could
  be selected in a farm according to the TLS SNI field. And if these servers
  have their weight set to zero, they will not be used for other traffic.

  Example :
     # intercept incoming TLS requests based on the SNI field
     use-server www if { req_ssl_sni -i www.example.com }
     server     www 192.168.0.1:443 weight 0
     use-server mail if { req_ssl_sni -i mail.example.com }
     server     mail 192.168.0.1:587 weight 0
     use-server imap if { req_ssl_sni -i imap.example.com }
     server     imap 192.168.0.1:993 weight 0
     # all the rest is forwarded to this server
     server  default 192.168.0.2:443 check
```

frobware:
> But that doesn't fix support for Varnish and H2O ?

This is a good point. I think then we can do some auto detection but the annotation value takes precedence.

frobware:
Docs[1] for 2.0 (which is what we're using) has:

"You can also use HTTP/2 without TLS. Remove any ssl and verify parameters from the server and/or bind lines. Then swap alpn h2 for proto h2 and HAProxy will use only the given protocol."

[1] https://www.haproxy.com/blog/haproxy-2-0-and-beyond/#end-to-end-http-2

frobware:
Some related discussion in this area: https://bugzilla.redhat.com/show_bug.cgi?id=1826990

frobware:
We're going to have to be a lot smarter here, or get opt-in for what is actually on the backend.

In this PR I proposed that if the route had the annotation we would swap out `alpn h2,http1.1` for `proto h2` on the `server` directive. This would certainly work for my simple gRPC server:

```go
		server := grpc.NewServer()
		testpb.RegisterTestServiceServer(server, interop.NewTestServer())

		lis, err := net.Listen("tcp", ":"+lookupEnv("H2C_PORT", defaultH2CPort))
		if err != nil {
			log.Fatalf("listen failed: %v", err)
		}

		log.Printf("Serving h2c at: %v", lis.Addr())

		if err = server.Serve(lis); err != nil {
			log.Fatalf("failed to serve: %v", err)
		}
```

However, not unreasonably, you could be multiplexing this single server port, ala: 

```go
		l, err := net.Listen("tcp", "127.0.0.1:23456")
		if err != nil {
			log.Fatal(err)
		}

		mux := cmux.New(l)

		// Match connections in order:
		// First grpc, then HTTP.
		grpcListener := mux.MatchWithWriters(cmux.HTTP2MatchHeaderFieldSendSettings("content-type", "application/grpc"))
		httpListener := mux.Match(cmux.HTTP1())

		grpcServer := grpc.NewServer()
		testpb.RegisterTestServiceServer(grpcServer, interop.NewTestServer())

		httpServer := &http.Server{
			Handler: h2c.NewHandler(&httpServer{}, &http2.Server{}),
		}

		go grpcServer.Serve(grpcListener)
		go httpServer.Serve(httpListener)

		log.Printf("Serving h2 and h2c at: %v", l.Addr())

		if err := mux.Serve(); err != nil {
			log.Fatalf("mux server failed: %v", err)
		}
```

and in this case gRPC would work, but HTTP would not.

danehans:
> A question I'm hoping someone more knowledgable can answer: is it possible to automatically make the correct server entry based on the Route itself without a new annotation? If not, why?

I think we can accomplish this use case without introducing an annotation, if we're ok with requiring k8s v1.18. In 1.18, the route controller can take advantage of the [appProtocol](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#serviceport-v1-core) field of a `Service` to discover the need to use h2c for a TLS route configured for edge termination. The route controller can look-up the `appProtocol` of the route's referenced Service `targetPort` to determine if htc should be used.

frobware:
> The route controller can look-up the `appProtocol` of the route's referenced Service `targetPort` to determine if htc should be used.

What would this mapping be? What value from `appProtocol` would make us change the template to add `proto h2`?

danehans:
> What would this mapping be? What value from appProtocol would make us change the template to add proto h2?

@frobware the appProtocol approach is similar to the annotation approach. The route controller gets the Service associated to the route. If the appProtocol field of the Service is marked with the IANA standard service name for htc (as per RFC-6335 and http://www.iana.org/assignments/service-names), then the controller programs the htc config in the associated backend. If htc is not listed in RFC-6335, we can use prefixed names such as mycompany.com/my-custom-protocol.

frobware:
> If the appProtocol field of the Service is marked with the IANA standard service name for htc (as per RFC-6335 and http://www.iana.org/assignments/service-names), then the controller programs the htc config in the associated backend. If htc is not listed in RFC-6335, we can use prefixed names such as mycompany.com/my-custom-protocol.

What happens if we were to switch to a different proxy? Would/could/should the semantics naturally carry over? I didn't see a mention of h2c in the list so if we had "openshift.io/enable-h2c" would that specific value break something else somewhere else that may not be expecting that value. All of this feels clunkiness feels very specific to patching up the haproxy template and I wonder whether generalising via `appProtocol` comes back to bite us.

frobware:
Closing this. We can revisit this or wait for the `appProtocol` field to be generally available. Right now we don't have a good story for the case where the port is multiplexed and just adding "proto h2" won't fix that case.

PR #106, opened 2020-03-24
Bug 1816544: [4.4] Backport "remove proto-version header field"

fmartingr:
cherry-picked changes from #90


danehans:
@sdodson I see you're the OCP Patch Manager. When you have a moment, PTAL and tag.

PR #51, opened 2019-10-09
Add ServiceUnitKey and ServiceAliasConfigKey types

Miciah:
Define and use the `ServiceUnitKey` and `ServiceAliasConfigKey` types for key values of maps of `ServiceUnit` values and of `ServiceAliasConfig` values, respectively, in order to make code dealing with such maps clearer and reduce risk of mixing key values.

* `pkg/router/template/configmanager/haproxy/backend.go` (`Backend`): Use `ServiceAliasConfigKey` for the name.
(`buildHAProxyBackends`, `newBackend`, `Name`, `ApplyChanges`): Adapt to use the new `ServiceAliasConfigKey` type.
* `pkg/router/template/configmanager/haproxy/blueprint_plugin_test.go` (`fakeConfigManager`): Use `ServiceAliasConfigKey` for the `blueprints` map.
(`newFakeConfigManager`, `FindBlueprint`, `Register`, `AddRoute`, `RemoveRoute`, `ReplaceRouteEndpoints`, `RemoveRouteEndpoints`, `routeKey`):
* `pkg/router/template/configmanager/haproxy/client.go` (`FindBackend`):
* `pkg/router/template/configmanager/haproxy/client_test.go` (`TestClientCommit`, `TestClientBackends`): Adapt to use the
`ServiceAliasConfigKey` type.
* `pkg/router/template/configmanager/haproxy/manager.go` (`configEntryMap`): Use the `ServiceAliasConfigKey` type for the map's values.
(`routeBackendEntry`): Use the `ServiceAliasConfigKey` type for `backendName` and the `poolRouteBackendName`, `backendEntries`, and `poolUsage` maps' keys.
(`NewHAProxyConfigManager`, `Register`, `AddRoute`, `RemoveRoute`, `ReplaceRouteEndpoints`, `RemoveRouteEndpoints`, `findFreeBackendPoolSlot`, `reset`, `BackendName`, `BuildMapAssociations`, `routeBackendName`, `applyMapAssociations`):
* `pkg/router/template/configmanager/haproxy/map.go` (`Add`, `addEntry`):
* `pkg/router/template/configmanager/haproxy/map_test.go` (`TestHAProxyMapAdd`): Adapt to use the `ServiceAliasConfigKey` type.
* `pkg/router/template/fake.go` (`NewFakeTemplateRouter`): Adapt to use `ServiceUnitKey` and `ServiceAliasConfigKey`.
[...]

PR #108, opened 2020-03-25
Use ServiceAliasConfigKeys for dynamic config

### LINKED ISSUE #107: Dynamic config manager not functional in 4.4
While trying the 4.4 router, I noticed that [enabling the dynamic configuration manager](https://docs.openshift.com/container-platform/3.11/install_config/router/default_haproxy_router.html#using-the-dynamic-configuration-manager) in a custom router deployment results in the router failing to start.

The following error is seen in the router pod logs:
 
```
E0324 18:57:02.396551       1 limiter.go:165] error executing template for file /var/lib/haproxy/conf/haproxy.config: template: haproxy-config.template:498:65: executing "/var/lib/haproxy/conf/haproxy.config" at <$cfgIdx>: wrong type for value; expected string; got templaterouter.ServiceAliasConfigKey
```

It seems that the change of type in https://github.com/openshift/router/pull/51 forgot some functions used only with the dynamic configuration manager. It looks like an easy fix that I'm happy to provide.

But that raises a question: @ramr @smarterclayton is the dynamic config manager still supported in Openshift 4? We are preparing the migration from Openshift 3.11 and I'm afraid we really need it: we have applications with many long-standing (websocket) connections. If the router reloads configuration often, old haproxy processes tend to stay around to keep those connections alive and accumulate to the point of exhausting all available memory on router nodes. The dynamic config manager alleviates the problem considerably.

alexcern:
Following https://github.com/openshift/router/pull/51 doing it for other template methods.

Verified that the router is functional with these changes and `ROUTER_HAPROXY_CONFIG_MANAGER='true'`

Fixes #107

### DIFF
--- a/pkg/router/template/configmanager/haproxy/blueprint_plugin_test.go
+++ b/pkg/router/template/configmanager/haproxy/blueprint_plugin_test.go
@@ -63,15 +63,15 @@ func (cm *fakeConfigManager) RemoveRouteEndpoints(id templaterouter.ServiceAlias
 func (cm *fakeConfigManager) Notify(event templaterouter.RouterEventType) {
 }
 
-func (cm *fakeConfigManager) ServerTemplateName(id string) string {
+func (cm *fakeConfigManager) ServerTemplateName(id templaterouter.ServiceAliasConfigKey) string {
 	return "fakeConfigManager"
 }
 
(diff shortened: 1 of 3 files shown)


Miciah:
Nice, thanks!
/lgtm

PR #110, opened 2020-04-03
Set log level 2 by default

smarterclayton:
Testing to see whether we have uncovered error paths resulting in
readiness failures.


smarterclayton:
/assign @ironcladlou 

you added the flag parse, but it was preventing `--help` from working.  Fixed that and set the default log level.  I manually tested that `a -v=2 -v=6` results in v = 6

ironcladlou:
I don't see why not...

/lgtm

PR #114, opened 2020-04-17
build: Add verify target

frobware:
Add verify build target that checks for:
- gofmt
- go module correctness

Ran `make verify` and fixed the one formatting change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,3 +38,8 @@ images/router/*/Dockerfile.rhel: images/router/base/Dockerfile.rhel
 
 check:
 	$(GO) test -race ./...
+
+.PHONY: verify
+verify:
+	hack/verify-gofmt.sh
+	hack/verify-deps.sh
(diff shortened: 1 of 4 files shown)


ironcladlou:
Nice, thanks... guess we'll need a corresponding PR to the release repo to create the verify job.

/lgtm
/approve

frobware:
> Nice, thanks... guess we'll need a corresponding PR to the release repo to create the verify job.

I did both, obviously need this first.

https://github.com/openshift/release/pull/8384

ironcladlou:
>I did both, obviously need this first.

(sidebar) it's actually a bit of chicken and egg? Since you can't test the release change against this PR, it seems at least possible if you commit this first, PRs after the release merge could fail  because of some error in verify we weren't able to test as part of the release PR...

On t he other hand, if you commit release first, all PRs would be blocked until you have one that successfully passes the new verify job.

Either way we always sort it out, just makes me wonder if there's a way commit this first and then rehearse the release job with e2e tests against master which now contains the verify target?

PR #113, opened 2020-04-15
Update to Kube 1.18 / Go 1.13

ironcladlou:

PR #116, opened 2020-04-17
Run go mod tidy

Miciah:
* `go.mod`:
* `go.sum`:
* `vendor/*`: Regenerate.

----

This should unblock https://github.com/openshift/release/pull/8384.


Miciah:
/close
https://github.com/openshift/release/pull/8384 merged even though the rehearsal jobs were failing, so evidently this PR is not needed.  #113 fixes up `go.mod` and `go.sum`, so let's see if it passes `/test verify` now.

PR #119, opened 2020-04-22
Bug 1825354: Removes ALPN from haproxy frontends

danehans:
Removes `alpn h2,http/1.1` from frontend `fe_sni` and `fe_no_sni`.


danehans:
Closing, this should be targeted for 4.4.

PR #122, opened 2020-04-23
Bug 1821095: Try to promote inactive routes following route deletion

ironcladlou:
Before this patch, inactive routes were not being promoted when the conflicting
routes were deleted. Fix it so that when deletes happen, all inactive routes are
given a chance to be promoted.


pkg/router/controller/hostindex/hostindex.go
```diff
@@ -284,7 +284,7 @@ func (r *hostRules) add(route *routev1.Route, fn RouteActivationFunc, changes *r
 func (r *hostRules) removeActive(i int, fn RouteActivationFunc, changes *routeChanges) {
 	r.active = append(r.active[0:i], r.active[i+1:]...)
 	// attempt to promote all inactive routes
-	if len(r.active) == 0 || i == 0 {
+	if len(r.inactive) > 0 {
```
ironcladlou, line 287: This is the critical change

pkg/router/template/plugin.go
```diff
@@ -112,18 +115,25 @@ func createTemplateWithHelper(t *template.Template) (*template.Template, error)
...
+
+	masterTemplate, err := template.New("config").Funcs(helperFunctions).Parse(templateContents)
 	if err != nil {
 		return nil, err
 	}
+
 	templates := map[string]*template.Template{}
 
 	for _, template := range masterTemplate.Templates() {
-		if template.Name() == templateBaseName {
```
ironcladlou, line 123: TODO: understand and potentially restore this

pkg/router/router_test.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	fmt.Printf("router working directory: %s\n", workdir)
+
+	// The template plugin which is wrapped
+	svcFetcher := templateplugin.NewListWatchServiceLookup(client.CoreV1(), 60*time.Second, namespace)
+	pluginCfg := templateplugin.TemplatePluginConfig{
+		WorkingDir:            workdir,
+		DefaultCertificateDir: workdir,
+		ReloadFn:              func(shutdown bool) error { return nil },
+		Template:              string(haproxy.MustAsset("images/router/haproxy/conf/haproxy-config.template")),
```
ironcladlou, line 97: @openshift/sig-network-edge if nobody wants to switch to using bindata for this generally (i.e. eliminate the template path option entirely), I can revert the bindata stuff and rely on env or something (e.g. default relative to pwd) to configure the existing path variable. I doubt the bindata stuff is worth maintaining just for this test

ironcladlou: I went ahead and rolled it back

ironcladlou:
@Miciah the e2es on this one keep failing for suspiciously consistent reasons which make me wonder if there's a subtle regression here... getting kind of nervous

ironcladlou:
Had to make some additional fixes in c2f1bbd3e696dd30e3a9b070047bb4845d4d24b2 for template output which was only breaking when running in-cluster

pkg/router/router_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+			mustDelete{"b"},
+
+			expectAdmitted{"a", "c", "e"},
+			expectRejected{"d"},
+
+			mustDelete{"c"},
+
+			expectAdmitted{"a", "d", "e"},
+			expectRejected{},
```
Miciah, line 140: Is this intended to assert that there are 0 rejected routes?  It looks like it actually asserts nothing because `(e expectRejected) Apply` will iterate 0 times and return nil.

pkg/router/template/router.go
```diff
@@ -497,16 +466,11 @@ func (r *templateRouter) writeConfig() error {
 
 	log.V(4).Info("router certificate manager config committed")
 
-	pathNames := make([]string, 0)
-	for k := range r.templates {
-		pathNames = append(pathNames, k)
-	}
-	sort.Strings(pathNames)
```
Miciah, line 504: I'm trying to understand why this was added in the first place.  It looks like https://github.com/openshift/origin/pull/19219/commits/1cfd4fde6c6a48c7df9d52fe034cf15424375382 added logic that wrote temporary map files under `/var/lib/haproxy/conf/.tmp/` and then used them to write the final map files under `/var/lib/haproxy/conf/`, so the sort was needed to make sure the `.tmp/` files were written first.  However, https://github.com/openshift/origin/pull/19492/commits/9e88ee03fca132c2aa0f5713031c4cf474d8cc21 removed the temporary files, at which point the sorting was no longer needed.  So this looks safe to remove.

pkg/router/template/router.go
```diff
@@ -375,17 +372,6 @@ func (r *templateRouter) writeDefaultCert() error {
 	return nil
 }
 
-func (r *templateRouter) readState() error {
-	data, err := ioutil.ReadFile(filepath.Join(r.dir, routeFile))
```
Miciah, line 379: Delete the declaration of `routeFile` too.  (We can get this in a follow-up.)

pkg/router/router_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+	namespace := "default"
+
+	h = &harness{
+		routeClient: routeClient,
+		namespace:   namespace,
+	}
+
+	// Other shared junk
+	routerSelection := &routercmd.RouterSelection{}
```
Miciah, line 71: If I understand correctly, this will test `hostindex.SameNamespace` but not `hostindex.OldestFirst`.  Do we test with `DisableNamespaceOwnershipCheck: true` too?  Could be a follow-up.

pkg/cmd/infra/router/template.go
```diff
@@ -150,7 +150,7 @@ func getIntervalFromEnv(name string, defaultValSecs int) time.Duration {
 }
 
 func (o *TemplateRouter) Bind(flag *pflag.FlagSet) {
-	flag.StringVar(&o.WorkingDir, "working-dir", "/var/lib/haproxy/router", "The working directory for the router plugin")
+	flag.StringVar(&o.WorkingDir, "working-dir", "/var/lib/haproxy", "The working directory for the router plugin")
```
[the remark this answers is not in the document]
ironcladlou: Not explicitly, but I will — thanks

pkg/router/router_test.go
ironcladlou: Ah, yeah, I forgot to actually implement that and so the empty assertion is merely cosmetic at the moment. I'll fix it. Thanks!

pkg/router/template/router.go
ironcladlou: Ha, interesting. Yeah, just seeing the code as it stood made no sense and so the refactoring seemed harmless. Thanks for contextualizing it.

pkg/router/router_test.go
ironcladlou: You're correct, this tests only the default configuration (i.e. DisableNamespaceOwnershipCheck=false)

pkg/router/router_test.go
ironcladlou: Removed those empty assertions.

pkg/router/template/router.go
ironcladlou: Removed

danehans:
It looks like commit `de9876b` addresses comments, so it LGTM. I'll let @Miciah tag.

danehans:
I believe @Miciah found an issue with the new directory layout introduced by this PR that requires [certDir and caCertDir](https://github.com/openshift/router/blob/c2f1bbd3e696dd30e3a9b070047bb4845d4d24b2/pkg/router/template/router.go#L37) to be updated.

Miciah:
#126 supersedes this PR, incorporating Dan's work plus some minor fix-ups.

Thanks, Dan, for the fix and the unit tests (which are really quite nice)!

Rats off to ya!

/close

PR #123, opened 2020-04-24
Bug 1826990: Removes ALPN from haproxy frontends

Miciah:


bparees:
bypassing bz process to land this in 4.4.0

PR #125, opened 2020-05-04
Bug 1832352: Enable ALPN http/2 on reencrypt and passthrough routes only

frobware:
This change supports http/2 negotiation for re-encrypt and pass-through routes only. For a re-encrypt route, a custom certificate must be specified to enable http/2 negotiation through the router to the server backend. 

The original diagnosis was made in BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1826990:

> In OpenShift 4.4, we enabled HTTP/2 on the frontend for the ingress controller; that is, we allow clients to ask the ingress controller to use HTTP/2, using ALPN.  Because the console and oauth-server are both behind the ingress controller's balancer and are using the same serving certificate (namely the ingress controller's default certificate), a browser might perform connection coalescing[1], meaning the browser re-uses the connection that it used to connect to the console route to connect to the oauth route.  Because we enabled HTTP/2 on the frontend, this means the browser may connect to the console route using HTTP/2 and then re-use the HTTP/2 connection to try to connect to the oauth route, which fails.

> In general, we cannot support HTTP/2 ALPN on routes that use the default certificate without risk of connection re-use/coalescing causing problems of this nature.  To unblock this issue, we can disable HTTP/2 on the frontend.  Later on, in order to support HTTP/2, we will need a solution that enables HTTP/2 only for routes that have custom certificates (which should prevent browsers from coalescing connections).

Asynchronous to this change we will also look at supporting edge terminated routes.

### DIFF
--- a/pkg/router/template/util/haproxy/map_entry.go
+++ b/pkg/router/template/util/haproxy/map_entry.go
@@ -101,9 +101,7 @@ func generateCertConfigMapEntry(cfg *BackendConfig) *HAProxyMapEntry {
 		}
 		if cfg.HasCertificate {
 			switch cfg.Termination {
-			case routev1.TLSTerminationEdge:
-				fallthrough
-			case routev1.TLSTerminationReencrypt:
+			case routev1.TLSTerminationEdge, routev1.TLSTerminationReencrypt:
 				e.SSLBindConfig = "[alpn h2,http/1.1]"
 			}
 		}
--- a/pkg/router/template/util/haproxy/map_entry.go
+++ b/pkg/router/template/util/haproxy/map_entry.go
@@ -95,10 +95,19 @@ func generateSNIPassthroughMapEntry(cfg *BackendConfig) *HAProxyMapEntry {
 // generateCertConfigMapEntry generates a cert config map entry.
 func generateCertConfigMapEntry(cfg *BackendConfig) *HAProxyMapEntry {
 	if len(cfg.Host) > 0 && (cfg.Termination == routev1.TLSTerminationEdge || cfg.Termination == routev1.TLSTerminationReencrypt) && cfg.HasCertificate {
-		return &HAProxyMapEntry{
+		e := &HAProxyMapEntry{
 			Key:   fmt.Sprintf("%s.pem", cfg.Name),
 			Value: templateutil.GenCertificateHostName(cfg.Host, cfg.IsWildcard),
 		}
+		if cfg.HasCertificate {
+			switch cfg.Termination {
+			case routev1.TLSTerminationEdge:
+				fallthrough
+			case routev1.TLSTerminationReencrypt:
+				e.SSLBindConfig = "[alpn h2,http/1.1]"
+			}
+		}
+		return e
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


pkg/router/template/util/haproxy/map_entry.go
```diff
@@ -95,10 +95,19 @@ func generateSNIPassthroughMapEntry(cfg *BackendConfig) *HAProxyMapEntry {
...
 func generateCertConfigMapEntry(cfg *BackendConfig) *HAProxyMapEntry {
 	if len(cfg.Host) > 0 && (cfg.Termination == routev1.TLSTerminationEdge || cfg.Termination == routev1.TLSTerminationReencrypt) && cfg.HasCertificate {
-		return &HAProxyMapEntry{
+		e := &HAProxyMapEntry{
 			Key:   fmt.Sprintf("%s.pem", cfg.Name),
 			Value: templateutil.GenCertificateHostName(cfg.Host, cfg.IsWildcard),
 		}
+		if cfg.HasCertificate {
+			switch cfg.Termination {
+			case routev1.TLSTerminationEdge:
```
frobware, line 104: @Miciah I guess we need to decide whether we do h2 for 4.5 at all. (TBD)

pkg/router/template/util/haproxy/map_entry_test.go
```diff
@@ -811,7 +811,11 @@ func TestGenerateCertConfigMapEntry(t *testing.T) {
 		}
 
 		certHost := templateutil.GenCertificateHostName(host, wildcard)
-		return &HAProxyMapEntry{Key: key, Value: certHost}
+		e := &HAProxyMapEntry{Key: key, Value: certHost}
+		if hascert {
+			e.SSLBindConfig = "[alpn h2,http/1.1]"
```
frobware, line 816: Should we deduce this from the certificate count in the template itself?

Miciah: This approach looks fine.  Putting the logic in the template would be extra hassle for little benefit.

frobware:
PR https://github.com/openshift/origin/pull/24953 validates the changes here.

pkg/router/template/util/haproxy/map_entry.go
```diff
@@ -95,10 +95,19 @@ func generateSNIPassthroughMapEntry(cfg *BackendConfig) *HAProxyMapEntry {
...
+				fallthrough
+			case routev1.TLSTerminationReencrypt:
```
Miciah, line 106: Better to use a single case:
```go
			case routev1.TLSTerminationEdge, routev1.TLSTerminationReencrypt:
				e.SSLBindConfig = "[alpn h2,http/1.1]"
```

frobware: Done in https://github.com/openshift/router/pull/125/commits/e604cfad701a5e481d34379cfb62647f0f09558f

smarterclayton:
> Later on, in order to support HTTP/2, we will need a solution that enables HTTP/2 only for routes that have custom certificates (which should prevent browsers from coalescing connections).

This actually isn't what i would consider a goal - the real bug here is in oauth, and it's possible that we simply need a different solution than putting it behind the router, because it's not truly designed to work in this fashion.

PR #126, opened 2020-05-04
Bug 1821095: Try to promote inactive routes following route deletion

Miciah:
#### Try to promote inactive routes following route deletion

Before this patch, inactive routes were not being promoted when the conflicting routes were deleted. Fix it so that when deletes happen, all inactive routes are given a chance to be promoted.

#### Fix certificate and whitelist directory paths

Fix the paths for the certificate and CA certificate and whitelist directories.  Commit c2f1bbd3e696dd30e3a9b070047bb4845d4d24b2 changed the router's working directory from `/var/lib/haproxy/router` to `/var/lib/haproxy` but did not update these directory paths, which are relative to the working directory path.

Follow-up to commit c2f1bbd3e696dd30e3a9b070047bb4845d4d24b2.

* `images/router/haproxy/conf/haproxy-config.template`:
* `images/router/nginx/conf/nginx-config.template`: Add `router/` to certificate directory paths.
* `pkg/router/template/router.go` (`certDir`, `caCertDir`): Prepend `router/`.
(`whitelistDir`): New constant with the path for whitelist files.
* `pkg/router/template/template_helper.go` (`generateHAProxyCertConfigMap`): Use `certDir`.
(`generateHAProxyWhiteListFile`): Use `whitelistDir`.
* `pkg/router/template/template_helper_test.go` (`TestGenerateHAProxyCertConfigMap`, `TestGenerateHAProxyMap`): Adjust expected paths (relative to the working directory) for certificates.

#### `writeDefaultCert`: Log error writing default cert

If writing the default certificate out to disk fails, log the error message and use the certificate in the container image.

Before this commit, the router swallowed the error message and used the certificate in the container image on the assumptions that (1) the error could only result from the absence if a configured default certificate and (2) that the absence of a configured default certificate implied that the user did not intend to configure a default certificate.  Logging the error makes it easier for the user to understand what is happening if one or both of these assumptions do not hold true.

* `pkg/router/template/router.go` (`writeDefaultCert`): If writing the default certificate fails, log the error message.  Improve the wording of the log message that indicates that the router is using the certificate from the container image.

---

This is #122 with some fix-ups.

@frobware, @danehans


Miciah:
Rebased.

PR #128, opened 2020-05-07
Bug 1832629: Add metrics for 1xx, 3xx, and 4xx request counters, add latency, track reused connections

### LINKED ISSUE #81: Router metrics don't expose 4xx
Is there a reason the router metrics don't expose any 4xx? 
We actually need those 4xx metrics for debugging and alerting.

smarterclayton:
A number of new metrics have been added to HAProxy that enable better insight into workloads.

The 1xx, 3xx, and 4xx counters will allow total request count to be assessed for non passthrough routes.

The queue and connect latency metrics will give visibility into moving window latency around overload and TLS termination latencies.

The reused connections metric will allow a rough breakdown of which services are experiencing connection churn.

The connections total metric is new and is more accurate than sessions, so use that for frontends.

Builds on top of #127

Fixes #81

### DIFF
--- a/pkg/router/metrics/haproxy/haproxy.go
+++ b/pkg/router/metrics/haproxy/haproxy.go
@@ -143,11 +143,11 @@ type Exporter struct {
 	// is invoked.
 	counterValues counterValuesByMetric
 	// counterIndices records the index in the packed array of metrics (under counterValuesByMetric).
-	// A zero indicates the field index is not a counter, a non-zero value is the index in the packed
-	// array. The first position is always zero
-	// Example: Given counter fields 2, 4, and 5, the array should be:
-	//          []byte{0, 0, 0, 1, 0, 2, 3}
-	//          indicating that position 0 in the packed array is where the second field is stored.
+	// A 0 indicates the field index is not a counter, a positive integer is the index in the packed
+	// array. The first element in the packed array is always zero to avoid having to do index math.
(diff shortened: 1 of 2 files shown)


pkg/router/metrics/haproxy/haproxy.go
```diff
@@ -222,7 +227,7 @@ func NewExporter(opts PrometheusOptions) (*Exporter, error) {
 			43: newFrontendMetric("http_responses_total", "Total of HTTP responses.", prometheus.Labels{"code": "5xx"}),
 			44: newFrontendMetric("http_responses_total", "Total of HTTP responses.", prometheus.Labels{"code": "other"}),
 			48: newFrontendMetric("http_requests_total", "Total HTTP requests.", nil),
-			60: newFrontendMetric("http_average_response_latency_milliseconds", "Average response latency of the last 1024 requests in milliseconds.", nil),
```
jeremyeder, line 225: why?

smarterclayton: Frontends don't expose this metric.

smarterclayton: And never have.

smarterclayton:
The router e2e failure is real, will investigate.

smarterclayton:
Ready for review.

PR #129, opened 2020-05-12
feat(route): Implements rewrite-target for ha-proxy

benoitf:
Annotation is haproxy.router.openshift.io/rewrite-target

Fixes https://github.com/openshift/origin/issues/20474

documentation: https://github.com/openshift/openshift-docs/pull/22021

This feature is also related to this enhancement: https://github.com/openshift/enhancements/pull/287

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -42,6 +42,9 @@
 {{- /* setForwardedHeadersDefaultValue is the default value if a route does not have the setForwardedHeadersAnnotation annotation.  */}}
 {{- $setForwardedHeadersDefaultValue := firstMatch $setForwardedHeadersPattern (env "ROUTER_SET_FORWARDED_HEADERS" "append") "append" -}}
 
+{{- /* pathRewriteTargetPattern: Match path rewrite-Target */}}
+{{- $pathRewriteTargetPattern := `^/.*$` -}}
+
 global
   maxconn {{env "ROUTER_MAX_CONNECTIONS" "20000"}}
 {{- $threads := env "ROUTER_THREADS" }}
@@ -532,6 +535,15 @@ backend {{genBackendNamePrefix $cfg.TLSTermination}}:{{$cfgIdx}}
     {{- end }}
   {{- end}}
 
+  {{- with $pathRewriteTarget := firstMatch $pathRewriteTargetPattern (index $cfg.Annotations "haproxy.router.openshift.io/rewrite-target") }}
+  # Path rewrite target
+    {{- if eq $pathRewriteTarget "/" }}
+  http-request replace-path ^{{ $cfg.Path }}/?(.*)$ {{ $pathRewriteTarget }}\1
+    {{- else }}
+  http-request replace-path ^{{ $cfg.Path }}(.*)$ {{ $pathRewriteTarget }}\1
+    {{- end }}
+  {{- end }}{{/* rewrite target */}}
+  
   {{- if not (isTrue (index $cfg.Annotations "haproxy.router.openshift.io/disable_cookies")) }}
   cookie {{firstMatch $cookieNamePattern (index $cfg.Annotations "router.openshift.io/cookie_name") (env "ROUTER_COOKIE_NAME" "") $cfg.RoutingKeyName}} insert indirect nocache httponly
     {{- if and (matchValues (print $cfg.TLSTermination) "edge" "reencrypt") (ne $cfg.InsecureEdgeTerminationPolicy "Allow") }} secure
--- a/pkg/router/template/configmanager/haproxy/manager.go
+++ b/pkg/router/template/configmanager/haproxy/manager.go
@@ -1104,5 +1104,6 @@ func modAnnotationsList(termination routev1.TLSTerminationType) []string {
 	annotations = append(annotations, "haproxy.router.openshift.io/disable_cookies")
 	annotations = append(annotations, "router.openshift.io/cookie_name")
 	annotations = append(annotations, "haproxy.router.openshift.io/hsts_header")
+	annotations = append(annotations, "haproxy.router.openshift.io/rewrite-target")
 	return annotations
 }


images/router/haproxy/conf/haproxy-config.template
```diff
@@ -439,6 +442,15 @@ backend {{genBackendNamePrefix $cfg.TLSTermination}}:{{$cfgIdx}}
   http-request add-header Forwarded for=%[src];host=%[req.hdr(host)];proto=%[req.hdr(X-Forwarded-Proto)]
   {{- end }}
 
+  {{- with $pathRewriteTarget := firstMatch $pathRewriteTargetPattern (index $cfg.Annotations "haproxy.router.openshift.io/rewrite-target") }}
+  # Path rewrite target
+    {{- if eq $pathRewriteTarget "/" }}
```
elafontaine, line 447: Why is this if required?  

If we could leave it to a simple pattern without side effect, I believe it should help in the long term.  

I may not be understanding why "/" is a special case though.

(Thanks for looking into this :) ).

elafontaine: Is it that we're trying to allow /$1 and / to do the same thing?

benoitf: Hello,
it's to avoid the missing `/` or the double `/` in the final request to the backend server.

elafontaine: https://haproxy-ingress.github.io/docs/configuration/keys/#rewrite-target doesn't seems to consider this a special case.

elafontaine: I think I just understood what it meant... Line 2 and line 5 in the haproxy example seems to consider it a special case to be handle only when you get "//" or "" so that the output is "/".   Those special case would be handled by the defautl behaviour of haproxy (point 1)

In here, we only handle the case where the rewrite path annotation would be "/" so that all traffic to the host only (i.e. http://endpoint) would still be routed correctly to the backend service (i.e. http://endpoint/).  is that it? (Point2)

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -439,6 +442,15 @@ backend {{genBackendNamePrefix $cfg.TLSTermination}}:{{$cfgIdx}}
...
+  http-request replace-path ^{{ $cfg.Path }}/?(.*)$ {{ $pathRewriteTarget }}\1
```
Miciah, line 448: The pattern looks suspicious.  Suppose `$cfg.Path` were `/foo`, `$pathRewriteTarget` were `/bar`, and a client requested `/foobar`; would this rule rewrite the request to `/barbar`?  I would expect the pattern to be something like this: `^{{ $cfg.Path }}(?:/(.*)|$)$`.  Similar applies to the pattern below.

Miciah: > `^{{ $cfg.Path }}(?:/(.*)|$)$`

Could probably be simplified to the following: `^{{ $cfg.Path }}(?:/(.*))?$`.

benoitf: @Miciah about your first remark, if user has `$pathRewriteTarget` set to `/bar` it doesn't go in this if cause but in the else cause

Besides, in your example if client request `/foobar` it will not even reach the backend so we will not use at all this stuff.

it's because `os_http_be.map` file will contain something like
``^mydomain\.com(:[0-9]+)?/foo(/.*)?$     be_http:default:example`

images/router/haproxy/conf/haproxy-config.template
elafontaine: (I'm still wondering about Point2 being right or not for my own understanding XD.  Awaiting your answer anxiously :) )

images/router/haproxy/conf/haproxy-config.template
elafontaine: This is surprising to me as i would have expected the behaviour described by @Miciah to be true ;
Looking at the documentation provided by HAproxy regarding this annotation, one could expect that `/foobar` be replaced with `/barbar` when the annotation is set to `/bar` and the ingress path is `/foo` . (ref. line 5 in the table https://haproxy-ingress.github.io/docs/configuration/keys/#rewrite-target)  

Ingress path | Request path | Rewrite target | Output
/foo | /foobar | /bar | /barbar

I wouldn't expect `/` to be treated differently in the frontend as it's part of the "path" string...  

Now, I'm just giving my opinion from a user perspective... please don't mind it if this is something that should be addressed in the haproxy project... (I wouldn't know where to go to refer this as a potential design issue...)

benoitf: @elafontaine : well if you specify route path being `/foobar` then you can play with annotation on it but if host of the route is only `/foo` then you can only play with `/foo` and `/foo/something`
another app could register a backend service with `/foobar` host

about `/`being handled differently, it's to act like nginx, haproxy and traefik rewrite-target annotations.

elafontaine: so I wouldn't be able to have `/foobar` still be redirected to `/barbar` because it wouldn't even be catched by the ingress resource  as being part of `/foo` ingress path.  Is my understanding right?

elafontaine: (it would make sense when you think of them as directory/file endpoint, but that's not how I was seeing it before :) )

csantanapr:
How would I be able to do this without annotation in the Route resource? 
Meaning what workarounds would you recommend to configure the haproxy using another method?

csantanapr:
We need this feature and looking for ideas on how to do this in a cluster we are using OCP 4.4 on IBM Cloud managed.

csantanapr:
This is for a customer that has a large application running on DIY kubernetes and using an ingress based on Nginx, and now moving to use OpenShift and trying to port the Ingress config to OCP Routes/haproxy

Miciah:
@benoitf, are you able to rebase the PR?

Miciah:
> How would I be able to do this without annotation in the Route resource?
> Meaning what workarounds would you recommend to configure the haproxy using another method?

There isn't a supported way.  The only workaround I can think of at the moment would be to deploy the router using a custom deployment rather than using the operator.

benoitf:
@Miciah rebased

Miciah:
Many thanks!
/lgtm

Miciah:
Weird, the PR is already approved and lgtm'd, but...
/ok-to-test

benoitf:
@miciah it's because I received a report saying that some tests failed :-/

benoitf:
It looks like infra issues for the failing tests

Miciah:
> It looks like infra issues for the failing tests

Yeah, we are having a lot of test flakes.  I'll keep an eye on the failures though.

> @Miciah it's because I received a report saying that some tests failed :-/

Sorry, I was saying that the following message from openshift-ci-robot was weird:

> Cannot trigger testing until a trusted user reviews the PR and leaves an `/ok-to-test` message.

The PR was already lgtm'd, approved, and being tested, so it's funny that the PR was not already "ok-to-test".

Miciah:
Last e2e failure looks related to [BZ1857928 – [sig-operator] an end user use OLM can subscribe to the cockroachdb operator](https://bugzilla.redhat.com/show_bug.cgi?id=1857928).
/test e2e

guanchaoli:
> This is for a customer that has a large application running on DIY kubernetes and using an ingress based on Nginx, and now moving to use OpenShift and trying to port the Ingress config to OCP Routes/haproxy

You could try using another Nginx as a reverse-proxy and write your own nginx.conf to do this. I am also surprised that the latest Openshift on IBM Cloud could not support rewrite-path by annotation.

mhagnumdw:
This feature is available from which version of OpenShift?

Miciah:
It is in OpenShift 4.6.

PR #130, opened 2020-05-13
Bug 1813894: Stop using service ca from service account token

marun:
/cc @sttts @mfojtik @deads2k


Miciah:
/lgtm
/hold
Before we merge this PR, we need to fix the operator to configure the service CA bundle from the proper location.

marun:
@miabbott Should the default be cleared as proposed and the flag/env var be set by the operator to the configmap-sourced file, or should the default be the configmap-sourced file?

I'll leave you to coordinate landing updates to the router and operator, I'm leaving this PR up as a placeholder until your PRs are proposed.

Miciah:
I think you mean to `@` me (Miciah, not Micah).  We should clear the default so that the router can be started without creating the configmap or specifying the environment variable.  Some Origin tests do this (deploy a router pod with some special configuration, without using the operator), and I could imagine developers or community users from expecting the same behavior.

Miciah:
I manually tested a cluster built with openshift/router#130 and openshift/origin#24956 and did not have any test failures.
/lgtm

Miciah:
And, for the record, openshift/cluster-ingress-operator#399 (merged) addresses https://github.com/openshift/router/pull/130#issuecomment-628717649.

Miciah:
ci-operator failure:

    could not initialize namespace: could not create role binding for: rolebindings.rbac.authorization.k8s.io is forbidden: User "system:serviceaccount:ci:ci-operator" cannot create resource "rolebindings" in API group "rbac.authorization.k8s.io" in the namespace "ci-op-zfqqgl57"

/test unit

PR #132, opened 2020-05-14
Bug 1835845: Report all backend metrics for when there are no endpoints

smarterclayton:
When no servers are exposed, backend metrics are needed to see how many requests are making it to a service. If we don't report them users can't see when no pods are up (a crashlooping service for instance would have no metrics at all because no pods are ready). Always report all backend
metrics. Increases cardinality by O(routes), but prometheus is a lot faster now and this should balance out.

### DIFF
--- a/pkg/router/metrics/haproxy/haproxy.go
+++ b/pkg/router/metrics/haproxy/haproxy.go
@@ -495,10 +495,7 @@ func (e *Exporter) collectMetrics(metrics chan<- prometheus.Metric) {
 	for _, m := range e.frontendMetrics {
 		m.Collect(metrics)
 	}
-	for i, m := range e.backendMetrics {
-		if _, ok := e.reducedBackendExports[i]; !e.serverLimited && !ok {
-			continue
-		}
+	for _, m := range e.backendMetrics {
 		m.Collect(metrics)
 	}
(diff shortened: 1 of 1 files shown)


smarterclayton:
Test failure is legitimate - before merging we'd disable that test.

metalmatze:
Nice! Thank you for taking care of this!!!
While average is better than nothing is there any chance of getting something similar to a histogram or things like the p90 or p99?

danehans:
`e2e-aws` failed due to:
```
FAIL: Some cluster operators never became ready: kube-apiserver (Progressing=True NodeInstaller: NodeInstallerProgressing: 1 nodes are at revision 5; 2 nodes are at revision 7)
```
/test e2e-aws

smarterclayton:
Average of what?  Latencies?  haproxy doesn't track that.

danehans:
e2e-aws job failed due to:
`FAIL: Some cluster operators never became ready: kube-apiserver (Progressing=True NodeInstaller: NodeInstallerProgressing: 1 nodes are at revision 6; 2 nodes are at revision 7)`

/test e2e-aws

smarterclayton:
/retest

now that the test has been updated

PR #133, opened 2020-05-20
Bug 1837251: Add unsupported http/2 kill switch

frobware:
This is to support the kill functionality added by:

  https://github.com/openshift/cluster-ingress-operator/pull/401

### DIFF
--- a/pkg/router/template/template_helper.go
+++ b/pkg/router/template/template_helper.go
@@ -175,7 +175,11 @@ func generateHAProxyCertConfigMap(td templateData) []string {
 		backendConfig := backendConfig(string(k), cfg, hascert)
 		if entry := haproxyutil.GenerateMapEntry(certConfigMap, backendConfig); entry != nil {
 			fqCertPath := path.Join(td.WorkingDir, certDir, entry.Key)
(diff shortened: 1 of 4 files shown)


pkg/router/template/template_helper.go
```diff
@@ -175,7 +175,11 @@ func generateHAProxyCertConfigMap(td templateData) []string {
 		backendConfig := backendConfig(string(k), cfg, hascert)
 		if entry := haproxyutil.GenerateMapEntry(certConfigMap, backendConfig); entry != nil {
 			fqCertPath := path.Join(td.WorkingDir, certDir, entry.Key)
-			lines = append(lines, strings.Join([]string{fqCertPath, entry.SSLBindConfig, entry.Value}, " "))
+			if td.DisableHTTP2 {
```
danehans, line 178: @frobware does it make sense to update the `TestGenerateHAProxyMap()` unit test to support this new logic? cc: @Miciah

frobware: If there's a unit test, then definitely.

frobware: Done in https://github.com/openshift/router/pull/133/commits/b0767d17c8b29269a303d75cd0e9c6336a9eb482

PR #135, opened 2020-06-04
add toList helper function

mostafahussein:
This PR adds a new helper called `toList` which can be used when you need to split a specific annotation and then iterate it while configuring HAProxy. e.g removing HTTP headers.


mostafahussein:
@Miciah any update on this pr please?

PR #136, opened 2020-06-05
Bug 1843856: Sanitize TLS config that has key bundled with cert

Miciah:
Modify the extended validation logic to copy any private key found in the certificate field of a route's TLS configuration to its key field.

Extended validation validates a route's TLS configuration using any key that is specified in either the TLS configuration's certificate field or its key field.  As a result, a route that specifies the certificate and key together in the certificate field may pass extended validation and be admitted.  However, before this commit, the certificate manager only wrote the certificate out if the TLS configuration had a nonempty key field.  As a result, a route with a valid certificate and key could be admitted but the certificate and key not written out, which would cause HAProxy to fail to load.

* `pkg/router/routeapihelpers/validation.go` (`splitCertKey`): New function. Take sanitized PEM data and split it into public and private parts.
(`ExtendedValidateRoute`): Use `splitCertKey` to parse out any private parts that are in the certificate field of the TLS configuration.  Prepend any private parts found in the certificate data to the TLS configuration's key.
* `pkg/router/routeapihelpers/validation_test.go`: Add test cases.

### DIFF
--- a/pkg/router/routeapihelpers/validation.go
+++ b/pkg/router/routeapihelpers/validation.go
@@ -122,6 +122,35 @@ func sanitizePEM(data []byte) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+// splitCertKey takes a slice of bytes containing sanitized PEM data and returns
+// two slices of bytes containing PEM data: one slice with the public
+// certificate block or blocks from the input PEM data and one slice with any
+// private key blocks.
+func splitCertKey(data []byte) ([]byte, []byte, error) {
+	var block *pem.Block
+	publicBuf := &bytes.Buffer{}
+	privateBuf := &bytes.Buffer{}
+	for len(data) > 0 {
+		block, data = pem.Decode(data)
+		if block == nil {
+			break
+		}
+		// Because data is sanitized PEM data, the following switch only
+		// needs cases for the block types that sanitizePEM produces.
+		switch block.Type {
+		case "PUBLIC KEY", "CERTIFICATE":
+			if err := pem.Encode(publicBuf, block); err != nil {
+				return nil, nil, err
+			}
+		case "EC PRIVATE KEY", "RSA PRIVATE KEY":
+			if err := pem.Encode(privateBuf, block); err != nil {
+				return nil, nil, err
+			}
+		}
+	}
+	return publicBuf.Bytes(), privateBuf.Bytes(), nil
+}
+
 // ExtendedValidateRoute performs an extended validation on the route
 // including checking that the TLS config is valid. It also sanitizes
 // the contents of valid certificates by removing any data that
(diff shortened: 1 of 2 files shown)


Miciah:
/bugzilla refresh

@openshift/openshift-team-network-edge, please review.

pkg/router/routeapihelpers/validation.go
```diff
@@ -200,6 +216,28 @@ func ExtendedValidateRoute(route *routev1.Route) field.ErrorList {
 		}
 	}
 
+	if len(tlsConfig.Certificate) > 0 {
+		if certBytes, keyBytes, err := splitCertKey([]byte(tlsConfig.Certificate)); err != nil {
+			result = append(result, field.Invalid(tlsFieldPath.Child("certificate"), "redacted key data", err.Error()))
+		} else {
+			if len(tlsConfig.Key) > 0 {
+				if len(keyBytes) > 0 {
+					keyBytes = append(keyBytes, byte('\n'))
```
sgreene570, line 225: I noticed there isn't a unit test that covers this branch. I point this out because I am slightly confused as to what's going on here. If `tlsConfig.Key` is set and `splitCertKey` yields private `keyBytes` from `tlsConfig.Certificate`, then `keyBytes` will become a new-line delimited buffer of both of the keys? Could you elaborate a little bit so I can understand?

pkg/router/routeapihelpers/validation.go
```diff
@@ -122,6 +122,33 @@ func sanitizePEM(data []byte) ([]byte, error) {
...
+		block, data = pem.Decode(data)
+		if block == nil {
+			break
+		}
+		switch block.Type {
+		case "PUBLIC KEY", "CERTIFICATE":
+			if err := pem.Encode(publicBuf, block); err != nil {
+				return nil, nil, err
+			}
+		case "EC PRIVATE KEY", "RSA PRIVATE KEY":
```
sgreene570, line 143: (im probably just out of the loop here, but...) Are EC and RSA the only acceptable types of private keys here?

pkg/router/routeapihelpers/validation.go
Miciah: Right, we'll move any private parts that are in `tlsConfig.Certificate` will get moved to `tlsConfig.Key`, preserving any existing private parts that were already in `tlsConfig.Key`.  I'm not sure whether it would be better just to ignore `keyBytes` if `tlsConfig.Key` is already set, or override `tlsConfig.Key` with `keyBytes` if the latter is nonempty, but I think this is is the safest approach, keeping in mind that we don't want to break existing routes that may be doing weird things in `tlsConfig`.

pkg/router/routeapihelpers/validation.go
Miciah: That's right.  More to the point, `splitCertKey` expects *sanitized* PEM data, meaning PEM data that came from `sanitizePEM`, and `sanitizePEM` (by way of `knownBlockDecoders` and `privateKeyBlockVerifier`) always uses either the block type `"RSA PRIVATE KEY"` or `"EC PRIVATE KEY"` for private keys.

pkg/router/routeapihelpers/validation.go
sgreene570: Understood. I agree this approach sounds like the way to go.

Miciah:
I added some comments, removed a spurious newline that the new logic was inserting when both `tls.Certificate` and `tls.Key` had private key data, and added a test case.

sgreene570:
New changes look good! Thanks for adding the additional unit test.
/lgtm

PR #138, opened 2020-06-05
Use generateCertKey consistently

Miciah:
Fix some lookups in the service alias configuration's certificates map to use `generateCertKey` instead of using the host as the map key.

* `pkg/router/template/router.go` (`hasRequiredEdgeCerts`):
* `pkg/router/template/template_helper.go` (`generateHAProxyCertConfigMap`): Use `generateCertKey` to get the map key for the certificate in the service alias configuration.


sgreene570:
:100: 
/lgtm

PR #141, opened 2020-06-18
Bug 1848687: Ensure router metrics are accurate and increase coverage

smarterclayton:
The backend metrics cover failures like 503 that are not part of server metrics (for instance, when there are no instances of a server available). This PR increases the set of metrics available to administrators at a potentially higher cardinality cost (less than a 2x increase in total series, but in general this is not the largest source of metrics).

The PR also includes a fix to ensure that only metrics which are actually counters (monotonic increasing values that are totals) are treated as counters across restarts. Currently almost all metrics were incorrectly being considered counters.


Miciah:
Are these `go.mod` changes needed too?  https://github.com/openshift/router/pull/127/commits/4aa8e05c8306bdbd87555447bb0cb50be819e278#diff-37aff102a57d3d7b797f152915a6dc16R18-R19

smarterclayton:
Test fixes have to be backported to origin too.

smarterclayton:
Requires https://github.com/openshift/origin/pull/25165 before merge.

shawn-hurley:
Need an LGTM after approving the PR that was needed first

smarterclayton:
go mod changes were automatic as we pulled in tracking code.  I’ll run tidy and verify though

smarterclayton:
So our go.mod was incorrect for 4.4, I moved that into its own commit and locked it to the value prior to the deleted tag.  Should be ready for review now.

PR #146, opened 2020-07-09
Allow configuring unique-id HTTP header

Miciah:
Implement the `ROUTER_UNIQUE_ID_FORMAT` and `ROUTER_UNIQUE_ID_HEADER_NAME` environment variables, which may be used to specify that a header with the given name and format should be injected into every HTTP request.  This header can be used by the application or printed in access logs in order to facilitate tracing requests.

* `images/router/haproxy/conf/haproxy-config.template:` Add `unique-id-format` and `unique-id-header` stanzas using the values from the `ROUTER_UNIQUE_ID_FORMAT` and `ROUTER_UNIQUE_ID_HEADER_NAME` environment variables if the values are nonempty.


sgreene570:
:spiral_notepad: 
/lgtm

PR #139, opened 2020-06-10
Add HTTP header capture

Miciah:
Add `--capture-http-request-headers` and `--capture-http-response-headers` command-line flags, as well as `ROUTER_CAPTURE_HTTP_REQUEST_HEADERS` and `ROUTER_CAPTURE_HTTP_RESPONSE_HEADERS` environment variables, to allow specifying HTTP request or response headers that should be captured for logging.

Each flag or environment variable accepts a comma-delimited list of `name:maxLength` values, where `name` is an HTTP header name and `maxLength` is a maximum length.  The router captures the named headers, which may be referenced in a custom `ROUTER_SYSLOG_FORMAT` string or are included in the default log message format if no custom format string is provided.  A header value that exceeds its maximum length is truncated to that length.

This change is related to [NE-320](https://issues.redhat.com/browse/NE-320).

* `images/router/haproxy/conf/haproxy-config.template`: Add `capture request header` and `capture response header` stanzas to the `public`, `fe_sni`, and `fe_no_sni` frontends.
* `pkg/cmd/infra/router/template.go` (`TemplateRouter`): Add `CaptureHTTPRequestHeadersString`, `CaptureHTTPResponseHeadersString`, `CaptureHTTPResponseHeaders`, and `CaptureHTTPRequestHeaders` fields.
(`Bind`): Add `--capture-http-request-headers` and `--capture-http-response-headers` flags, which default to the values of the `ROUTER_CAPTURE_HTTP_REQUEST_HEADERS` and `ROUTER_CAPTURE_HTTP_RESPONSE_HEADERS` environment variables, respectively.
(`httpHeaderNameRE`): New variable with a regular expression that matches valid HTTP header names.
(`parseCaptureHeaders`): New function.  Parse a string value (such as provided with the newly added command-line flags) into a slice of `CaptureHTTPHeader` values.
(`Complete`): Use `parseCaptureHeaders` to parse `CaptureHTTPRequestHeadersString` and `CaptureHTTPResponseHeadersString` into `CaptureHTTPResponseHeaders` and `CaptureHTTPRequestHeaders`.
(`Run`): Specify `CaptureHTTPRequestHeaders` and `CaptureHTTPResponseHeaders` in the plugin config.
* `pkg/router/template/plugin.go` (`TemplatePluginConfig`): Add `CaptureHTTPRequestHeaders` and `CaptureHTTPResponseHeaders` fields.
[...]

PR #147, opened 2020-07-09
Add HTTP cookie capture

Miciah:
Add a `--capture-http-cookie` command-line flag, as well as `ROUTER_CAPTURE_HTTP_COOKIE` environment variable, to allow specifying an HTTP cookie that should be captured for logging.

The flag and environment variable each accept a `name:maxLength` value, where `name` is an HTTP cookie and `maxLength` is a maximum length.  The router captures the named cookie, which may be referenced in a custom `ROUTER_SYSLOG_FORMAT` string or is included in the default log message format if no custom format string is provided.  A cookie value that exceeds its maximum length is truncated to that length.  The value of the first cookie that matches the provided name captured.  By default, names are compared using a prefix match.  An equals sign ("=") may be appended to the name to indicate that the comparison should use an exact match.

* `images/router/haproxy/conf/haproxy-config.template`: Add `capture cookie` stanzas to the `public`, `fe_sni`, and `fe_no_sni` frontends.
* `pkg/cmd/infra/router/template.go` (`TemplateRouter`): Add `CaptureHTTPCookieString` and `CaptureHTTPCookie` fields.
(`Bind`): Add the `--capture-http-cookie` flag, which defaults to the value of the `ROUTER_CAPTURE_HTTP_COOKIE` environment variable.
(`parseCaptureCookie`): New function.  Parse a string value (such as provided with the newly added command-line flag) into a `CaptureHTTPCookie` value.
(`Complete`): Use `parseCaptureCookie` to parse `CaptureHTTPCookieString` into `CaptureHTTPCookie`.
(`Run`): Specify `CaptureHTTPCookie` in the plugin config.
* `pkg/router/template/plugin.go` (`TemplatePluginConfig`): Add `CaptureHTTPCookie` field.
(`NewTemplatePlugin`): Specify `CaptureHTTPCookie` in the internal template router config.
* `pkg/router/template/router.go` (`templateRouter`):
(`templateRouterCfg`): Add `captureHTTPCookie` field.
(`templateData`): Add `CaptureHTTPCookie` field.
(`newTemplateRouter`): Specify `captureHTTPCookie` in the template router.
(`writeConfig`): Specify `CaptureHTTPCookie` in the template parameters.
* `pkg/router/template/types.go` (`CaptureHTTPCookie`): New type.  Specify an HTTP cookie name, maximum value length, and match type.
(`CookieMatchType`): New type.
(`CookieMatchTypeExact`, `CookieMatchTypePrefix`): New constants.

----

Depends on #139.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -173,6 +173,9 @@ frontend public
   {{- range $idx, $captureHeader := .CaptureHTTPResponseHeaders }}
   capture response header {{ $captureHeader.Name }} len {{ $captureHeader.MaxLength }}
   {{- end }}
+  {{- with $captureCookie := .CaptureHTTPCookie }}
+  capture cookie {{ $captureCookie.Name }}{{ if eq $captureCookie.MatchType "exact" }}={{ end }} len {{ $captureCookie.MaxLength }}
+  {{- end }}
 
   # Strip off Proxy headers to prevent HTTpoxy (https://httpoxy.org/)
   http-request del-header Proxy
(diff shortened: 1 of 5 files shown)


sgreene570:
:cookie: 
/lgtm

PR #153, opened 2020-07-15
Handle errors from commit and reload 

frobware:
Whilst looking at the existing unit tests we have for endpoints (because we plan to switch to endpoint slices) I noticed that the unit tests were not creating the config files when calling `writeConfig()`. Working back through the call chain it highlights that we ignore potential errors from `commitAndReload()`.

This correction now means the unit tests correctly write the config files in the temporary directories. But we may now see errors at runtime as they were previously masked.

I split the commits in two. The first just handles the unit tests, which I think we want anyway. The second commit just handles the case where we were ignoring the error.


sgreene570:
looks good.
/lgtm

PR #158, opened 2020-07-28
pkg/router/: add new counter for http and unix socket fetch totals and failures

RiRa12621:
We are currently missing when haproxy is dead but he pod is alive.
This can happy with a misconfiguration like https://bugzilla.redhat.com/show_bug.cgi?id=1861383

These metrics allow us to get a better inside if the router is at least alive.

The assumption is that all fetches will fail if haproxy is dead and we can calculate an error rate and base alerts of it.
If the stats work, everything should work about haproxy so treating this as internal errors, we can even build an error based SLO out if it.

### DIFF
--- a/pkg/router/metrics/haproxy/haproxy.go
+++ b/pkg/router/metrics/haproxy/haproxy.go
@@ -361,6 +361,23 @@ func (e *Exporter) CollectNow() {
 }
 
 func fetchHTTP(uri string, timeout time.Duration) func() (io.ReadCloser, error) {
+
+	// Create a new counter to count how mnany times we try to fetch stats via http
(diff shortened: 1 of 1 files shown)


RiRa12621:
Before opening this for review it still needs:
* [ ] alerts
* [ ] tests

RiRa12621:
/close
This should already be covered by 
https://github.com/openshift/router/blob/master/pkg/router/metrics/haproxy/haproxy.go#L189-L193
and
https://github.com/openshift/router/blob/master/pkg/router/metrics/haproxy/haproxy.go#L194-L198
and
https://github.com/openshift/router/blob/master/pkg/router/metrics/haproxy/haproxy.go#L1216-L220
which need fixing

PR #165, opened 2020-08-05
Bug 1861455: Remove initial haproxy template commitAndReload

sgreene570:
Waiting for the router factory/controller to call the router plugin's `Commit` function guarantees that route/endpoint resources are available to the template code, and thus removes the possibility of a "routeless" router. A "routeless" router can be detrimental during upgrades if the routes carrying over during the upgrade break a newer version HAProxy, since the "routeless" configuration will continue to run if future successive reloads fail, and the upgrade will succeed when it should not.
The initial (premature) call to the router's `commitAndReload()` function outside of the rate limited loop is older code (3.11 era) and should be removed.  

This PR also adds a failed reload count metric for tracking failed reloads that happen well after a new router pod is created. This will be used for cluster alerting. 

See the linked BZ for more context. 

/assign @Miciah 
/cc @frobware @danehans


Miciah:
Thanks!  This looks terrific!
/lgtm

PR #169, opened 2020-08-11
Bug 1867971: extend unit tests for endpointslices

frobware:
Extends the unit tests to verify the behaviour of adding/deleting endpointslices.

Notable changes:

- Switches the service name label to the upstream definition
- Adds go routines leak checks in the new unit tests


sgreene570:
Verified locally, looks good!

/lgtm

PR #175, opened 2020-09-01
Reload client CA and CRL when updated

Miciah:
#### `watchSecretDir`: Use `log.Error`

* `pkg/router/template/router.go` (`watchSecretDir`): Use `log.Error` instead of `log.Info` when logging errors.

#### `watchSecretDir`: Register change directly

* `pkg/router/template/router.go` (`watchSecretDir`): Simplify by directly registering the change to trigger a router reload.

#### `watchSecretDir`: Improve event logging

* `pkg/router/template/router.go` (`watchSecretDir`): Improve log messages for fsnotify events to indicate that they are coming from fsnotify and to show the operation and file path more clearly.

#### Replace `watchSecretDir` with `watchVolumeMountDir`

* `pkg/router/template/router.go` (`watchSecretDir`): Rename...
(`watchVolumeMountDir`): ...to this.  Factor out the logic specific to the default certificate.  Instead, take a path to a volume mount and a callback function, configure fsnotify to watch the `..data` symlink in the volume mount, and call the function when fsnotify sends an event if the symlink has changed.
(`writeDefaultCert`): Update to use `watchVolumeMountDir`.

#### Reload client CA and CRL when updated

* `pkg/router/template/router.go` (`newTemplateRouter`): Call the new `watchMutualTLSCert` method.
(`watchMutualTLSCert`): New method.  If `ROUTER_MUTUAL_TLS_AUTH_CA` or `ROUTER_MUTUAL_TLS_AUTH_CRL` are set, configure fsnotify watches on the associated volume mount or mounts using `watchVolumeMountDir,` and register the change to trigger a router reload.

----

Related  to https://github.com/openshift/enhancements/pull/462.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -296,16 +296,18 @@ func secretToPem(secPath, outName string) error {
 	return ioutil.WriteFile(outName, pemBlock, 0444)
 }
 
-// watchSecretDir adds a watcher on the input directory that updates the output
-// PEM file when a change is detected.
-func (r *templateRouter) watchSecretDir(secPath, outName string) error {
+// watchVolumeMountDir adds a watcher on path, which should be a secret or
+// configmap volume mount, and calls reloadFn when a change is detected.
+func (r *templateRouter) watchVolumeMountDir(path string, reloadFn func()) error {
 	watcher, err := fsnotify.NewWatcher()
 	if err != nil {
 		return err
 	}
 
-	// We need to know when the content of tls.crt is updated, but tls.crt
-	// is really a symlink with a path that includes another symlink:
+	// Suppose path has the value "/etc/pki/private".  We need to know when
+	// the content of any file in /etc/pki/private is updated, but the files
+	// are really symlinks with a path that includes another symlink.  For
+	// example, a secret might have a tls.crt file:
 	//
 	//     tls.crt -> ..data/tls.crt
 	//     ..data -> ..YY_mm_dd-HH_MM_SS.NN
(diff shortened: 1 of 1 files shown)


Miciah:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_router/175/pull-ci-openshift-router-master-e2e-metal-ipi/1300642425735221248 had no artifacts.  :thinking:

Miciah:
https://github.com/openshift/release/pull/12570 renamed the e2e-metal-ipi job.
/refresh
/skip

pkg/router/template/router.go
```diff
@@ -385,7 +382,17 @@ func (r *templateRouter) writeDefaultCert() error {
 		} else {
 			r.defaultCertificatePath = outPath
 		}
-		if err := r.watchSecretDir(r.defaultCertificateDir, outPath); err != nil {
+		reloadFn := func() {
+			log.V(0).Info("updating default certificate", "path", outPath)
+			os.Remove(outPath)
```
frobware, line 390: What happens if `os.Remove()` is successful but the subsequent write fails in `secretToPem`? Are we in a state where any other subsequent reload of the router will fail (because haproxy barfs with incomplete files and/or data)?

frobware (commented):
Overall LGTM with the proviso that removing files and a subsequent write (in secretToPem) that fails doesn't leave haproxy failing at every subsequent reload.

pkg/router/template/router.go
Miciah: Now that you mention it, we should improve this behavior.  That said, `writeDefaultCert` had the same issue before this change, so I don't think this makes matters worse than before.  What do you think?

frobware: > Now that you mention it, we should improve this behavior. That said, `writeDefaultCert` had the same issue before this change, so I don't think this makes matters worse than before. What do you think?

In a straight run I think things will be generally fine. Just wanted to raise this; we can iterate post this PR merging.

PR #179, opened 2020-09-11
Bug 1752814: Fix decreasing counter metrics when reloading HAProxy

sgreene570:
The HAProxy router [reload process calls](https://github.com/openshift/router/blob/master/pkg/cmd/infra/router/template.go#L539) `CollectNow()` in `pkg/router/metrics/haproxy/haproxy.go`. Currently, `CollectNow()` _does not_ update `e.lastScrape`, which means it is possible for the normal scrape function `Collect(...)` to be called _just after_ `CollectNow()`, but before HAProxy has actually reloaded, since `CollectNow()` essentially "resets" the metrics scrape interval.

The idea in this PR is to have `CollectNow()` set the `e.lastScrape` time, since `CollectNow()` is calling `scrape()` anyways. This will prevent counter metrics from decreasing across reloads since metrics are only saved for reload preservation during `CollectNow()` to minimize memory usage.

### DIFF
--- a/pkg/router/metrics/haproxy/haproxy.go
+++ b/pkg/router/metrics/haproxy/haproxy.go
@@ -355,9 +355,11 @@ func (e *Exporter) CollectNow() {
 	e.mutex.Lock()
 	defer e.mutex.Unlock()
 
+	now := time.Now()
+
 	e.resetMetrics()
 	e.scrape(true)
-	e.lastScrape = nil
+	e.lastScrape = &now
 }
 
 func fetchHTTP(uri string, timeout time.Duration) func() (io.ReadCloser, error) {
--- a/pkg/router/metrics/haproxy/haproxy_test.go
+++ b/pkg/router/metrics/haproxy/haproxy_test.go
@@ -195,7 +195,6 @@ be_edge_http:openshift-console:downloads,BACKEND,0,0,0,0,1,0,0,0,0,0,,0,0,0,0,UP
 	}
 
 	// expect no scrape due to the interval set by the last gather
-	e.lastScrape = &now
 	f = gatherMetrics(t, r)
 	if e.counterValues[secondConsolePodID][e.counterIndices[connectionsTotalIndex]] != 245 {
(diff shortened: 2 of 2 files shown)


frobware:
I would be interested to know whether this changes the behaviour of https://github.com/openshift/origin/pull/25484

sgreene570:
> I would be interested to know whether this changes the behaviour of [openshift/origin#25484](https://github.com/openshift/origin/pull/25484)

Definitely could be the case.
Let me try to reproduce the test failure from https://github.com/openshift/origin/pull/25484 and see if this PR remedies the situation at all.

sgreene570:
Re-triggering e2e test to get an idea if this PR hits https://bugzilla.redhat.com/show_bug.cgi?id=1835371

/test e2e

sgreene570:
gcp quota problems. Will retest later today.

danehans:
@sgreene570 should `pkg/router/metrics/haproxy/haproxy_test.go` be updated? It looks like the test is setting `e.lastScrape` which should be removed due to the changes introduced by this PR, right?

sgreene570:
> @sgreene570 should `pkg/router/metrics/haproxy/haproxy_test.go` be updated? It looks like the test is setting `e.lastScrape` which should be removed due to the changes introduced by this PR, right?

Good point! Thanks for catching this. 
Some of the time in `haproxy_test.go`, `e.lastScrape` is set intentionally to test that the scrape interval is working. That being said, I did find one instance of `e.lastScrape =` that can be removed from the test file in this PR. Thanks!

sgreene570:
```
go: github.com/openshift/library-go@v0.0.0-20200324092245-db2a8546af81 requires
	bitbucket.org/ww/goautoneg@v0.0.0-20120707110453-75cd24fc2f2c: reading https://api.bitbucket.org/2.0/repositories/ww/goautoneg?fields=scm: 404 Not Found
```

Looks like `library-go` needs to be bumped for the router as well. I'll amends this bump to https://github.com/openshift/router/pull/187

sgreene570:
Blocked on https://github.com/openshift/router/pull/187

PR #180, opened 2020-09-11
Bug 1878319: Allow trailing dots in host names

Miciah:
## `generateSNIPassthroughMapEntry`: Match only host

Change the generated `os_sni_passthrough.map` map file to match server names without port numbers or paths.

* `pkg/router/template/util/util.go` (`generateRouteHostRegexp`): New function.  Generate a regular expression just for the route host, for use in other regular expressions.
(`GenerateRouteRegexp`): Use `generateRouteHostRegexp`.
(`GenerateSNIRegexp`): New function.  Use `generateRouteHostRegexp` to generate a regular expression to match route hosts against a server name in a TLS client hello message.
* `pkg/router/template/util/haproxy/map_entry.go` (`generateTCPMapEntry`): Use `GenerateSNIRegexp`.
* `pkg/router/template/util/haproxy/map_entry_test.go` (`TestGenerateSNIPassthroughMapEntry`): Expect keys that match only server names (not port number or path).
* `pkg/router/template/template_helper_test.go` (`TestGenerateHAProxyMap`): Expect `os_sni_passthrough.map` to have patterns that match only server names (not port number or path).

## Allow trailing dots in host names

When generating the pattern for a map entry to match a route host and path, make sure the pattern allows for a trailing dot in the host name.

[RFC 7230, section 5.4](https://tools.ietf.org/html/rfc7230#section-5.4), specifies that the HTTP "host" header value includes the URI host as defined in [RFC 3986, section 3.2.2](https://tools.ietf.org/html/rfc3986#section-3.2.2), which indicates that a trailing dot is permitted.

Note that the same treatment is not required for entries in the certificate map because [RFC 3546, section 3.1](https://tools.ietf.org/html/rfc3546#section-3.1), specifies that the server name in a TLS client hello message does *not* have a trailing dot.

* `pkg/router/template/util/util.go` (`GenerateRouteRegexp`): Match an optional trailing dot in the host name.
* `pkg/router/template/template_helper_test.go` (`TestGenerateHAProxyMap`):
* `pkg/router/template/util/haproxy/map_entry_test.go` (`TestGenerateWildcardDomainMapEntry`, `TestGenerateHttpMapEntry`, `TestGenerateEdgeReencryptMapEntry`, `TestGenerateHttpRedirectMapEntry`, `TestGenerateTCPMapEntry`):
* `pkg/router/template/util/util_test.go` (`TestGenerateRouteRegexp`): Verify that the trailing dot is permitted.


pkg/router/template/util/util.go
```diff
@@ -47,6 +48,22 @@ func GenerateRouteRegexp(hostname, path string, wildcard bool) string {
 	return "^" + hostRE + portRE + pathRE + subpathRE + "$"
 }
 
+// GenerateSNIRegexp generates a regular expression to match route hosts against
+// a server name in a TLS client hello message.
+func GenerateSNIRegexp(hostname string, wildcard bool) string {
```
sgreene570, line 53: Could the common code in `GenerateSNIRegexp` and the first half of `GenerateRouteRegexp` be condensed into one "base" function? I think that would make these changes a little easier to follow along with. Also would bring an added benefit of better unit test coverage :smile:

sgreene570 (commented):
Awesome work. Pointed out one small concern.

pkg/router/template/util/util.go
Miciah: Done.  Thanks for the suggestion!

sgreene570:
Looks great!
/lgtm

PR #183, opened 2020-09-15
[WIP]  DO NOT MERGE- Update to haproxy 1.8.26

miheer:
Update to haproxy 1.8.26

Fixes https://issues.redhat.com/browse/RFE-600


sgreene570:
@miheer, https://issues.redhat.com/browse/RFE-600 is for OCP 3.11, correct? The https://github.com/openshift/router repo contains all OCP 4.x  router code, while the OCP 3.x router code still lives in https://github.com/openshift/origin. Would it make more sense to have a test PR in openshift/origin, instead of here, so the right code and CI tooling is tested?

miheer:
I create a PR in origin https://github.com/openshift/origin/pull/25513

PR #140, opened 2020-06-15
Bug 1852728: haproxy 503 error page: replace all LF by CRLF to be fully RFC compliant

felixkrohn:
[rfc2616](https://tools.ietf.org/html/rfc2616#section-19.3) and [rfc7230](https://tools.ietf.org/html/rfc7230#section-3) both demand that in HTTP messages lines be ended with CRLF.
Although both RFCs stipulate that LF-only line endings MAY or even SHOULD be accepted, It's better to be fully compliant. Very picky HTTP clients, such as web access firewalls, may trip over this and blame Openshift's router.
This PR is whitespace-only and converts all LF line terminators in this file to CRLF.

PR #184, opened 2020-09-15
[release-4.5] Bug 1879120: Backport haproxy 503 error page: replace all LF by CRLF to be fully RFC compliant 

sgreene570:
This PR is the 4.5 backport of https://github.com/openshift/router/pull/140 as per https://bugzilla.redhat.com/show_bug.cgi?id=1879120

---

rfc2616 and rfc7230 both demand that in HTTP messages lines be ended with CRLF.
Although both RFCs stipulate that LF-only line endings MAY or even SHOULD be accepted, It's better to be fully compliant. Very picky HTTP clients, such as web access firewalls, may trip over this and blame Openshift's router.
This PR is whitespace-only and converts all LF line terminators in this file to CRLF.


markmc:
Looks like a super low risk change for something that could cause real-world issues

PR #111, opened 2020-04-06
WIP: Do not merge: Test periodic reaper with the router

mrunalp:
Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

PR #190, opened 2020-09-21
Bug 1859134: Switch to periodic process reaper

sgreene570:
Switch the router process reaper to use the new periodic reaper brought in by openshift/library-go#767 as per https://bugzilla.redhat.com/show_bug.cgi?id=1859134.

This PR bumps `openshift/library-go` to the latest version to accomplish this. Note the function signature change in `pkg/cmd/infra/router/template.go`. 

Previous POC PR for additional context. https://github.com/openshift/router/pull/111

/assign @frobware
/cc @Miciah @danehans


vendor/github.com/openshift/library-go/pkg/proc/proc_linux.go
```diff
@@ -0,0 +1,82 @@
...
+	files, err := ioutil.ReadDir("/proc")
+	if err != nil {
+		return nil, err
+	}
+
+	var zombies []int
+	for _, file := range files {
+		processID, err := strconv.Atoi(file.Name())
+		if err != nil {
+			break
```
Miciah, line 28: How can we be sure there are no file names in procfs that get sorted before numerics?

sgreene570: Seems like there is no guarantee since `ioutil.ReadDir(...)` will [sort by name ](https://golang.org/src/io/ioutil/ioutil.go?s=3208:3259#L90)(letters before numbers). The `Atoi` err check would break out of the loop on any non-numerical named procfs files. I suppose the `files` list could be better sorted first, or the non-numerical filenames could be removed from the list. @Miciah do you bring this up out of time complexity concerns, or just trying to improve the vendored code?

Miciah: I'm wondering whether the vendored code is correct.  For example, if /proc had a file named `.foo`, it would be sorted before the numeric names, and the loop would break before scanning any pids.

Miciah: @mrunalp, am I missing something?  Does some API guarantee exist for procfs that prohibits file names that would be sorted before numeric names?

Miciah:
/lgtm
But let's follow up on https://github.com/openshift/router/pull/190#discussion_r492779293.

sgreene570:
`e2e` test failures appear to be a series of unlucky flakes. Keeping an eye on CI here to make sure this PR isn't breaking any tests.

sgreene570:
/retest

Investigating failures from last night.

sgreene570:
https://github.com/openshift/router/pull/191 to confirm that failures are most likely unrelated to these changes

sgreene570:
/test e2e
--

PR #196, opened 2020-10-07
Bug 1861383: Clip haproxy.router.openshift.io/timeout annotation values to prevent bricking on upgrade

sgreene570:
This PR adds the function `clipHAProxyTimeoutValue(...)` to `pkg/router/template/template_helper.go` to prevent invalid Route timeout annotations from breaking router reloads. In version 2.x, HAProxy supports a maximum timeout of `2147483647ms`, which is ~24.8 days. `clipHAProxyTimeoutValue` will replace any timeout above the maximum amount with the maximum allowable amount to prevent HAProxy from crashing after upgrades that bump HAProxy 1.8 to 2.x.

This PR also adds unit tests cases for `clipHAProxyTimeoutValue` to verify that it works as intended.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -470,7 +470,7 @@ backend {{genBackendNamePrefix $cfg.TLSTermination}}:{{$cfgIdx}}
       {{- end }}
   tcp-request content reject if !whitelist
     {{- end }}
-    {{- with $value := firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")}}
+    {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
   timeout server  {{$value}}
     {{- end }}
 
@@ -642,7 +642,7 @@ backend {{genBackendNamePrefix $cfg.TLSTermination}}:{{$cfgIdx}}
       {{- end }}
   tcp-request content reject if !whitelist
     {{- end }}
-    {{- with $value := firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")}}
+    {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
   timeout tunnel  {{$value}}
     {{- end }}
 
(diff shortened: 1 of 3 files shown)


sgreene570:
Doesnt appear to be working as intended when I verified manually
/hold for now

sgreene570:
I now realize there's a better approach. Pushing a new fix soon.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -463,7 +463,7 @@ backend {{genBackendNamePrefix $cfg.TLSTermination}}:{{$cfgIdx}}
       {{- end }}
   tcp-request content reject if !whitelist
     {{- end }}
-    {{- with $value := firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")}}
+    {{- with $value := clipHAProxyTimeoutValue firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")}}
```
Miciah, line 466:
```suggestion
    {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
```

sgreene570: fixed

frobware:
/retest

Should clear metal-ipi jobs that are not mandatory.

sgreene570:
> Should clear metal-ipi jobs that are not mandatory.

hm, don't think that did the trick.
/refresh maybe?

frobware:
> hm, don't think that did the trick.
> /refresh maybe?

Or close and reopen? I recall we had a similar issue with another PR where the release job criteria had changed.

pkg/router/template/template_helper_test.go
```diff
@@ -767,3 +767,37 @@ func TestProcessEndpointsForAlias(t *testing.T) {
...
+		{
+			value:    "10d",
+			expected: "10d",
+		},
+		{
+			value:    "100d",
+			expected: haproxyMaxTimeout,
+		},
+		{
+			value:    "1000h",
```
frobware, line 793: And a [test] case without any units defaults to milliseconds?

frobware: > And a [test] case without any units defaults to milliseconds?

haproxy defaults to milliseconds: 

```
timeout client <timeout>
timeout clitimeout <timeout> (deprecated)
  Set the maximum inactivity time on the client side.
  May be used in sections :   defaults | frontend | listen | backend
                                 yes   |    yes   |   yes  |   no
  Arguments :
    <timeout> is the timeout value specified in milliseconds by default, but
              can be in any other unit if the number is suffixed by the unit,
              as explained at the top of this document.

```

sgreene570: If no time unit is specific, HAProxy will assume milliseconds.

Added a test to highlight this.

PR #202, opened 2020-10-08
[release-4.4] Bug 1886215: haproxy-config.template: Turn off HTX

Miciah:
Turn off HTX.  Enabling HTX causes HAProxy to down-case HTTP header names, which breaks non-conformant legacy HTTP clients and servers.  HTTP/2 requires HTX, but OpenShift 4.4 does not support HTTP/2.

* `images/router/haproxy/conf/haproxy-config.template`: Turn off HTX.

----

This PR supersedes https://github.com/openshift/router/pull/201.


quarterpin:
/lgtm

Verified via pre-merge verification workflow, more reference related to the test can be found in:
https://bugzilla.redhat.com/show_bug.cgi?id=1886215#c5

PR #209, opened 2020-10-28
Bug 1892338: metrics: Rework template_router_reload_failure metric

sgreene570:
pkg/router/template/router.go: Replace the current template_router_reload_fails counter metric with a new gauge metric, titled template_router_reload_failure, that tracks the result of the most recent HAProxy reload. If a reload fails, the template_router_reload_failure metric will be set to 1 until a successful reload sets the metric to 0.

Previously, the template_router_reload_fails counter metric would monotonically increase when an HAProxy reload failed. This counter metric is difficult to alert on since an increasing counter essentially gives no indication that the reload failures have been resolved.

This new boolean based metric is trivial to properly alert on since the metric will hold "1" if and only if the router is _not_ successfully reloading.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -98,8 +98,8 @@ type templateRouter struct {
 	stateChanged bool
 	// metricReload tracks reloads
 	metricReload prometheus.Summary
-	// metricReloadFails tracks reload failures
-	metricReloadFails prometheus.Counter
+	// metricReloadFailure tracks reload failures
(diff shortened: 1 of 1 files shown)


wking (approved):

Miciah:
Failing with `level=fatal msg=failed to initialize the cluster: Cluster operator cloud-credential is still updating`, which is a known issue.  
/hold

sgreene570:
CI is back!
/hold cancel

Issue #208, opened 2020-10-27
gosec findings


SzekeresB:
Hello!

I just ran `gosec` to scan the go source code statically and got multiple errors. In branch `release-4.3`. My question is do you plan to use gosec during the build or before releases?

```
gosec ./... 
[gosec] 2020/10/27 14:47:47 Including rules: default
[gosec] 2020/10/27 14:47:47 Excluding rules: default
[gosec] 2020/10/27 14:47:47 Import directory: /openshift-router/pkg/version
[gosec] 2020/10/27 14:47:47 Checking package: version
[gosec] 2020/10/27 14:47:47 Checking file: /openshift-router/pkg/version/version.go
[gosec] 2020/10/27 14:47:47 Import directory: /openshift-router/pkg/cmd/infra/router
[gosec] 2020/10/27 14:48:00 Checking package: router
[gosec] 2020/10/27 14:48:00 Checking file: /openshift-router/pkg/cmd/infra/router/clientcmd.go
[gosec] 2020/10/27 14:48:00 Checking file: /openshift-router/pkg/cmd/infra/router/router.go
[gosec] 2020/10/27 14:48:00 Checking file: /openshift-router/pkg/cmd/infra/router/template.go
[gosec] 2020/10/27 14:48:00 Import directory: /openshift-router/pkg/router/controller/factory
[gosec] 2020/10/27 14:48:05 Checking package: factory
[gosec] 2020/10/27 14:48:05 Checking file: /openshift-router/pkg/router/controller/factory/doc.go
[gosec] 2020/10/27 14:48:05 Checking file: /openshift-router/pkg/router/controller/factory/factory.go
[gosec] 2020/10/27 14:48:05 Import directory: /openshift-router/pkg/router/metrics
[gosec] 2020/10/27 14:48:06 Checking package: metrics
[gosec] 2020/10/27 14:48:06 Checking file: /openshift-router/pkg/router/metrics/health.go
[gosec] 2020/10/27 14:48:06 Checking file: /openshift-router/pkg/router/metrics/metrics.go
[gosec] 2020/10/27 14:48:06 Import directory: /openshift-router/pkg/router/template/configmanager/haproxy
[gosec] 2020/10/27 14:48:06 Checking package: haproxy
[gosec] 2020/10/27 14:48:06 Checking file: /openshift-router/pkg/router/template/configmanager/haproxy/backend.go
[...]

PR #210, opened 2020-11-05
Address gosec G601 issues

frobware:
$ gosec -version
Version: 2.5.0
Git tag: v2.5.0
Build date: 2020-10-26T11:52:22Z

On simple inspection the G601 issues from gosec looked innocuous
enough but I decided to clean these up anyway.

For the changes to router.go and template_helper.go I took the
simplest approach of creating a new variable (and copy). I could have
changed everything to be index based but I chose to go with the
smallest change.

PR inspired by https://github.com/openshift/router/issues/208

It still leaves open the question asked in #208 on whether we will run
this tool as during build and/or prior to a release.

### DIFF
--- a/pkg/router/controller/contention.go
+++ b/pkg/router/controller/contention.go
@@ -260,14 +260,12 @@ func ingressConditionTouched(ingress *routev1.RouteIngress) *metav1.Time {
 }
 
 func ingressChanged(oldRoute, route *routev1.Route, routerName string) *routev1.RouteIngress {
-	var ingress *routev1.RouteIngress
 	for i := range route.Status.Ingress {
 		if route.Status.Ingress[i].RouterName == routerName {
-			ingress = &route.Status.Ingress[i]
-			for _, old := range oldRoute.Status.Ingress {
-				if old.RouterName == routerName {
-					if !ingressEqual(ingress, &old) {
-						return ingress
+			for j := range oldRoute.Status.Ingress {
+				if oldRoute.Status.Ingress[j].RouterName == routerName {
+					if !ingressEqual(&route.Status.Ingress[i], &oldRoute.Status.Ingress[j]) {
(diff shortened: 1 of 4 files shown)


frobware:
```

Run multi-stage test e2e-agnostic - e2e-agnostic-gather-audit-logs container test expand_less | 1m1s
-- | --
[must-gather      ] OUT Get "https://api.ci-op-6tjfrnvr-4c798.origin-ci-int-gce.dev.openshift.com:6443/apis/image.openshift.io/v1/namespaces/openshift/imagestreams/must-gather": dial tcp 34.75.225.219:6443: i/o timeout [must-gather      ] OUT  [must-gather      ] OUT Using must-gather plug-in image: registry.redhat.io/openshift4/ose-must-gather:latest Unable to connect to the server: dial tcp 34.75.225.219:6443: i/o timeout error: failed to execute wrapped command: exit status 1
```

/retest

sgreene570:
> On simple inspection the G601 issues from gosec looked innocuous
> enough but I decided to clean these up anyway.

Are there any other gosec codes we should be worried about?

frobware:
> Are there any other gosec codes we should be worried about?

Those were the immediate ones but there's a big list in #208 - some are more interesting than others (e.g., not handling errors from `write()`. I chose to do the memory aliasing as a discrete change.

sgreene570:
Maybe we should consider setting something up like this

https://github.com/openshift/cluster-logging-operator/blob/master/golangci.yaml

frobware:
> Maybe we should consider setting something up like this
> 
> https://github.com/openshift/cluster-logging-operator/blob/master/golangci.yaml

Depends on how many false positives all of those throw up. Often there's too much noise. I used to run https://github.com/openshift/origin/pull/25663 in development to make sure that for at least the lines I touch I don't introduce anything new.

PR #225, opened 2020-11-27
Bug 1900989: Move idle check from endpoints to service

frobware:
As we have moved to endpointslices we now check the service for the
idled annotation. `oc idle` has been updated to annotate the service.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1900989

### DIFF
--- a/pkg/router/template/plugin.go
+++ b/pkg/router/template/plugin.go
@@ -254,42 +254,62 @@ func getPartsFromEndpointsKey(key ServiceUnitKey) (string, string) {
 	return namespace, name
 }
 
+// subsetHasAddresses returns true if subsets has any addresses.
+func subsetHasAddresses(subsets []kapi.EndpointSubset) bool {
+	for i := range subsets {
+		if len(subsets[i].Addresses) > 0 {
+			return true
+		}
+	}
+	return false
+}
+
+// serviceIsIdled return true if the service has been annotated with
+// unidlingapi.IdledAtAnnotation.
+func serviceIsIdled(service *kapi.Service) bool {
+	value, _ := service.Annotations[unidlingapi.IdledAtAnnotation]
+	return len(value) > 0
+}
+
 // createRouterEndpoints creates openshift router endpoints based on k8s endpoints
 func createRouterEndpoints(endpoints *kapi.Endpoints, excludeUDP bool, lookupSvc ServiceLookup) []Endpoint {
 	// check if this service is currently idled
 	wasIdled := false
 	subsets := endpoints.Subsets
-	if _, ok := endpoints.Annotations[unidlingapi.IdledAtAnnotation]; ok && len(endpoints.Subsets) == 0 {
+
+	if !subsetHasAddresses(subsets) {
 		service, err := lookupSvc.LookupService(endpoints)
 		if err != nil {
-			utilruntime.HandleError(fmt.Errorf("unable to find idled service corresponding to idled endpoints %s/%s: %v", endpoints.Namespace, endpoints.Name, err))
+			utilruntime.HandleError(fmt.Errorf("unable to find service %s/%s: %v", endpoints.Namespace, endpoints.Name, err))
 			return []Endpoint{}
 		}
 
-		if !isServiceIPSet(service) {
-			utilruntime.HandleError(fmt.Errorf("headless service %s/%s was marked as idled, but cannot setup unidling without a cluster IP", endpoints.Namespace, endpoints.Name))
-			return []Endpoint{}
-		}
-
-		svcSubset := kapi.EndpointSubset{
-			Addresses: []kapi.EndpointAddress{
-				{
-					IP: service.Spec.ClusterIP,
-				},
-			},
-		}
-
-		for _, port := range service.Spec.Ports {
-			endptPort := kapi.EndpointPort{
-				Name:     port.Name,
-				Port:     port.Port,
-				Protocol: port.Protocol,
+		if serviceIsIdled(service) {
+			if !isServiceIPSet(service) {
+				utilruntime.HandleError(fmt.Errorf("headless service %s/%s was marked as idled, but cannot setup unidling without a cluster IP", endpoints.Namespace, endpoints.Name))
+				return []Endpoint{}
 			}
-			svcSubset.Ports = append(svcSubset.Ports, endptPort)
-		}
 
-		subsets = []kapi.EndpointSubset{svcSubset}
-		wasIdled = true
+			svcSubset := kapi.EndpointSubset{
+				Addresses: []kapi.EndpointAddress{
+					{
+						IP: service.Spec.ClusterIP,
+					},
+				},
+			}
+
+			for _, port := range service.Spec.Ports {
+				endptPort := kapi.EndpointPort{
+					Name:     port.Name,
+					Port:     port.Port,
+					Protocol: port.Protocol,
+				}
+				svcSubset.Ports = append(svcSubset.Ports, endptPort)
+			}
+
+			subsets = []kapi.EndpointSubset{svcSubset}
+			wasIdled = true
+		}
 	}
 
 	out := make([]Endpoint, 0, len(endpoints.Subsets)*4)


frobware:
```
Installing from initial release registry.build02.ci.openshift.org/ci-op-t951pb9z/release@sha256:b0b724e3b077f637528f0ea67b196214e5f661148a7873bbdd2173730a3dc41a
level=info msg=Credentials loaded from environment variable "GOOGLE_CLOUD_KEYFILE_JSON", file "/etc/openshift-installer/gce.json"
level=warning msg=Found override for release image. Please be warned, this is not advised
level=info msg=Consuming Install Config from target directory
level=fatal msg=failed to fetch Cluster: failed to fetch dependency of "Cluster": failed to generate asset "Platform Provisioning Check": baseDomain: Internal error: failed to list DNS Zones: googleapi: Error 429: Resource has been exhausted (e.g. check quota)., rateLimitExceeded
```
/test e2e-upgrade

sgreene570:
> An `oc idle <service>` will annotate the endpoint with an 'idle time'
> annotation. We don't or can't add this annotation to endpointslices in
> openshift/oc (idle) because the underlying endpointslices are managed
> by the kube-controller-manager and that doesn't propagate annotations
> as part of the mirroring service.

Is it at all possible to have `oc idle` pick up the endpointslice for a service and annotate that directly?

frobware:
> Is it at all possible to have `oc idle` pick up the endpointslice for a service and annotate that directly?

This was my first thought too. And this is how I initially implemented it. But it turns out that the endpointslice is managed by the endpoints/endpointslice mirroring service so although you can annotate the endpointslice and, with good luck/timing you do see that delivered, it is soon overwritten by the mirroring service.  I got caught out when testing this because the first few times it worked (as in I received and endpointslice with the expected annotation) but sometimes I was stepping through the behaviour in the debugger and I noticed that sometimes the annotation had disappeared.

pkg/router/controller/factory/factory.go
```diff
@@ -425,3 +436,29 @@ func (f *RouterControllerFactory) createEndpointSliceSharedInformer() {
...
+		return nil, err
+	}
+
+	if endpointExists {
+		ep, ok := item.(*kapi.Endpoints)
+		if !ok {
+			return nil, fmt.Errorf("expecting type %q, got %T", epType, ep)
+		}
+		if val, ok := ep.Annotations[unidling.IdledAtAnnotation]; ok && len(val) != 0 {
+			if objMeta.Annotations == nil {
```
sgreene570, line 456: Worried about dropping any annotations that exist on the endpointslice here. I think the endpointslice mirroring controller applies it's own (and there might be a chance that this code could throw off that controller). Can we just append to the endpointslices current annotations, or return them untouched if we aren't idling?

sgreene570: Ah the annotations are just consumed internally so nevermind, no harm in dropping what we dont care about.

pkg/router/controller/router_controller.go
```diff
@@ -229,6 +229,15 @@ func (c *RouterController) HandleEndpoints(eventType watch.EventType, obj interf
 	c.Commit()
 }
 
+// TODO(frobware) - it this necessary TBD
+func (c *RouterController) HandleEndpointsNamespace(eventType watch.EventType, obj interface{}) {
+	endpoints := obj.(*kapi.Endpoints)
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.RecordNamespaceEndpoints(eventType, endpoints)
```
sgreene570, line 237: Im not sure if this is necessary either... what exactly does this do?

pkg/router/controller/router_controller.go
```diff
@@ -304,6 +313,10 @@ func convertEndpointSliceToEndpointSubset(items []discoveryv1beta1.EndpointSlice
 			}
 		}
 
+		if len(addresses) == 0 {
```
sgreene570, line 316: Is this just an optimization missed in prior endpointslice work?

sgreene570: If this is idle specific, maybe a comment would be helpful?

frobware: I thought I had commented on this. IIRC (and my memory is fuzzy since I last looked at this) this is/was absolutely essential.  Will debug and add a comment.

frobware:
We may be better off carrying https://github.com/kubernetes/kubernetes/pull/98116

frobware:
This is the fix: https://github.com/openshift/openshift-controller-manager/pull/159

frobware:
Needs https://github.com/openshift/oc/pull/720

pkg/router/controller/router_controller.go
frobware: Marking as resolved as we're now checking the service for the idle annotation instead and the logic has moved to `subsetHasAddresses`.

frobware:
> Is it at all possible to have `oc idle` pick up the endpointslice for a service and annotate that directly?

We're now annotating the service; see https://github.com/openshift/oc/pull/720

pkg/router/template/plugin.go
```diff
@@ -254,42 +254,62 @@ func getPartsFromEndpointsKey(key ServiceUnitKey) (string, string) {
...
+			}
 
-		svcSubset := kapi.EndpointSubset{
-			Addresses: []kapi.EndpointAddress{
-				{
-					IP: service.Spec.ClusterIP,
+			svcSubset := kapi.EndpointSubset{
+				Addresses: []kapi.EndpointAddress{
+					{
+						IP: service.Spec.ClusterIP,
```
aojea, line 296: just FYI, with dual stack this will be an slice `service.Spec.ClusterIPs`

pkg/router/template/plugin.go
```diff
@@ -254,42 +254,62 @@ func getPartsFromEndpointsKey(key ServiceUnitKey) (string, string) {
...
+}
+
 // createRouterEndpoints creates openshift router endpoints based on k8s endpoints
 func createRouterEndpoints(endpoints *kapi.Endpoints, excludeUDP bool, lookupSvc ServiceLookup) []Endpoint {
 	// check if this service is currently idled
 	wasIdled := false
 	subsets := endpoints.Subsets
-	if _, ok := endpoints.Annotations[unidlingapi.IdledAtAnnotation]; ok && len(endpoints.Subsets) == 0 {
+
+	if !subsetHasAddresses(subsets) {
```
sgreene570, line 280: If `!subsetHasAddresses(subsets)` evaluates to `true`, does that guarantee that the service is being idled? I ask because the current error wrapper for `lookupSvc.LookupService(endpoints)` seems to assume so. Maybe that error message should be updated to make more sense?

pkg/router/template/plugin.go
```diff
@@ -254,42 +254,62 @@ func getPartsFromEndpointsKey(key ServiceUnitKey) (string, string) {
 	return namespace, name
 }
 
+// subsetHasAddresses returns true if subsets has any addresses.
+func subsetHasAddresses(subsets []kapi.EndpointSubset) bool {
+	for i := range subsets {
+		if len(subsets[i].Addresses) > 0 {
```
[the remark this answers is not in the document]
frobware: Why bother with the copy? It's not zero cost vis-a-vis accessing via the index.

pkg/router/template/plugin.go
frobware: Your suggestion is to drop the word "idled" here?

```go
"unable to find idled service corresponding to idled endpoints %s/%s: %v"
```

I ask because I did that first time, then decided to not to... but it did seem like the natural change at the time.

pkg/router/template/plugin.go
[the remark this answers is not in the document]
sgreene570: Consistency, I guess. But you do bring up a point about the copy, so I don't feel strongly enough to contend.

pkg/router/template/plugin.go
sgreene570: Yea, I think it would make sense to drop the word "idled" in both instances here. Any objections?

frobware: Took a punt in https://github.com/openshift/router/pull/225/commits/b873ba19babcad05914a27ca2b4768b0c26b062e.

I think the notion of idle is irrelevant - I looked up a service and I don't have it.

pkg/router/template/plugin.go
sgreene570: Spoke with @frobware, we can address this in a follow-up

PR #233, opened 2020-12-08
Bug 1903821: Pass stop channel to RouterControllers

sgreene570:
pkg/cmd/infra/router/template.go: Pass the stopCh parameter from the Run function to the RouterControllerFactory.Create calls (as opposed to using a hard-coded wait.NeverStop). This is to ensure that the RouterController(s) are stopped during a router pod's termination process. This prevents router pods from a previous deployment that are in the process of terminating from fighting with newly created router pods over route admission status.

This resolves BZ-1903821, in which toggling the router deployment's ROUTER_DISABLE_NAMESPACE_CHECK can result in inactive routes holding an "Admitted: True" condition.

--- 

See https://github.com/openshift/router/pull/94 for context when the `Run` function in `pkg/cmd/infra/router/template.go` added the `stopCh` parameter.
Also see https://github.com/openshift/router/pull/169, where `factory.Create` was given a stop channel parameter (that was set to `wait.NeverStop` at the time.

### DIFF
--- a/pkg/cmd/infra/router/template.go
+++ b/pkg/cmd/infra/router/template.go
@@ -22,7 +22,6 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/authentication/authenticatorfactory"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
 	"k8s.io/apiserver/pkg/authorization/authorizerfactory"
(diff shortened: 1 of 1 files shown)


sgreene570:
Tagging WIP while I verify that this change doesn't break anything.

pkg/cmd/infra/router/template.go
```diff
@@ -648,11 +647,11 @@ func (o *TemplateRouterOptions) Run(stopCh <-chan struct{}) error {
 	var recorder controller.RejectionRecorder = controller.LogRejections
 	if o.UpdateStatus {
 		lease := writerlease.New(time.Minute, 3*time.Second)
-		go lease.Run(wait.NeverStop)
+		go lease.Run(stopCh)
```
sgreene570, line 650: I think changing lines 650 and 654 to use the `stopCh` makes sense, but someone please correct me if i'm wrong.

Miciah:
#233 merged in time to ship in OpenShift, but #240 did not.  That could cause confusion but shouldn't cause additional problems (other than that the original problem isn't completely fixed in 4.7).

PR #234, opened 2020-12-09
[release-4.5] Bug 1906132: back port: Change router's internal endpoint.ID to prevent HAProxy server line collisions

sgreene570:
HAProxy backend stanzas cannot have duplicate `server` lines. Services with multiple port specs that reference the same target port break this rule when a route is created with an integer target port that matches the (duplicate) service target port. Services that meet this criteria should also be counted as 1 (one) endpoint when calculating service endpoint weights and when determining if an HAProxy backend health probe is applicable. 

---

This is a manual backport of https://github.com/openshift/router/pull/170.


pkg/router/template/plugin.go
```diff
@@ -303,13 +303,13 @@ func createRouterEndpoints(endpoints *kapi.Endpoints, excludeUDP bool, lookupSvc
...
-						ep.ID = fmt.Sprintf("pod:%s:%s:%s:%d", ep.TargetName, endpoints.Name, a.IP, p.Port)
+						ep.ID = fmt.Sprintf("pod:%s:%s:%s:%s:%d", ep.TargetName, endpoints.Name, p.Name, a.IP, p.Port)
 					} else {
-						ep.ID = fmt.Sprintf("ept:%s:%s:%d", endpoints.Name, a.IP, p.Port)
+						ep.ID = fmt.Sprintf("ept:%s:%s:%s:%d", endpoints.Name, p.Name, a.IP, p.Port)
 					}
 				} else {
-					ep.TargetName = ep.IP
-					ep.ID = fmt.Sprintf("ept:%s:%s:%d", endpoints.Name, a.IP, p.Port)
+					ep.TargetName = a.IP
```
Miciah, line 311: Ah, this must be the fuzz that inhibited automated cherry-picking.  Yeah, on release-4.6, `ep.IP` and `a.IP` may be different, and using `a.IP` is the correct thing to do whereas on release-4.5, `ep.IP` and `a.IP` should be the same, so it doesn't matter which you use.  ✓

Miciah (commented):
/lgtm

quarterpin:
/lgtm 

Verified via pre-merge verification workflow, more reference related to the test can be found in:
https://bugzilla.redhat.com/show_bug.cgi?id=1906132#c1

pkg/router/template/plugin.go
sgreene570: You are correct, this was one of the merge conflicts I had to resolve. There were also merge conflicts over `pkg/router/template/template_helper_test.go` since `TestClipHAProxyTimeoutValue` was recently backported.

PR #238, opened 2020-12-11
Bug 1896977: Enhance host name validation in router

candita:
Router host name validation did not properly report an issue with a DNS label over length.  Update it to use a k8s apimachinery host name validation that reports the issue properly, as well as enforcing host name must have at least two labels (new), with each label less than 63 chars, and total length less than 253 chars.  Trailing dots are allowed (new).

Also remove an unneeded/noisy error log in host admitter and correct/add unit tests.


candita:
Companion PR in openshift/apiserver: https://github.com/openshift/openshift-apiserver/pull/167

pkg/router/controller/host_admitter.go
```diff
@@ -143,7 +143,6 @@ func (p *HostAdmitter) HandleRoute(eventType watch.EventType, route *routev1.Rou
 		switch eventType {
 		case watch.Added, watch.Modified:
 			if err := p.addRoute(route); err != nil {
-				log.Error(err, "route not admitted", "routeNameKey", routeNameKey(route))
```
sgreene570, line 146: Was this change accidentally committed, or is this being removed for a specific reason?

candita: This is being removed because it is not needed, the err is returned.  I checked with Miciah about removing it due to its noisiness in the log, and he agreed.

sgreene570:
Looks good to me.
I'll defer to Miciah for final LGTM (in case he has any final comments).

pkg/router/controller/unique_host_test.go
```diff
@@ -67,6 +76,24 @@ func TestValidateHostName(t *testing.T) {
...
+		},
+		{
+			name: "invalid-name-asterisk",
+			route: &routev1.Route{
+				Spec: routev1.RouteSpec{
+					Host: "*",
+				},
+			},
+			expectedErrors: true,
+		},
```
Miciah, line 96: Might as well add a couple test cases with the annotation.

candita: I can add some. There are some test cases in the complement PR: https://github.com/openshift/openshift-apiserver/pull/167/files#diff-e571a492e8cca2251f9208e5274c4c9a4e67368113b1b6f131dd5b685c5eaa56

Miciah: It'd be good to add a test case or two here to get coverage to make sure `ValidateHostName` correctly looks up the annotation and passes it to `ValidateHost`.

Miciah:
Thanks!
/lgtm

PR #239, opened 2020-12-17
Bug 1905100: Add tunnel-timeout and hard-stop-after options to haproxy template

frobware:
The existing annotation "haproxy.router.openshift.io/timeout" affects
"timeout server" for cleartext/edge/reencrypt routes and "timeout
tunnel" for passthrough routes.

If your cleartext/edge/reencrypt route is upgraded to become a
tunnel (e.g., a websocket) then your backend will get the global
default ROUTER_DEFAULT_TUNNEL_TIMEOUT for "timeout tunnel" which is
unexpected and possibly a vastly different value given that the
default is 1 hour.

From the haproxy-2.0 docs:

```
   If some long-lived sessions are mixed with short-lived sessions (e.g.
   WebSocket and HTTP), it's worth considering "timeout tunnel", which
   overrides "timeout client" and "timeout server" for tunnels.

   This timeout supersedes both the client and server timeouts once the
   connection becomes a tunnel.
```

This commit enables a "timeout tunnel" value to be emitted for
cleartext/edge/reencrypt routes.

If you annotate a cleartext/edge/reencrypt route with
"haproxy.router.openshift.io/timeout-tunnel=123s" then the backend
will now be written as follows:

```
  backend be_tcp:default:foo-edge
    balance source
    timeout server  <haproxy.router.openshift.io/timeout> (if set)
    timeout tunnel  123s
    ...
```
For passthrough routes this new annotation will override any existing
"haproxy.router.openshift.io/timeout" annotation.

With only "haproxy.router.openshift.io/timeout=10s" set:
```
  backend be_tcp:default:foo-passthrough
    balance source
    timeout tunnel  10s
    ...
```

But if both are set, for example:

```
  "haproxy.router.openshift.io/timeout=10s"
  "haproxy.router.openshift.io/timeout-tunnel=42s"
```
```
  backend be_tcp:default:foo-passthrough
    balance source
    timeout tunnel  42s
    ...
```
And removing the "haproxy.router.openshift.io/timeout-tunnel"
annotation takes us back to:
```
  backend be_tcp:default:foo-passthrough
    balance source
    timeout tunnel  10s
    ...
```
The alternative was not to introduce a new annotation but emit the
same value for both "tunnel server" and "tunnel timeout" for
cleartext/edge/reencrypt routes but that's likely to be a breaking
change given how long the current arrangement has been in effect.

Related BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1905100

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -46,6 +46,9 @@
 {{- $pathRewriteTargetPattern := `^/.*$` -}}
 
 global
+{{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (env "ROUTER_HARD_STOP_AFTER")) }}
+  hard-stop-after {{$value}}
+{{- end}}
   maxconn {{env "ROUTER_MAX_CONNECTIONS" "20000"}}
 {{- $threads := env "ROUTER_THREADS" }}
 {{- if ne "" (firstMatch "[1-9][0-9]*" $threads) }}
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -486,6 +486,9 @@ backend {{genBackendNamePrefix $cfg.TLSTermination}}:{{$cfgIdx}}
     {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
   timeout server  {{$value}}
     {{- end }}
+    {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel")) }}
+  timeout tunnel  {{$value}}
+    {{- end }}
 
     {{- if isTrue (index $cfg.Annotations "haproxy.router.openshift.io/rate-limit-connections") }}
   stick-table type ip size 100k expire 30s store conn_cur,conn_rate(3s),http_req_rate(10s)
@@ -655,7 +658,7 @@ backend {{genBackendNamePrefix $cfg.TLSTermination}}:{{$cfgIdx}}
       {{- end }}
   tcp-request content reject if !whitelist
     {{- end }}
-    {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
+    {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel") (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
   timeout tunnel  {{$value}}
     {{- end }}
 


frobware:
/hold 

Need to discuss if we want a new annotation or whether we simply emit "timeout tunnel" in addition for cleartext/edge/reencrypt routes.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -655,8 +658,14 @@ backend {{genBackendNamePrefix $cfg.TLSTermination}}:{{$cfgIdx}}
...
-    {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
+    {{- if (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel") }}
+      {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel")) }}
   timeout tunnel  {{$value}}
+      {{- end }}
+    {{- else }}
+      {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
+  timeout tunnel  {{$value}}
+      {{- end }}
     {{- end }}
```
Miciah, line 669: I believe this could be simplified slightly by deleting the first `{{ if }}` and raising the nested `{{ with }}`:
```go
    {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel")) }}
  timeout tunnel  {{$value}}
    {{- else }}
  timeout tunnel  {{$value}}
      {{- end }}
    {{ -end }}
```
(Unfortunately, there is no `{{ else with ... }}` according to <https://golang.org/pkg/text/template/>, which would facilitate further simplification.)

Miciah (commented):
One small suggestion, but
/lgtm
if you want to cancel the hold and merge it as is.

images/router/haproxy/conf/haproxy-config.template
Miciah: Actually, the following would be even simpler:
```go
    {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel") (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
  timeout tunnel  {{$value}}
    {{ -end }}
```

frobware: Ah, great. Prefer the combined/simpler version. Thanks!

frobware:
@Miciah do we want to roll the `hard-stop-after` into this PR as it part of the same bug fix and is also a template change?

Miciah:
> @Miciah do we want to roll the `hard-stop-after` into this PR as it part of the same bug fix and is also a template change?

You mean adding a `ROUTER_HARD_STOP_AFTER` environment variable?  If adding it in this PR easier, I don't object to doing so.

frobware:
> You mean adding a `ROUTER_HARD_STOP_AFTER` environment variable? If adding it in this PR is easier, I don't object to doing so.

Specifically the addition to the template https://github.com/openshift/router/pull/241#discussion_r553810226.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -46,6 +46,9 @@
 {{- $pathRewriteTargetPattern := `^/.*$` -}}
 
 global
+{{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (env "ROUTER_HARD_STOP_AFTER")) }}
```
sgreene570, line 49: This is minor but for a future cleanup, we should probably change the error messages and code comments for `clipHAProxyTimeoutValue` to make sense for any timeout value generically (rather than just route timeout annotations).
https://github.com/openshift/router/blob/master/pkg/router/template/template_helper.go#L307

PR #240, opened 2021-01-06
Bug 1903821: contention: Contend over route ingress admission condition 

sgreene570:
**contention: Contend over route ingress admission condition **

pkg/router/controller/contention.go: Enhance `ingressEqual` to compare
the given route ingress' admission condition, ignoring transition
times. This is to remedy situations where router replicas disagree
explicitly on an ingress admission condition for a route when no changes
to a route's spec or status (excluding status.conditions) are made,
such as when an ingress controller switches from a `namespaceOwnership`
policy of `InterNamespaceAllowed` to `Strict`.

vendor: pull in go-cmp/cmp/cmpopts source.
(go-cmp is already in go.mod)

### DIFF
--- a/pkg/router/controller/contention.go
+++ b/pkg/router/controller/contention.go
@@ -7,6 +7,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/cache"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
 	routev1 "github.com/openshift/api/route/v1"
 )
 
@@ -243,7 +246,10 @@ func (t *SimpleContentionTracker) Clear(id string, current *routev1.RouteIngress
 }
 
 func ingressEqual(a, b *routev1.RouteIngress) bool {
-	return a.Host == b.Host && a.RouterCanonicalHostname == b.RouterCanonicalHostname && a.WildcardPolicy == b.WildcardPolicy && a.RouterName == b.RouterName
+	// In addition to the RouteIngress' string fields, compare the available admission condition to determine
+	// if the given ingress' are equal. See https://bugzilla.redhat.com/show_bug.cgi?id=1908389.
+	return a.Host == b.Host && a.RouterCanonicalHostname == b.RouterCanonicalHostname && a.WildcardPolicy == b.WildcardPolicy && a.RouterName == b.RouterName &&
+		cmp.Equal(findCondition(a, routev1.RouteAdmitted), findCondition(b, routev1.RouteAdmitted), cmpopts.IgnoreFields(routev1.RouteIngressCondition{}, "LastTransitionTime"))
 }
 
 func ingressConditionTouched(ingress *routev1.RouteIngress) *metav1.Time {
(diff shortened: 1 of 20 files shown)


sgreene570:
@Miciah @frobware are there any fundamental issues with this approach? Please TAL when you can.

pkg/router/controller/contention.go
```diff
@@ -243,7 +246,25 @@ func (t *SimpleContentionTracker) Clear(id string, current *routev1.RouteIngress
 }
 
 func ingressEqual(a, b *routev1.RouteIngress) bool {
-	return a.Host == b.Host && a.RouterCanonicalHostname == b.RouterCanonicalHostname && a.WildcardPolicy == b.WildcardPolicy && a.RouterName == b.RouterName
+	return a.Host == b.Host && a.RouterCanonicalHostname == b.RouterCanonicalHostname && a.WildcardPolicy == b.WildcardPolicy && a.RouterName == b.RouterName &&
+		cmp.Equal(findMostRecentIngressCondition(a), findMostRecentIngressCondition(b), cmpopts.IgnoreFields(routev1.RouteIngressCondition{}, "LastTransitionTime"))
```
Miciah, line 250: Would checking the `Admitted` condition be sufficient?  It seems simpler and easier to reason about.  
```suggestion
		cmp.Equal(findCondition(a, routev1.RouteAdmitted), findCondition(b, routev1.RouteAdmitted), cmpopts.IgnoreFields(routev1.RouteIngressCondition{}, "LastTransitionTime"))
```

sgreene570: Yes, that's definitely the preferable approach (for some reason I didn't realize that a route ingress can only have one _admission_ condition). Fixed, thanks!

sgreene570:
Need to verify that the newest changes still resolve the bug (they should, I just want to double check so the BZ doesn't fail QA).
/hold

sgreene570:
Launcher a cluster-bot cluster with this PR and the new changes look good!
/hold cancel
@Miciah PTAL when you get the chance.

Miciah:
Sorry for the delay!
/lgtm

sgreene570:
> Sorry for the delay!
> /lgtm

Thanks! :grin:

Miciah:
#233 merged in time to ship in OpenShift, but #240 did not.  That could cause confusion but shouldn't cause additional problems (other than that the original problem isn't completely fixed in 4.7).

pkg/router/controller/contention.go
```diff
@@ -243,7 +246,10 @@ func (t *SimpleContentionTracker) Clear(id string, current *routev1.RouteIngress
 }
 
 func ingressEqual(a, b *routev1.RouteIngress) bool {
-	return a.Host == b.Host && a.RouterCanonicalHostname == b.RouterCanonicalHostname && a.WildcardPolicy == b.WildcardPolicy && a.RouterName == b.RouterName
+	// In addition to the RouteIngress' string fields, compare the available admission condition to determine
+	// if the given ingress' are equal. See https://bugzilla.redhat.com/show_bug.cgi?id=1908389.
```
Miciah, line 250: Whoops, this was presumably supposed to be https://bugzilla.redhat.com/show_bug.cgi?id=1903821.

PR #241, opened 2021-01-06
[WIP] - PR 239 plus hard stop after -- DO NOT MERGE

frobware:
This is like PR #239 but also adds the "hard-stop-after" option if the environment variable `ROUTER_HARD_STOP_AFTER <duration>` is present. The purpose of this PR is a quick sanity check through CI before dropping this onto a customer cluster.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -46,6 +46,9 @@
 {{- $pathRewriteTargetPattern := `^/.*$` -}}
 
 global
+{{- with (env "ROUTER_HARD_STOP_AFTER") }}
+  hard-stop-after {{env "ROUTER_HARD_STOP_AFTER"}}
+{{- end}}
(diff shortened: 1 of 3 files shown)


images/router/haproxy/conf/haproxy-config.template
```diff
@@ -46,6 +46,9 @@
 {{- $pathRewriteTargetPattern := `^/.*$` -}}
 
 global
+{{- with (env "ROUTER_HARD_STOP_AFTER") }}
+  hard-stop-after {{env "ROUTER_HARD_STOP_AFTER"}}
```
Miciah, line 50: Does the same range apply as for other time values?  Let's validate the value using something like `{{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (env "ROUTER_HARD_STOP_AFTER")) }}`.

frobware:
This is being done as part of PR #239.

PR #243, opened 2021-01-13
Bug 1905748: Prevent unnecessary reloads in router shards

frobware:
If the router instance is sharded then ensure that unrelated endpoint
changes (add/delete) don't cause the router to reload unnecessarily.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -588,15 +588,14 @@ func (r *templateRouter) FilterNamespaces(namespaces sets.String) {
 		r.serviceUnits = make(map[ServiceUnitKey]ServiceUnit)
 		r.stateChanged = true
 	}
-	for k := range r.serviceUnits {
+	for key, service := range r.serviceUnits {
 		// TODO: the id of a service unit should be defined inside this class, not passed in from the outside
 		//   remove the leak of the abstraction when we refactor this code
-		ns, _ := getPartsFromEndpointsKey(k)
+		ns, _ := getPartsFromEndpointsKey(key)
 		if namespaces.Has(ns) {
 			continue
 		}
-		delete(r.serviceUnits, k)
-		r.stateChanged = true
+		r.deleteServiceUnitInternal(key, service)
 	}
 
 	for k := range r.state {
@@ -634,7 +633,6 @@ func (r *templateRouter) createServiceUnitInternal(id ServiceUnitKey) {
 	}
 
 	r.serviceUnits[id] = service
-	r.stateChanged = true
 }
 
 // findMatchingServiceUnit finds the service with the given id - internal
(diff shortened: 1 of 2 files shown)


frobware:
/hold

Need to update unit tests (they will currently fail)
I also want to see this approach go through CI

pkg/router/template/router.go
```diff
@@ -832,7 +832,13 @@ func (r *templateRouter) DeleteEndpoints(id ServiceUnitKey) {
 
 	r.serviceUnits[id] = service
 
-	r.stateChanged = true
+	for alias := range service.ServiceAliasAssociations {
+		if _, exists := r.state[alias]; exists {
+			r.stateChanged = true
+			break
+		}
+	}
```
Miciah, line 840: Would
```suggestion
	r.stateChanged = len(frontend.ServiceAliasAssociations) != 0
```
suffice?  That is, is it sufficient to check if there are any associations, without checking that the associations are with things that exist?

Miciah (commented):
Should `createServiceUnitInternal` be modified as well?  It sets `r.stateChanged = true` even though it initializes `ServiceAliasAssociations` to an empty map.

frobware:
some debug support for verifying runtime behaviour & sanity testing: 

  https://github.com/frobware/haproxy-hacks/tree/master/BZ1905748

pkg/router/template/router_test.go
```diff
@@ -161,6 +193,23 @@ func TestDeleteEndpoints(t *testing.T) {
 		t.Errorf("Unable to find serivce unit %s after creation", suKey)
 	}
 
+	// Deleting endpoints without an associated route won't result
+	// in a state change so create an associated route for the
+	// service unit.
+	router.AddRoute(&routev1.Route{
```
sgreene570, line 199: Could you add a similar route addition to `TestDeleteServiceUnit()` so your changes in this PR to `DeleteServiceUnit()` are covered by unit tests?

frobware:
```
E0119 09:26:10.942709   36633 test_context.go:485] Failed to setup provider config for "gce": Error building GCE/GKE provider: unexpected response listing zones: googleapi: Error 403: Quota exceeded for quota group 'ListGroup' and limit 'List requests per 100 seconds' of service 'compute.googleapis.com' for consumer 'project_number:1053217076791'., rateLimitExceeded

```

pkg/router/template/router.go
```diff
@@ -657,13 +655,22 @@ func (r *templateRouter) DeleteServiceUnit(id ServiceUnitKey) {
...
+	service, ok := r.findMatchingServiceUnit(id)
 	if !ok {
 		return
 	}
 
+	r.DeleteServiceUnitInternal(id, service)
+}
+
+// DeleteServiceUnitUnitInternal deletes the service with the given
+// id. It differs from DeleteServiceUnitInternal() as it assumes the
```
sgreene570, line 667:
```suggestion
// DeleteServiceUnitInternal deletes the service with the given
// id. It differs from DeleteServiceUnit() as it assumes the
```

pkg/router/template/router.go
```diff
@@ -657,13 +655,22 @@ func (r *templateRouter) DeleteServiceUnit(id ServiceUnitKey) {
...
+// caller has taken the lock.
+func (r *templateRouter) DeleteServiceUnitInternal(id ServiceUnitKey, service ServiceUnit) {
```
Miciah, line 669: Can `DeleteServiceUnitInternal` be private?

frobware: > Can `DeleteServiceUnitInternal` be private?

No. But only because of the package naming, as we discovered yesterday.

Miciah: > > Can `DeleteServiceUnitInternal` be private?
> 
> No. But only because of the package naming, as we discovered yesterday.

I thought that was only true of `DeleteServiceUnit`.  It doesn't look like `DeleteServiceUnitInternal` is used outside of the file in which it is defined.

frobware: Fixed in https://github.com/openshift/router/pull/243/commits/922203ce01bba78a2f07275c81cc29c35a85d568

frobware:
-  Using must-gather plug-in image: registry.redhat.io/openshift4/ose-must-gather:latest
- Unable to connect to the server: dial tcp: lookup api.ci-op-xwwr34td-4c798.origin-ci-int-gce.dev.openshift.com on 172.30.0.10:53: no such host

/retest

PR #258, opened 2021-02-10
Bug 1925697: Route admission should not use the override

candita:
The override annotation was supposed to allow a route to be created, but not admitted, if it had an invalid host name. However, use of the override now allows the route to be admitted with an invalid host name. Fix this so the route will not be admitted with an invalid host name.


Miciah:
Thanks!
/lgtm

PR #261, opened 2021-02-19
template: Drop http-use-htx option

sgreene570:
Commit 8580700 bumped the HAProxy version of the router
to v2.2. In HAProxy 2.2, the `http-use-htx` option is deprecated.

Commit da09224 added configurable HTTP header case names to the router.
With commits da09224 and 8580700 in place, we can safely remove the
`http-use-htx` option (and thus silence any HAProxy warnings about it's
deprecation).

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -147,8 +147,6 @@ defaults
   compression type {{env "ROUTER_COMPRESSION_MIME" "text/html text/plain text/css"}}
 {{- end }}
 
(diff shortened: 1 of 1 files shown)


sgreene570:
This will resolve the following router pod log error in 4.8 nightlies:

```
I0219 17:25:18.023556       1 router.go:578] template "msg"="router reloaded"  "output"="[WARNING] 049/172517 (123) : parsing [/var/lib/haproxy/conf/haproxy.config:52]: option 'http-use-htx' is deprecated and ignored. The HTX mode is now the only supported mode.\n - Proxy protocol on, checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
```

PR #262, opened 2021-02-24
Add support for IPv6 CIDR

mic4ael:
This adds support for IPv6 in `haproxy.router.openshift.io/ip_whitelist` annotation set on `Route` objects.


mic4ael:
Any feedback on this?

sgreene570:
> Any feedback on this?

Hey @mic4ael ! Thanks for creating this PR!

We actually just received an internal bug report along the lines of what you're solving here. Somebody from the team should be able to take a look at this new regex in the near future. Thank you for your patience.

alebedev87:
Hello @mic4ael!
Thank you for your PR! It was useful during the analysis.
However we decided to take a slightly different approach: implement a golang helper function which would parse the annotation contents. This way we have something which is more testable and slightly more approachable than a long regexp.

If you are interrrested in the details, you can take a look at https://github.com/openshift/router/pull/307.

PR #266, opened 2021-03-04
Bug 1909906: Exit gracefully if metrics listener fails

Miciah:
Before this PR, the metrics listener code panicked on errors.  This PR changes the listener code to log the error and shut down the router gracefully when an error is encountered.

* `pkg/router/shutdown/shutdown.go` (`RequestShutdown`): New function.  Trigger shutdown by sending a the shutdown handler a fake signal.
* `pkg/router/metrics/metrics.go` (`Listen`): Log errors, and use the new `RequestShutdown` function to initiate graceful shutdown if the metrics listener fails, instead of panicking on errors.

### DIFF
--- a/pkg/router/metrics/metrics.go
+++ b/pkg/router/metrics/metrics.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/apiserver/pkg/authorization/authorizer"
 
 	logf "github.com/openshift/router/log"
+	"github.com/openshift/router/pkg/router/shutdown"
 )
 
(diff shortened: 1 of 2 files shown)


Miciah:
level=fatal msg=failed to initialize the cluster: Working towards 4.8.0-0.ci.test-2021-03-04-025102-ci-op-r7zt2z07: 667 of 668 done (99% complete), waiting on monitoring

/retest

smarterclayton:
does it exit with a non zero exit code to indicate a problem happened?

PR #268, opened 2021-03-11
Bug 1937972: router/template: Cache compiled regular expressions

frobware:
This is an optimisation that caches compiled regular expressions.

When the template is written calls to `matchString` and `firstMatch`
would recompile the regular expression on every call. When you are
processing ~13,000 routes the time to compile on each invocation becomes
significant. This change brings a reduction of 60% when processing and
writing `haproxy.config`.

```console
router.go:456] template "msg"="writeConfig"  "duration"="10.663160501s"
router.go:456] template "msg"="writeConfig"  "duration"="10.328840858s"
router.go:456] template "msg"="writeConfig"  "duration"="10.762876672s"
```

And if we cache the REs:

```console
template "msg"="writeConfig"  "duration"="3.659229627s"
template "msg"="writeConfig"  "duration"="3.718457818s"
template "msg"="writeConfig"  "duration"="3.672131941s"
```

Related to: https://bugzilla.redhat.com/show_bug.cgi?id=1929821

### DIFF
--- a/pkg/router/template/template_helper.go
+++ b/pkg/router/template/template_helper.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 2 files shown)


knobunc (approved):

knobunc:
Looks good to me!  But waiting for a second reviewer.

pkg/router/template/template_helper.go
```diff
@@ -80,7 +105,7 @@ func matchValues(s string, allowedValues ...string) bool {
 
 func matchPattern(pattern, s string) bool {
 	log.V(7).Info("matchPattern called", "pattern", pattern, "s", s)
-	status, err := regexp.MatchString(`\A(?:`+pattern+`)\z`, s)
+	status, err := matchString(`\A(?:`+pattern+`)\z`, s)
```
Miciah, line 108: `matchPattern` isn't used in the template, so this shouldn't affect performance (but it shouldn't hurt anything either).

pkg/router/template/template_helper.go
```diff
@@ -31,9 +32,33 @@ func isTrue(s string) bool {
 	return v
 }
 
+var compiledRegexp sync.Map
```
Miciah, line 35: I was wondering if we needed garbage collection for this map.  I believe we do not as every use of `firstMatch` specifies a pattern that is a hard-coded in the template itself (as opposed to specifying a pattern from an annotation, for example), so the size of the map is bounded by a function of the static set of patterns that are used in `haproxy-config.template`.

pkg/router/template/template_helper.go
```diff
@@ -31,9 +32,33 @@ func isTrue(s string) bool {
...
+
+func cachedRegexpCompile(pattern string) (*regexp.Regexp, error) {
```
Miciah, line 37: Would you mind adding some godoc?

Miciah (commented):
This is terrific.  
/lgtm
/hold
in case you want to wait for further reviews or do any follow-up work such as comments or adding a BZ reference.

sgreene570:
/retitle Bug 1937972: router/template: Cache compiled regular expressions

pkg/router/template/template_helper.go
frobware: This was explicit as `MatchString` compiles a regexp too:

```go
// MatchString reports whether the string s
// contains any match of the regular expression pattern.
// More complicated queries need to use Compile and the full Regexp interface.
func MatchString(pattern string, s string) (matched bool, err error) {
	re, err := Compile(pattern)
	if err != nil {
		return false, err
	}
	return re.MatchString(s), nil
}
```

pkg/router/template/template_helper.go
frobware: Force-pushed.

pkg/router/template/template_helper.go
Miciah: I mean, this entire function is effectively dead code anyway, unless someone is using a custom template that uses `matchPattern`.  That said, it is possible that someone _is_ using a custom template that uses `matchPattern`, so we might as well optimize it.

pkg/router/template/template_helper.go
```diff
@@ -31,9 +32,45 @@ func isTrue(s string) bool {
...
+			return nil, err
+		}
+		compiledRegexp.Store(pattern, re)
+		return re, nil
+	}
+	return v.(*regexp.Regexp), nil
+}
+
+func matchString(pattern string, s string) (bool, error) {
+	status, err := regexp.MatchString(`\A(?:`+pattern+`)\z`, s)
```
Miciah, line 62: What's this line for?  Copy and paste error?

frobware: Errr. yes. Removing. :)

PR #273, opened 2021-03-25
Bug 1942534: [4.7 backport] Route gets admitted with invalid host name if override annotation is used

candita:
The override added to address https://bugzilla.redhat.com/show_bug.cgi?id=1896977 was supposed to allow a route to be created, but not admitted, if it had an invalid host name.  However, it allows the route to be admitted with an invalid host name.  This fixes that.


sgreene570:
@candita could you amend this PR's commit title to reflect the 4.7 bug number?

candita:
> @candita could you amend this PR's commit title to reflect the 4.7 bug number?

Done.

PR #276, opened 2021-03-30
haproxy-config.template: Re-indent.

Miciah:
Use consistent indentation for the HAProxy configuration template.  Note that the HAProxy configuration stanzas and the Go template stanzas are interpolated but have different indentation, which can be confusing but should be less confusing if the Go template indentation is consistent.

* `images/router/haproxy/conf/haproxy-config.template`: Re-indent Go template
actions.

---

I used [a quick-and-dirty AWK script ](https://gist.github.com/Miciah/8a2d5245ba8150a626acb16a16926021) to re-indent the template.  If this seems like a good idea, we could add some automated verification to make sure the Go template stanzas are consistently indented.


sgreene570:
This is terrific!

PR #282, opened 2021-05-11
Bug 1947478: Switch from discovery v1beta1 to discovery v1

sgreene570:
Discovery.k8s.io/v1beta1 EndpointSlices are deprecated in favor of
discovery.k8s.io/v1 in Kubernetes 1.21.

pkg/router/controller/factory/factory.go &
pkg/router/controller/endpointsubset/converter.go &
pkg/router/controller/endpointsubset/converter_test.go &
pkg/router/controller/factory/factory_endpointslices_test.go:

Import `discoveryv1` and `corev1` explicitly. Update calling sites.
Update inline EndpointSlice resource API Version and Kind fields.
Update kube client API calls to use `discoveryv1`.

### DIFF
--- a/pkg/router/controller/endpointsubset/converter.go
+++ b/pkg/router/controller/endpointsubset/converter.go
@@ -1,22 +1,22 @@
 package endpointsubset
 
 import (
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/api/discovery/v1beta1"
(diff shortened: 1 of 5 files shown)


sgreene570:
No deprecated API errors visible in router pods in CI logs

/retest

PR #285, opened 2021-05-19
Bug 1961550: Add a condition to check if the Endpoints ID is duplicated

bysnupy:
* Description: EndpointSlice can be duplicated. For suppressing broken haproxy.config, add a condition to check if the Endpoints is duplicated before adding the new endpoints actually.
* Reference: https://bugzilla.redhat.com/show_bug.cgi?id=1961550


sgreene570:
/rename Bug 1961550: Add a condition to check if the Endpoints ID is duplicated

sgreene570:
/retitle Bug 1961550: Add a condition to check if the Endpoints ID is duplicated

sgreene570:
Thanks @bysnupy !
I discussed this patch with @frobware and we are happy with it as is. We are currently planning on back-porting this fix to 4.6.

/lgtm

PR #291, opened 2021-05-25
Bug 1955822: cmd: Add option to source stats auth from file

sgreene570:
This commit adds the capability to specify router
stats (metrics) authentication credentials via separate
username and password files.

This commit does not remove the prior behavior, in which stats
credentials are fed into the router via the STATS_USERNAME
and STATS_PASSWORD environment variables. These environment
variables are ignored when both STATS_PASSWORD_FILE and
STATS_USERNAME_FILE are provided.

pkg/cmd/infra/router/template.go:

Add new startup options, `stats-user-file` and `stats-password-file`.

Add a new helper function, `getStatsAuth`, that
returns the contents of STATS_USERNAME_FILE and STATS_PASSWORD_FILE,
or the values set in STATS_USERNAME and STATS_PASSWORD, depending on which
are provided. Preference is given to stats auth credentials provided via
files.

This commit is in support of Bug 1955822.

### DIFF
--- a/pkg/cmd/infra/router/template.go
+++ b/pkg/cmd/infra/router/template.go
@@ -185,15 +185,19 @@ func (o *TemplateRouter) Bind(flag *pflag.FlagSet) {
 }
 
 type RouterStats struct {
-	StatsPortString string
-	StatsPassword   string
-	StatsUsername   string
+	StatsPortString   string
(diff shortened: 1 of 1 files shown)


Miciah:
Thanks!
/lgtm

PR #295, opened 2021-05-31
Bug 1964486:  template helper - generateHAProxyWhiteListFile, use right arg type

alebedev87:
Fix for the following error: 
```
error executing template for file /tmp/router004931081/conf/haproxy.config: template: haproxy-config.template:492:84: executing "conf/haproxy.config" at <$cfgIdx>: wrong type for value; expected string; got templaterouter.ServiceAliasConfigKey
```
It happens when the whitelist of IPs has to be put into a file due to [the limitation of HAProxy config](https://github.com/openshift/router/blob/master/pkg/router/template/util/haproxy/whitelist.go#L11).

### DIFF
--- a/pkg/router/router_test.go
+++ b/pkg/router/router_test.go
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strconv"
+	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -20,6 +23,7 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	kubefake "k8s.io/client-go/kubernetes/fake"
@@ -46,6 +50,8 @@ func (h *harness) nextUID() types.UID {
 }
 
 var h *harness
+
+var reloadInterval = time.Duration(100 * time.Millisecond)
 
 func TestMain(m *testing.M) {
 	logFlags := flag.FlagSet{}
@@ -94,7 +100,7 @@ func TestMain(m *testing.M) {
 		DefaultCertificateDir: workdir,
 		ReloadFn:              func(shutdown bool) error { return nil },
 		TemplatePath:          "../../images/router/haproxy/conf/haproxy-config.template",
-		ReloadInterval:        15 * time.Second,
+		ReloadInterval:        reloadInterval,
 	}
 	plugin, err = templateplugin.NewTemplatePlugin(pluginCfg, svcFetcher)
 	if err != nil {
@@ -147,6 +153,8 @@ func TestAdmissionEdgeCases(t *testing.T) {
 		},
 	}
 
+	defer cleanUpRoutes(t)
+
 	for name, expectations := range tests {
 		for _, expectation := range expectations {
 			err := expectation.Apply(h)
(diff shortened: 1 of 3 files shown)


alebedev87:
@Miciah : to you first, to check whether the PR matches the standards.

pkg/router/router_test.go
```diff
@@ -157,24 +163,101 @@ func TestAdmissionEdgeCases(t *testing.T) {
...
+	wg.Add(1)
+	go func(ch chan error, stop chan struct{}, wg *sync.WaitGroup) {
+		defer wg.Done()
+	out:
+		for {
+			select {
+			case err := <-ch:
+				caughtErrors = append(caughtErrors, err)
+			case <-stop:
+				break out
```
Miciah, line 180: I think we can simplify the flow control and avoid using a label by using `return`:
```suggestion
		for {
			select {
			case err := <-ch:
				caughtErrors = append(caughtErrors, err)
			case <-stop:
				return
```
There is nothing after the `for` loop, and the deferred `wg.Done()` will still be executed when the `return` is reached.

pkg/router/router_test.go
```diff
@@ -157,24 +163,101 @@ func TestAdmissionEdgeCases(t *testing.T) {
...
+	// let the router reload
+	time.Sleep(reloadInterval * 2)
+
+	stopCh <- struct{}{}
+	wg.Wait()
+
+	// check for errors
+	for _, e := range caughtErrors {
+		if strings.Contains(e.Error(), "error executing template") {
+			t.Fatalf("Template exection failure found: %s", e)
```
Miciah, line 225: You have a typo in "execution", and you might as well use `%v` for the error value:
```suggestion
			t.Fatalf("Template execution failed: %v", e)
```

pkg/router/template/template_helper_test.go
```diff
@@ -805,3 +807,71 @@ func TestClipHAProxyTimeoutValue(t *testing.T) {
 		}
 	}
 }
+
+func TestGenerateHAProxyWhiteListFile(t *testing.T) {
+	workDir, err := ioutil.TempDir("", "haproxy")
+	if err != nil {
+		t.Fatal("Unable to create a temporary working directory")
+	}
+	defer os.RemoveAll(workDir)
```
Miciah, line 816: If you use `t.TempDir`, the resulting temporary directory will be automatically cleaned up when the test ends:
```suggestion
	workDir, err := t.TempDir("", "haproxy")
	if err != nil {
		t.Fatal("Unable to create a temporary working directory")
	}
```

pkg/router/template/template_helper_test.go
```diff
@@ -805,3 +807,71 @@ func TestClipHAProxyTimeoutValue(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			file := generateHAProxyWhiteListFile(tc.workDir, tc.id, strings.Join(tc.expectedWhiteList, " "))
+			if tc.failureExpected {
+				if file != "" {
+					t.Fatal("Failure expected but didn't happen")
+				}
+				return
+			} else {
+				if file == "" {
+					t.Fatal("Not expected failure happened")
```
Miciah, line 863:
```suggestion
					t.Fatal("Unexpected failure")
```

pkg/router/template/template_helper_test.go
```diff
@@ -805,3 +807,71 @@ func TestClipHAProxyTimeoutValue(t *testing.T) {
...
+				}
+			}
+
+			contents, err := ioutil.ReadFile(file)
+			if err != nil {
+				t.Fatalf("Unable to read from the generated file: %q", err)
```
Miciah, line 869: Is there a reason to use `%q` here?  
```suggestion
				t.Fatalf("Unable to read from the generated file: %v", err)
```

pkg/router/template/template_helper_test.go
```diff
@@ -805,3 +807,71 @@ func TestClipHAProxyTimeoutValue(t *testing.T) {
...
+			}
+			gotWhiteList := strings.Fields(string(contents))
+			if !reflect.DeepEqual(tc.expectedWhiteList, gotWhiteList) {
+				t.Errorf("Wrong white list written. Expected: %v but got %v", tc.expectedWhiteList, gotWhiteList)
```
Miciah, line 873: It makes sense to me to use `%q` here:
```suggestion
				t.Errorf("Wrong whitelist written: expected %q, got %q", tc.expectedWhiteList, gotWhiteList)
```

Miciah (commented):
Thanks!  The fix looks correct, and the added test is terrific.  I have some small suggestions to make the code a little neater.

pkg/router/template/template_helper_test.go
alebedev87: No, not much, I thought it would look a bit more clear, but I think you are right - `%v` would do this thing.

Miciah:
/approve
/lgtm
/hold
to give others a chance to review.

pkg/router/router_test.go
```diff
@@ -94,7 +100,7 @@ func TestMain(m *testing.M) {
 		DefaultCertificateDir: workdir,
 		ReloadFn:              func(shutdown bool) error { return nil },
 		TemplatePath:          "../../images/router/haproxy/conf/haproxy-config.template",
-		ReloadInterval:        15 * time.Second,
+		ReloadInterval:        reloadInterval,
```
frobware, line 103: Why such a big jump? And why do we then `* 2` a little later on?

pkg/router/router_test.go
```diff
@@ -157,24 +163,100 @@ func TestAdmissionEdgeCases(t *testing.T) {
...
+		for _, expectation := range expectations {
+			err := expectation.Apply(h)
+			if err != nil {
+				t.Fatalf("%s failed: %v", name, err)
+			}
+		}
+	}
+
+	// let the router reload
+	time.Sleep(reloadInterval * 2)
```
frobware, line 216: Is this a synthetic sleep? Why does sleeping 200ms work here? Would it work without the multiplication?

pkg/router/router_test.go
alebedev87: I had to make sure the router is reloaded, the way I'm doing it is by sleeping more than the `reloadInterval`.
I didn't want the unit test to last >15 seconds, so I dropped the reloadInterval to a more appropriate value.

pkg/router/router_test.go
alebedev87: Yes, it's synthetic. I  wanted to be sure the reload happened, that's why waiting twice more that the reload interval.

As we discussed, I made a stress test of this test function, it looks stable:
```bash
$ go test -run=TestConfigTemplateExecution -count=1000 ./pkg/router/
ok  	github.com/openshift/router/pkg/router	201.083s
$ stress -p=4 ./router.test -test.run=TestConfigTemplateExecution
5s: 52 runs so far, 0 failures
10s: 108 runs so far, 0 failures
15s: 164 runs so far, 0 failures
20s: 224 runs so far, 0 failures
25s: 280 runs so far, 0 failures
30s: 336 runs so far, 0 failures
...
12m0s: 8122 runs so far, 0 failures
12m5s: 8178 runs so far, 0 failures
12m10s: 8234 runs so far, 0 failures
12m15s: 8290 runs so far, 0 failures
^C
```

alebedev87:
@Miciah : I added the cleanup at the end of the test function to be able to rerun it many times (for the stress test).
Can you please have another look at the test?

pkg/router/router_test.go
frobware: Thanks. 

/lgtm

pkg/router/router_test.go
```diff
@@ -157,24 +163,104 @@ func TestAdmissionEdgeCases(t *testing.T) {
...
+	// check for errors
+	for _, e := range caughtErrors {
+		if strings.Contains(e.Error(), "error executing template") {
+			t.Fatalf("Template execution failed: %v", e)
+		}
+	}
+
+	if err := cleanUpRoutes(h); err != nil {
+		t.Errorf("Failed to clean up: %v", err)
+	}
```
Miciah, line 230: Would it make sense to do this in every test in this file that adds routes?  

This won't be executed if the above code calls `t.Fatalf`, which terminates execution of the test.  Could the call to `cleanupRoutes` be done using `defer` to ensure it always gets executed?

You could simplify usage of `cleanupRoutes` by adding a parameter for `t`, removing `h` (which is already a global), and calling `t.Errorf` inside `cleanupRoutes`, so callers could simply use `defer cleanupRoutes(t)` right after adding a bunch of routes.

alebedev87: Done

Miciah:
Thanks again!
/lgtm
/hold cancel

PR #298, opened 2021-06-03
Bug 1953113: template config - HSTS header's pattern accepts case insensitive and white spaces

alebedev87:
This PR makes HAProxy's config template accept HSTS headers with case insensitive directives and white spaces as stated in [RFC6797](https://datatracker.ietf.org/doc/html/rfc6797#section-6.1).

Apart from the unit test, tested manually:
```bash
$ oc get route console
NAME      HOST/PORT                                    PATH   SERVICES   PORT    TERMINATION          WILDCARD
console   console-openshift-console.apps-crc.testing          console    https   reencrypt/Redirect   None

$ oc get route console --template='{{$a := index .metadata.annotations "haproxy.router.openshift.io/hsts_header"}}{{$a}}{{"\n"}}'
MAX-age=99999 ;   includesubdomains;   Preload

$ curl -k -is https://console-openshift-console.apps-crc.testing | grep -i strict-transport
strict-transport-security: MAX-age=99999 ;   includesubdomains;   Preload

```

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -33,8 +33,9 @@
 {{- $timeSpecPattern := `[1-9][0-9]*(us|ms|s|m|h|d)?` }}
 
 {{- /* hsts header in response: */}}
+{{- /* Not fully compliant to RFC6797#6.1 yet: has to accept not conformant directives */}}
 {{- $hstsOptionalTokenPattern := `(?:includeSubDomains|preload)` }}
-{{- $hstsPattern := printf `(?:%[1]s[;])*max-age=(?:\d+|"\d+")(?:[;]%[1]s)*`  $hstsOptionalTokenPattern -}}
+{{- $hstsPattern := printf `(?i)(?:%[1]s\s*[;]\s*)*max-age\s*=\s*(?:\d+|"\d+")(?:\s*[;]\s*%[1]s)*`  $hstsOptionalTokenPattern -}}
 
 {{- /* setForwardedHeadersPattern matches valid options for how and when Forwarded: and X-Forwarded-*: headers are set. */}}
 {{- $setForwardedHeadersPattern := `(?:append|replace|if-none|never)` -}}
@@ -602,7 +603,7 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
 
         {{- if matchValues (print $cfg.TLSTermination) "edge" "reencrypt" }}
           {{- with $hsts := firstMatch $hstsPattern (index $cfg.Annotations "haproxy.router.openshift.io/hsts_header") }}
-  http-response set-header Strict-Transport-Security {{ $hsts }}
+  http-response set-header Strict-Transport-Security '{{ $hsts }}'
           {{- end }}{{/* hsts header */}}
         {{- end }}{{/* is "edge" or "reencrypt" */}}
 
--- a/pkg/router/router_test.go
+++ b/pkg/router/router_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
@@ -42,6 +43,8 @@ type harness struct {
 
 	namespace string
 	uidCount  int
+	workdir   string
+	dirs      map[string]string
 }
 
 func (h *harness) nextUID() types.UID {
@@ -92,6 +95,13 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 	fmt.Printf("router working directory: %s\n", workdir)
+
+	h.workdir = workdir
+	h.dirs = map[string]string{
+		"whitelist": filepath.Join(workdir, "router", "whitelists"),
+	}
+
+	createRouterDirs()
 
 	// The template plugin which is wrapped
 	svcFetcher := templateplugin.NewListWatchServiceLookup(client.CoreV1(), 60*time.Second, namespace)
@@ -165,7 +175,7 @@ func TestAdmissionEdgeCases(t *testing.T) {
 	}
 }
 
-func TestConfigTemplateExecution(t *testing.T) {
+func TestConfigTemplate(t *testing.T) {
 	// watching for errors
 	caughtErrors := []error{}
 	errCh, stopCh := make(chan error), make(chan struct{})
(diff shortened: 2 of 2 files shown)


pkg/router/router_test.go
```diff
@@ -86,13 +89,17 @@ func TestMain(m *testing.M) {
...
+	var err error
+	workdir, err = ioutil.TempDir("", "router")
 	if err != nil {
 		fmt.Println(err)
 		os.Exit(1)
 	}
 	fmt.Printf("router working directory: %s\n", workdir)
 
+	// directory for whitelist files (best effort, to not see error logs)
+	os.MkdirAll(path.Join(workdir, "router", "whitelists"), 0775)
```
candita, line 101: Was this mistakenly included as a part of this PR?

pkg/router/router_test.go
```diff
@@ -203,6 +211,18 @@ func TestConfigTemplateExecution(t *testing.T) {
 				tlsTermination: routev1.TLSTerminationEdge,
 			},
 		},
+		"case insesitive, with white spaces HSTS header": {
```
candita, line 214: typo: insesitive

pkg/router/router_test.go
```diff
@@ -86,13 +89,17 @@ func TestMain(m *testing.M) {
 
 	var plugin router.Plugin
 
-	workdir, err := ioutil.TempDir("", "router")
+	var err error
+	workdir, err = ioutil.TempDir("", "router")
```
candita, line 93: this isn't necessary, you could keep the original

pkg/router/router_test.go
```diff
@@ -86,13 +89,17 @@ func TestMain(m *testing.M) {
...
 	if err != nil {
 		fmt.Println(err)
 		os.Exit(1)
 	}
 	fmt.Printf("router working directory: %s\n", workdir)
 
+	// directory for whitelist files (best effort, to not see error logs)
```
candita, line 100: You may need to update.  Is this change purposeful?

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -34,7 +34,7 @@
 
 {{- /* hsts header in response: */}}
 {{- $hstsOptionalTokenPattern := `(?:includeSubDomains|preload)` }}
-{{- $hstsPattern := printf `(?:%[1]s[;])*max-age=(?:\d+|"\d+")(?:[;]%[1]s)*`  $hstsOptionalTokenPattern -}}
+{{- $hstsPattern := printf `(?i)(?:%[1]s\s*[;]\s*)*max-age=(?:\d+|"\d+")(?:\s*[;]\s*%[1]s)*`  $hstsOptionalTokenPattern -}}
```
candita, line 37: Nice.  Does the spec allow spaces around the equal sign?

pkg/router/router_test.go
```diff
@@ -165,7 +172,7 @@ func TestAdmissionEdgeCases(t *testing.T) {
 	}
 }
 
-func TestConfigTemplateExecution(t *testing.T) {
+func TestConfigTemplate(t *testing.T) {
```
candita, line 175: Is there a reason to change the method name?

pkg/router/router_test.go
```diff
@@ -203,6 +211,18 @@ func TestConfigTemplateExecution(t *testing.T) {
...
+				name: "h",
+				host: "hstsexample.com",
+				path: "",
+				time: start,
+				annotations: map[string]string{
+					"haproxy.router.openshift.io/hsts_header": hstsHeaderValue,
+				},
+				tlsTermination: routev1.TLSTerminationEdge,
+			},
+		},
```
candita, line 225: One more test that would fail is good.

pkg/router/router_test.go
```diff
@@ -228,6 +248,23 @@ func TestConfigTemplateExecution(t *testing.T) {
 			t.Fatalf("Template execution failed: %v", e)
 		}
 	}
+
+	// check the generated config
```
candita, line 252: This is a nice addition. Can you add the test cases for the whitelist too?

pkg/router/router_test.go
alebedev87: I wanted `workdir` variable to be accessible by all the tests, that's why I moved it to the global scope.
The problem here is with the shadowing of the global variables when `:=` is used.
That is, if I keep it the way it was - the value of the global `workdir` won't change, Go will just create a new local variable and will set it (not the global).

pkg/router/router_test.go
alebedev87: It's just an error message I didn't notice when I first implemented the fix for the IP whitelists.
The error is about a missing directory for the whitelist files (long IP lists are put into a file as HAProxy config has a hard coded limitation of the length of config args).

I guess it brings a "better looking test" output.

pkg/router/router_test.go
alebedev87: Yes, I think it reflects better what the test is doing. Before it was only testing the parsing of the template, not it also checks for some patterns in the generated config file.

images/router/haproxy/conf/haproxy-config.template
alebedev87: Good question!
Looks like RFC2616 is allowing LWS (linear white spaces) between any word and separator (`=` is also considered as separator): [here](https://datatracker.ietf.org/doc/html/rfc2616#section-2.1) (`implied *LWS` notion).
Didn't see examples with LWS around `=` though.

pkg/router/router_test.go
alebedev87: Put `workdir` to the global harness. This helped to revert this code back to how it looked before.

pkg/router/router_test.go
alebedev87: Changed this part a bit by adding `createRouteDirs` function.

pkg/router/router_test.go
alebedev87: I added some more patterns and not matching patterns too.
Now it should easier to understand what is expected and what is not.

It also helped me to find a problem in the config template, I used double quotes for `Strict-Transport-Security` but `max-age` can have double quotes too. This may be badly interpreted by HAProxy.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -34,7 +34,7 @@
 
 {{- /* hsts header in response: */}}
 {{- $hstsOptionalTokenPattern := `(?:includeSubDomains|preload)` }}
-{{- $hstsPattern := printf `(?:%[1]s[;])*max-age=(?:\d+|"\d+")(?:[;]%[1]s)*`  $hstsOptionalTokenPattern -}}
+{{- $hstsPattern := printf `(?i)(?:%[1]s\s*[;]\s*)*max-age\s*=\s*(?:\d+|"\d+")(?:\s*[;]\s*%[1]s)*`  $hstsOptionalTokenPattern -}}
```
candita, line 37: Actually, as I read your unit tests, I realized this gets a little complicated.  In https://datatracker.ietf.org/doc/html/rfc6797#section-6.1 it says that it's okay to supply invalid directives, they just get ignored.  max-age is the only one that MUST be there.  So any other wrong directive is okay, even a wrongly spelled includeSubdomains or preload.  If max-age is missing (or spelled wrong), that would make it not match and no HSTS header printed.

pkg/router/router_test.go
```diff
@@ -191,16 +201,126 @@ func TestConfigTemplateExecution(t *testing.T) {
...
+		},
+		"Wrong HSTS header directive": {
+			mustCreateWithConfig{
+				mustCreate: mustCreate{
+					name: "g",
+					host: "gexample.com",
+					path: "",
+					time: start,
+					annotations: map[string]string{
+						"haproxy.router.openshift.io/hsts_header": "max-age=99999;includesubdomains;preload;wrongdirective",
```
candita, line 302: This should pass okay -- see my comment above.

pkg/router/router_test.go
```diff
@@ -191,16 +201,126 @@ func TestConfigTemplateExecution(t *testing.T) {
...
+					host: "hexample.com",
+					path: "",
+					time: start,
+					annotations: map[string]string{
+						"haproxy.router.openshift.io/hsts_header": "max-age=99999;includesubdomain",
+					},
+					tlsTermination: routev1.TLSTerminationEdge,
 				},
-				tlsTermination: routev1.TLSTerminationEdge,
+				configPattern: `http-response set-header Strict-Transport-Security 'max-age=99999;includesubdomain'`,
```
candita, line 322: This should also pass ok, because `includeSubDomains` is optional.

images/router/haproxy/conf/haproxy-config.template
candita: Per our conversation, we will probably be removing non-matching directives from the header in the API when we implement https://issues.redhat.com/browse/NE-310. I think the pattern should be okay as is, and if there are further issues with writing HSTS before NE-310, we have to remember this conversation.

alebedev87: As discussed, I keep it the was it is adding the comment about the not full conformance

pkg/router/router_test.go
```diff
@@ -93,6 +96,13 @@ func TestMain(m *testing.M) {
 	}
 	fmt.Printf("router working directory: %s\n", workdir)
 
+	h.workdir = workdir
+	h.dirs = map[string]string{
+		"whitelist": path.Join(workdir, "router", "whitelists"),
```
Miciah, line 101: We should use the filepath package:
```suggestion
		"whitelist": filepath.Join(workdir, "router", "whitelists"),
```

pkg/router/router_test.go
```diff
@@ -228,6 +365,32 @@ func TestConfigTemplateExecution(t *testing.T) {
...
+	contents, err := ioutil.ReadFile(config)
+	if err != nil {
+		t.Fatalf("Failed to read the generated config: %v", err)
+	}
+
+	for name, expectations := range tests {
+		for _, expectation := range expectations {
+			createExpectation, ok := expectation.(mustCreateWithConfig)
+			if !ok {
+				continue
```
Miciah, line 380: Could we call `t.Fatalf` here to avoid silently skipping a test?  Alternatively, could we use a `map[string]mustCreateWithConfig` type for the test cases?

pkg/router/router_test.go
```diff
@@ -277,6 +440,12 @@ func (e mustCreate) Apply(h *harness) error {
 	return err
 }
 
+type mustCreateWithConfig struct {
+	mustCreate
+	configPattern string
```
Miciah, line 445: The name is confusing; the test actually performs a substring search, not a pattern match (such as a regexp), right?

pkg/router/router_test.go
alebedev87: Test cases' type changed to `map[string][]mustCreateWithConfig` to avoid the cast.

pkg/router/router_test.go
alebedev87: Didn't have many idea for the name, renamed to `configSnippet`. Tell me if you have better ideas.

pkg/router/router_test.go
```diff
@@ -190,17 +200,144 @@ func TestConfigTemplateExecution(t *testing.T) {
...
+					name: "g",
+					host: "gexample.com",
+					path: "",
+					time: start,
+					annotations: map[string]string{
+						"haproxy.router.openshift.io/hsts_header": "min-age=99999",
+					},
+					tlsTermination: routev1.TLSTerminationEdge,
+				},
+				configSnippet: `http-response set-header Strict-Transport-Security 'min-age=99999'`,
```
Miciah, line 306: Can we check for just the substring "http-response set-header Strict-Transport-Security" to make sure there are no false negatives?
```suggestion
				configSnippet: `http-response set-header Strict-Transport-Security`,
```

pkg/router/router_test.go
```diff
@@ -190,17 +200,144 @@ func TestConfigTemplateExecution(t *testing.T) {
...
+					name: "h",
+					host: "hexample.com",
+					path: "",
+					time: start,
+					annotations: map[string]string{
+						"haproxy.router.openshift.io/hsts_header": "max-age=99999;includesubdomains;preload;wrongdirective",
+					},
+					tlsTermination: routev1.TLSTerminationEdge,
+				},
+				configSnippet: `http-response set-header Strict-Transport-Security 'max-age=99999;includesubdomains;preload;wrongdirective'`,
```
Miciah, line 323:
```suggestion
				configSnippet: `http-response set-header Strict-Transport-Security`,
```

pkg/router/router_test.go
Miciah: Oh, the state isn't cleared between test cases, is it?  So with my suggestion, this test case would always fail.

Miciah:
Thanks!  I love the improved test coverage!
/lgtm
/approve

PR #300, opened 2021-06-08
Bug 1967228: error pages - don't use bootstrap/normalize

alebedev87:
@sg00dwin proposed me new CSS which they are using for the management console. It's based on PatternFly4 which is not using bootstrap and normalize.css anymore.

This is [how 503 page looked before](https://codepen.io/alebedev87/pen/NWpBGdP).
This is [how it looks now.](https://codepen.io/alebedev87/pen/bGqjVEX).
This is [how 404 page looks now](https://codepen.io/alebedev87/pen/mdWjJKL).

Note that nginx 503 page was not using CRLF, I kept it this way.


Miciah:
Thanks!
/lgtm

PR #307, opened 2021-06-16
Bug 1964482: config template: accept IPv6 IPs for whitelisting

alebedev87:
New template helper to parse a list of IPs/CIDRs for the whitelisting.
Takes a different approach than https://github.com/openshift/router/pull/262 which uses a long regexp for IPv6 IPs/CIDRs.
The new function aims to keep the same behavior as the previous approach hence some corner cases (i.e. leading/trailing spaces).

### DIFF
--- a/pkg/router/template/template_helper.go
+++ b/pkg/router/template/template_helper.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math/rand"
+	"net"
 	"os"
 	"path"
 	"regexp"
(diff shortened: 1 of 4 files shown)


pkg/router/template/template_helper.go
```diff
@@ -358,6 +359,34 @@ func clipHAProxyTimeoutValue(val string) string {
...
+
+	// same behavior as the previous approach with regexp
+	if strings.TrimSpace(list) != list {
+		log.V(7).Info("parseIPList leading/trailing spaces found")
+		return ""
+	}
+
+	ipList := strings.Fields(list)
+	if len(ipList) == 0 {
+		log.V(7).Info("parseIPList didn't find any IP/CIRD")
```
sgreene570, line 375:
```suggestion
		log.V(7).Info("parseIPList didn't find any IP/CIDR")
```

Typo?

pkg/router/template/template_helper_test.go
```diff
@@ -871,3 +871,82 @@ func TestGenerateHAProxyWhiteListFile(t *testing.T) {
...
+			input:         "",
+			expectedEmpty: true,
+		},
+		{
+			name:          "Wrong IPv4",
+			input:         "192.168.",
+			expectedEmpty: true,
+		},
+		{
+			name:          "Wrong IPv6",
```
sgreene570, line 927: Maybe add some unit test cases for single bad IPv4 and IPv6 CIDRs?

sgreene570:
Overall looks great! Some small comments. Would like to get @Miciah's input on this as well.

pkg/router/template/template_helper.go
```diff
@@ -358,6 +359,34 @@ func clipHAProxyTimeoutValue(val string) string {
 	return val
 }
 
+// parseIPList parses white space separated list of IPs/CIDRs (IPv4/IPv6)
+// aims at providing the same behavior as the previous approach with regexp in the template file
+func parseIPList(list string) string {
```
sgreene570, line 364: Would it make sense to return "" immediately if len(string) is 0? Worried about the logging call on line 375 being spammy since its technically the "default" case. WDYT?

mic4ael:
Hey, are you planning to fix the cherry-picking error?

PR #311, opened 2021-06-18
router test: config template: don't match against the whole file

alebedev87:
Improvement to the test introduced in https://github.com/openshift/router/pull/307.
The check for the presence of config snippets is not made against the whole config file anymore.
It now searches only the backend part of the config corresponding to the route.
This helps to avoid the false positives and negatives.

Note the usage of https://github.com/haproxytech/config-parser


pkg/router/router_test.go
```diff
@@ -368,15 +369,21 @@ func TestConfigTemplate(t *testing.T) {
 
 	// check the generated config
 	config := filepath.Join(h.workdir, "conf", "haproxy.config")
-	contents, err := ioutil.ReadFile(config)
+	parser := haproxyutil.NewConfigParser(config)
```
frobware, line 372: As an aside I think we should consider using this for our "lobfuscator" ticket - https://issues.redhat.com/browse/NE-419.

pkg/router/template/util/haproxy/config_parser.go
```diff
@@ -0,0 +1,165 @@
...
+const (
+	commentChar     = "#"
+	globalKeyword   = "global"
+	defaultsKeyword = "defaults"
+	frontendKeyword = "frontend"
+	backendKeyword  = "backend"
+	edgeBePrefix    = "be_edge_http"
+)
+
+// ConfigParser is a primitive HAProxy config parser.
```
frobware, line 19: I'm wondering whether you looked at https://github.com/haproxytech/config-parser and if that would do what we need.

alebedev87: No, I didn't, I wasn't expecting any golang libs from HAProxy community. But this one is even from the upstream! Thanks a lot for the advice. This module looks very promising:
```golang
$ cat config.cfg  | grep -A 15 be_edge_http:test-route:hello-openshift2
backend be_edge_http:test-route:hello-openshift2
  mode http
  option redispatch
  option forwardfor
  balance 

  timeout check 5000ms
  http-response set-header Strict-Transport-Security 'max-age=99999;includesubdomain'
  cookie f7e936ebe97cd64e8c888d1eba08cb29 insert indirect nocache httponly secure attr SameSite=None
  server pod:hello-openshift-7b8c68587c-mtck6:hello-openshift:8080-tcp:10.217.0.24:8080 10.217.0.24:8080 cookie 97ddca1460eca46376928fd7bcf8c89c weight 256 check inter 5000ms

$ cat main.go 
package main

import (
	"fmt"
	"github.com/haproxytech/config-parser/v4"
	"github.com/haproxytech/config-parser/v4/options"
	"github.com/haproxytech/config-parser/v4/types"
)

func main() {
	p, err := parser.New(options.Path("config.cfg"))
	if err != nil {
		fmt.Println("failed to parse the config:", err)
	}

	data, err := p.Get(parser.Backends, "be_edge_http:test-route:hello-openshift2", "http-response", false)
	if err != nil {
		fmt.Println("failed to get backend:", err)
	}

	responses := data.([]types.HTTPAction)
	for _, r := range responses {
		fmt.Println(r)
	}
}

$ go run main.go 
set-header Strict-Transport-Security 'max-age=99999;includesubdomain'
```

Let me try to integrate it into my PR.

alebedev87:
/hold

I will try to integrate https://github.com/haproxytech/config-parser module instead of the home made parser.

pkg/router/router_test.go
```diff
@@ -453,8 +501,64 @@ func (e mustCreate) Apply(h *harness) error {
...
+		}
+	case []haproxyconfparsertypes.ACL:
+		acls := data.([]haproxyconfparsertypes.ACL)
+		for _, a := range acls {
+			if a.Name+" "+a.Criterion+" "+a.Value == m.value {
+				contains = true
+				break
+			}
+		}
+	}
```
frobware, line 543:
```suggestion
	switch data := data.(type) {
	case []haproxyconfparsertypes.HTTPAction:
		for _, a := range data {
			if a.String() == m.value {
				contains = true
				break
			}
		}
	case []haproxyconfparsertypes.ACL:
		for _, a := range data {
			if a.Name+" "+a.Criterion+" "+a.Value == m.value {
				contains = true
				break
			}
		}
	}
```

frobware (approved):
It's great that the upstream config parser slips in so easily.

pkg/router/router_test.go
alebedev87: Changed

PR #312, opened 2021-06-22
WIP: reload-haproxy: Adjust oom_score_adj of old procs

Miciah:
Add logic to increase the `oom_score_adj` setting for each old `haproxy` process when reloading HAProxy.

* `images/router/haproxy/reload-haproxy` (`adjustOomScores`): New function. Increase the `oom_score_adj` setting for each of the provided pids by 10.  Call `adjustOomScores` at the end of `reload-haproxy`.

### DIFF
--- a/images/router/haproxy/reload-haproxy
+++ b/images/router/haproxy/reload-haproxy
@@ -69,6 +69,18 @@ function haproxyHealthCheck() {
 }
 
 
+function adjustOomScores() {
+    for pid; do
+	old_score=$(< /proc/$pid/oom_score_adj)
+	new_score=$((old_score + 10))
+	# The maximum value for oom_score_adj is 1000.
+	[[ $new_score -gt 1000 ]] && continue
+	echo $new_score > /proc/$pid/oom_score_adj
+	echo " - Adjusted oom_score_adj for haproxy[$pid] from $old_score to $new_score."
+    done
+}
+
+
 old_pids=$(pidof haproxy)
 
 # If signaled, stop accepting new connections and drain the current processes
@@ -101,3 +113,4 @@ fi
 
 [ $reload_status -ne 0 ] && exit $reload_status
 haproxyHealthCheck
+adjustOomScores $old_pids


images/router/haproxy/reload-haproxy
```diff
@@ -69,6 +69,16 @@ function haproxyHealthCheck() {
 }
 
 
+function adjustOomScores() {
+    for pid; do
+	old_score=$(< /proc/$pid/oom_score_adj)
+	new_score=$((old_score + 10))
```
frobware, line 75: Any concern that this (hypothetically) wraps which (assuming it goes negative) would defeat the intent here?
I didn't look at the underlying types - just a throwaway observation.

images/router/haproxy/reload-haproxy
```diff
@@ -101,3 +111,4 @@ fi
 
 [ $reload_status -ne 0 ] && exit $reload_status
 haproxyHealthCheck
+adjustOomScores $old_pids
```
frobware, line 114: TIL: choom(1) - https://github.com/karelzak/util-linux/commit/8fa223daba1963c34cc828075ce6773ff01fafe3
Doesn't appear to be available in rhel8 though.

images/router/haproxy/reload-haproxy
Miciah: The max value is 1000, so I'll add a check to avoid going above that.

Miciah:
This is getting errors like the following:

    /var/lib/haproxy/reload-haproxy: line 76: /proc/153/oom_score_adj: Permission denied

Looks like we would need to grant the router pod the `CAP_SYS_RESOURCE` capability, possibly using a custom SCC, in order for this change to work.

jacksgt:
Hi @Miciah , for which reason was the PR ultimately closed / abandoned?
We are quite interested in this feature since we're struggling with lots of old processes due to long-lived connections (WebSockets) on our infrastructure.

PR #322, opened 2021-07-29
Bug 1987238: Validate ROUTER_INSPECT_DELAY env value generating haproxy config

rfredette:
If `ROUTER_INSPECT_DELAY` does not match `$timeSpecPattern`, fall back to the default value

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -201,7 +201,7 @@ frontend public
     {{- end }}
   {{- if isTrue (env "ROUTER_USE_PROXY_PROTOCOL") }} accept-proxy{{ end }}
   mode http
-  tcp-request inspect-delay {{ env "ROUTER_INSPECT_DELAY" "5s" }}
+  tcp-request inspect-delay {{ firstMatch $timeSpecPattern (env "ROUTER_INSPECT_DELAY") "5s" }}
   tcp-request content accept if HTTP
 
     {{- if (eq .StatsPort -1) }}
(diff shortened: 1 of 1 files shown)


images/router/haproxy/conf/haproxy-config.template
```diff
@@ -201,7 +201,7 @@ frontend public
     {{- end }}
   {{- if isTrue (env "ROUTER_USE_PROXY_PROTOCOL") }} accept-proxy{{ end }}
   mode http
-  tcp-request inspect-delay {{ env "ROUTER_INSPECT_DELAY" "5s" }}
```
candita, line 204: Some of the timeout values use the `clipHAProxyTimeoutValue`, like: https://github.com/openshift/router/blob/master/images/router/haproxy/conf/haproxy-config.template#L44

Do you need to consider that?

rfredette: It shouldn't be necessary because the value is evaluated through `durationToHAProxyTimespec` when it's put into the environment variable in the first place, which acts similar to `clipHAProxyTimeoutValue`. (used here: https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/deployment.go#L540)

That said, I think it would also make sense to weed out bad timeout values in the ingress operator

quarterpin:
/lgtm
/label qe-approved

Verified via pre-merge verification workflow, more references related to the test can be found in:
https://bugzilla.redhat.com/show_bug.cgi?id=1987238#c1

PR #321, opened 2021-07-22
[release-4.8] Bug 1984565: config template: accept IPv6 IPs for whitelisting

alebedev87:
Manual backport of https://github.com/openshift/router/pull/307

PR #325, opened 2021-08-17
Bug 1994645: config template: accept IPv6 IPs for whitelisting

Miciah:
Manual cherry-pick of #321.

### DIFF
--- a/pkg/router/template/template_helper.go
+++ b/pkg/router/template/template_helper.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math/rand"
+	"net"
 	"os"
 	"path"
 	"regexp"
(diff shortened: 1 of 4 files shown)


Miciah:
[sig-cluster-lifecycle][Feature:Machines][Early] Managed cluster should have same number of Machines and Nodes [Suite:openshift/conformance/parallel]
    fail [github.com/openshift/origin/test/extended/machines/cluster.go:56]: Expected
        <int>: 5
    to equal
        <int>: 6

/test e2e-upgrade

pkg/router/template/template_helper.go
```diff
@@ -358,6 +359,35 @@ func clipHAProxyTimeoutValue(val string) string {
...
+	if trimmedList != list {
+		log.V(7).Info("parseIPList leading/trailing spaces found")
+		return ""
+	}
+
+	ipList := strings.Fields(list)
+	for _, ip := range ipList {
+		if net.ParseIP(ip) == nil {
+			if _, _, err := net.ParseCIDR(ip); err != nil {
+				log.V(7).Info("parseIPList found not IP/CIDR item", "value", ip, "err", err)
```
candita, line 382:
```suggestion
				log.V(7).Info("parse error - invalid IP/CIDR", "value", ip, "err", err)
```

pkg/router/template/template_helper.go
```diff
@@ -358,6 +359,35 @@ func clipHAProxyTimeoutValue(val string) string {
 	return val
 }
 
+// parseIPList parses white space separated list of IPs/CIDRs (IPv4/IPv6)
+// aims at providing the same behavior as the previous approach with regexp in the template file
+func parseIPList(list string) string {
```
candita, line 364: Great idea to remove more regexps.

pkg/router/template/template_helper.go
candita: Never mind, just realized this is a cherry-pick.

deads2k:
(patch manager)

high bz, validation, 0.5

PR #326, opened 2021-08-20
haproxy-config.template: Use Unix sockets

Miciah:
Use Unix sockets instead of TCP sockets to connect the `be_sni` backend to the `fe_sni` frontend and the `be_no_sni` backend to the `fe_no_sni` frontend.

This change avoids the need for dedicated TCP ports.  It might also have a beneficial effect on performance.

* `images/router/haproxy/conf/haproxy-config.template`: Use Unix sockets instead of TCP sockets for the `be_sni`, `fe_sni`, `be_no_sni`, and `fe_no_sni` frontends and backends.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -282,11 +282,11 @@ frontend public_ssl
 # traffic
 ##########################################################################
 backend be_sni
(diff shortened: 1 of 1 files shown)


frobware:
/lgtm
/hold

- Added /hold should you want to do additional verification. 
- Would be interested in some performance runs against this change.

alebedev87:
LGTM

frobware:
/lgtm
/hold cancel

Removing the hold as we will / can assess performance separately.

PR #353, opened 2021-09-30
[WIP] Add new metrics to track router reloads

frobware:
- When debugging production issues we don't have a clear picture of the number of HAProxy reloads that take place
- Experiment with a metric that allows us to track route_{add,delete,modified}


frobware:
This has fallen off my radar; closing for now but we may resurrect this.

PR #356, opened 2021-11-02
Bug 2013238: Verify routes with intermediates in the Certificate field

Jamstah:
Suggesting a fix to my own bug raise.

The router will concatenate the chains from the Certificate and the
CACertificate fields together to present to the client, so using a
Certificate of (leaf+int) and a CACertificate of (ca) provides
the same chain to the client as (leaf) and (int+ca).

Before this change, one configuration would fail to verify and the
other would not. After this change they both verify successfully,
recognising the valid TLS chain presented in each configuration.

As a side effect, it no longer attempts to verify routes that have
no valid CA certificates. These cases are already fail to validate
due to the malformed CA field, so the verification attempt was
not required.


Jamstah:
Can I get some thoughts on this one? Its 3 weeks old now.

Jamstah:
@brandisher these look like volume mount timeouts, so assuming infra instability - is that common in this repo or should I look deeper at the failures?

brandisher:
> @brandisher these look like volume mount timeouts, so assuming infra instability - is that common in this repo or should I look deeper at the failures?

It ebbs and flows a bit but I haven't see these errors before. I'll give it another run (or a few) for good measure but mostly I wanted to see how far we could get through the CI with your changes. Overall, we're looking pretty good but stay tuned for further updates!

Jamstah:
Better :)

LMK if a rebase or anything would be useful, the commit is quite old.

brandisher:
@Jamstah it looks like we have some more relevant failures in the most recent upgrade test. Would you mind doing a rebase so we can rule that out as a potential issue?

Jamstah:
> @Jamstah it looks like we have some more relevant failures in the most recent upgrade test. Would you mind doing a rebase so we can rule that out as a potential issue?

Done.

Jamstah:
So that didn't help :)

What are the tests that you think could be broken by this change? I've taken a look but nothing seems to immediately come back as "tls broken here", and not knowing the test suite it makes it hard for me to know what I should be poking.

PR #357, opened 2021-11-03
NE-882: Support subdomain field when host is unset

Miciah:
If a route has an empty `spec.host` and a nonempty `spec.subdomain` and the router has a domain configured, use the route's subdomain and router's domain as the route's host.

* `pkg/cmd/infra/router/router.go` (`RouterSelection`): Add `RouterDomain` field.
(`Bind`): Add a `--router-domain` flag and `ROUTER_DOMAIN` environment variable to specify the router's domain.
(`RouteUpdate`): Set the route's `spec.host` using the route's subdomain and router's domain if the host is empty and the subdomain and domain are nonempty.

### DIFF
--- a/pkg/cmd/infra/router/router.go
+++ b/pkg/cmd/infra/router/router.go
@@ -39,6 +39,7 @@ type RouterSelection struct {
 	UpdateStatus bool
 
 	HostnameTemplate string
+	RouterDomain     string
 	OverrideHostname bool
 	OverrideDomains  []string
 	RedactedDomains  sets.String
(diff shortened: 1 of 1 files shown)


pkg/cmd/infra/router/router.go
```diff
@@ -101,6 +103,11 @@ func (o *RouterSelection) Bind(flag *pflag.FlagSet) {
 
 // RouteUpdate updates the route before it is seen by the cache.
 func (o *RouterSelection) RouteUpdate(route *routev1.Route) {
+	// If the route specifies a subdomain and no host name and we a router
+	// domain, set the host field using the subdomain and domain.
+	if len(route.Spec.Host) == 0 && len(route.Spec.Subdomain) > 0 && len(o.RouterDomain) != 0 {
```
gcs278, line 108: Is this where we could do some warnings? Or would that be somewhere else?

Potential warnings:
1. What if you specified a subdomain, but don't have a router domain?
2. As mention in the enhancement, if you have both `spec.host` and `spec.subdomain`.

Miciah: > Is this where we could do some warnings? Or would that be somewhere else?
> 
> Potential warnings:
> 
>     1. What if you specified a subdomain, but don't have a router domain?

Maybe.  It could be noisy though.  Would it be better to log something at startup to indicate whether a router domain is set?  

>     2. As mention in the enhancement, if you have both `spec.host` and `spec.subdomain`.

Maybe, but again, it could be noisy.  Maybe `log.V(7).Info`?  (I suggest 7 based on the fact that `clipHAProxyTimeoutValue` uses it.)  I think that would better be handled as a warning during route admission; the user who creates the route may not be able to read router logs but would be more likely to see a warning printed in response to the route create or update.  Alternatively, we could report it in the route status somehow (maybe a new status condition?), though if you're looking at the status, you can already see that the router set the host in status to the host in spec.

gcs278: Sure, right here would be noisy because it would do it every update.

So you are saying `status.ingress.host` would not be set if the user messed up the subdomain or router domain? 

Yea I'm not sure if this is a big deal, as I am not sure there a very good way of getting warnings to the user other than status.

Miciah: > Sure, right here would be noisy because it would do it every update.

Right.  

> So you are saying `status.ingress.host` would not be set if the user messed up the subdomain or router domain?

That isn't quite what I'm saying.  Generally, spec expresses intent, and status reflects reality.  The API definition explicitly allows the user to specify both and states that `spec.host` takes precedence even if the router supports `spec.subdomain`, which is admittedly potentially confusing, but if the user expresses seemingly contradictory intent by specifying both `spec.host` and `spec.subdomain`, then the status reflects the actual host name that the router is using in status, which should help clear up the confusion.  

> Yea I'm not sure if this is a big deal, as I am not sure there a very good way of getting warnings to the user other than status.

I'll look into it.  However, I see it as more of a nice-to-have than a hard requirement.

gcs278: Agreed, it's a nice to have.

Miciah: I added a warning to the admission code (see https://github.com/openshift/openshift-apiserver/pull/254#issuecomment-1108076241).

gcs278: nice!

gcs278:
/lgtm
:green_apple:

Miciah:
```
[sig-devex][Feature:Templates] templateinstance impersonation tests should pass impersonation update tests [Suite:openshift/conformance/parallel]
fail [github.com/openshift/origin/test/extended/templates/templateinstance_impersonation.go:141]: Unexpected error:
    <*errors.errorString | 0xc00030caf0>: {
        s: "timed out waiting for the condition",
    }
    timed out waiting for the condition
occurred
```
/test e2e-metal-ipi-ovn-ipv6

PR #359, opened 2021-11-12
Bug 1894431: Add missing newlines to default certificate and key

Miciah:
Before this change, if the administrator provided a default certificate that was missing the newline character for the last line, the router would write out a corrupt PEM file for HAProxy, causing HAProxy to fail to start. This change adds logic to add the missing newline character to an incomplete line when writing out the PEM file.  As a result, the router now writes out a valid PEM file so that HAProxy can start even if the input is missing a newline character.

* `pkg/router/template/router.go` (`secretToPem`): Check for missing newline characters and add them if needed.
* `pkg/router/template/router_test.go` (`testWildcardCertificate`, `testWildcardCertificateKey`): New consts.
(`TestSecretToPem`): Verify that `secretToPem` writes out the expected PEM file.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -291,6 +291,9 @@ func secretToPem(secPath, outName string) error {
 	if err != nil {
 		return err
 	}
+	if len(pemBlock) > 0 && pemBlock[len(pemBlock)-1] != byte('\n') {
+		pemBlock = append(pemBlock, byte('\n'))
+	}
 	keys, err := privateKeysFromPEM(pemBlock)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


pkg/router/template/router.go
```diff
@@ -291,6 +291,9 @@ func secretToPem(secPath, outName string) error {
 	if err != nil {
 		return err
 	}
+	if pemBlock[len(pemBlock) - 1] != byte('\n') {
```
frobware, line 294: Should we protect against the case where len(pemBlock) is 0?

frobware (changes requested):

melvinjoseph86:
Verified via pre-merge verification workflow, more references related to the test can be found in:
https://bugzilla.redhat.com/show_bug.cgi?id=1894431#c6

pkg/router/template/router.go
Miciah: Indeed!  Thanks!  Fixed.

Miciah:
7f78d34 → 8a3760b fixes some gofmt issues.

PR #377, opened 2022-03-22
NE-822 Don't scale route weight on single service routes

gcs278:
Change to not scale weight to 256 if there is only one service for the route. More information can be found here: [NE-709 Impact of Server Weight on Memory Allocation](https://docs.google.com/document/d/1QX7GpJTWTpBAga4TMrshDCi5mHAd-4-IizcQS1CCMS4/edit)

In a nutshell, scaling the weight to 256 is redundant for single services routes since all servers in the haproxy backend will always have the same weight, so therefore weight 1 == weight 256. Reducing the weight helps with reducing the static memory allocation.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -1286,6 +1286,18 @@ func (r *templateRouter) getActiveEndpoints(serviceUnits map[ServiceUnitKey]int3
 // The above assumes roundRobin scheduling.
 func (r *templateRouter) calculateServiceWeights(serviceUnits map[ServiceUnitKey]int32) map[ServiceUnitKey]int32 {
 	serviceUnitNames := make(map[ServiceUnitKey]int32)
+
+	// If there is only 1 service unit, then always set the weight 1 for all the endpoints.
+	// Scaling the weight to 256 is redundant and causes haproxy to allocate more memory on startup.
+	if len(serviceUnits) == 1 {
+		for key := range serviceUnits {
+			if r.numberOfEndpoints(key) > 0 {
+				serviceUnitNames[key] = 1
+			}
+		}
+		return serviceUnitNames
+	}
+
 	// portion of service weight for each endpoint
 	epWeight := make(map[ServiceUnitKey]float32)
 	// maximum endpoint weight
--- a/pkg/router/template/router_test.go
+++ b/pkg/router/template/router_test.go
@@ -948,6 +948,46 @@ func TestCalculateServiceWeights(t *testing.T) {
 				suKey2: 256,
 			},
 		},
+		{
+			name: "a single service with a single endpoint",
+			serviceUnits: map[ServiceUnitKey][]Endpoint{
+				suKey1: {ep1},
+			},
+			serviceWeights: map[ServiceUnitKey]int32{
+				suKey1: 50,
+			},
+			expectedWeights: map[ServiceUnitKey]int32{
+				suKey1: 1,
+			},
+		},
+		{
+			name: "a single service with a multiple endpoints",
+			serviceUnits: map[ServiceUnitKey][]Endpoint{
+				suKey1: {ep1, ep2},
+			},
+			serviceWeights: map[ServiceUnitKey]int32{
+				suKey1: 50,
+			},
+			expectedWeights: map[ServiceUnitKey]int32{
+				suKey1: 1,
+			},
+		},
+		{
+			name:            "no services with no endpoints",
+			serviceUnits:    map[ServiceUnitKey][]Endpoint{},
+			serviceWeights:  map[ServiceUnitKey]int32{},
+			expectedWeights: map[ServiceUnitKey]int32{},
+		},
+		{
+			name: "service with no endpoint",
+			serviceUnits: map[ServiceUnitKey][]Endpoint{
+				suKey1: {},
+			},
+			serviceWeights: map[ServiceUnitKey]int32{
+				suKey1: 100,
+			},
+			expectedWeights: map[ServiceUnitKey]int32{},
+		},
 	}
 
 	for _, tc := range testCases {


frobware:
- 28 failures to create the sandbox

/retest

frobware:
cluster bootstrap failed

/test e2e-upgrade

pkg/router/template/router.go
```diff
@@ -1286,6 +1286,16 @@ func (r *templateRouter) getActiveEndpoints(serviceUnits map[ServiceUnitKey]int3
 // The above assumes roundRobin scheduling.
 func (r *templateRouter) calculateServiceWeights(serviceUnits map[ServiceUnitKey]int32) map[ServiceUnitKey]int32 {
 	serviceUnitNames := make(map[ServiceUnitKey]int32)
+
+	// If there is only 1 service unit, then set always set the weight 1 for all the endpoints.
```
frobware, line 1290: typo: "then always set"

gcs278: Done

gcs278:
/hold

I think we should wait for perf&scale test.

pkg/router/template/router.go
```diff
@@ -1286,6 +1286,16 @@ func (r *templateRouter) getActiveEndpoints(serviceUnits map[ServiceUnitKey]int3
 // The above assumes roundRobin scheduling.
 func (r *templateRouter) calculateServiceWeights(serviceUnits map[ServiceUnitKey]int32) map[ServiceUnitKey]int32 {
 	serviceUnitNames := make(map[ServiceUnitKey]int32)
+
+	// If there is only 1 service unit, then always set the weight 1 for all the endpoints.
+	// Scaling the weight to 256 is redundant and causes haproxy to allocate more memory on startup.
+	if len(serviceUnits) == 1 {
+		for key := range serviceUnits {
+			serviceUnitNames[key] = 1
+		}
```
Miciah, line 1295: What if the service unit has 0 endpoints?  We should check for that case:
```suggestion
		for key := range serviceUnits {
			if r.numberOfEndpoints(key) > 0 {
				serviceUnitNames[key] = 1
			}
		}
```
Granted, omitting the `r.numberOfEndpoints(key) > 0` check might not change the ultimate result from config template:
```
        {{- range $serviceUnitName, $weight := $cfg.ServiceUnitNames }}
          {{- if ge $weight 0 }}{{/* weight=0 is reasonable to keep existing connections to backends with cookies as we can see the HTTP headers */}}
            {{- with $serviceUnit := index $.ServiceUnits $serviceUnitName }}
              {{- range $idx, $endpoint := processEndpointsForAlias $cfg $serviceUnit (env "ROUTER_BACKEND_PROCESS_ENDPOINTS" "") }}
                {{/* [actual HAProxy config stuff here] */}}
            {{- end }}{{/* end range processEndpointsForAlias */}}
          {{- end }}{{/* end get serviceUnit from its name */}}
        {{- end }}{{/* end range over serviceUnitNames */}}

```
In the template, the effect is the same whether `range $serviceUnitName, $weight := $cfg.ServiceUnitNames` iterates 0 times or whether `range $idx, $endpoint := processEndpointsForAlias $cfg $serviceUnit (env "ROUTER_BACKEND_PROCESS_ENDPOINTS" "")` iterates 0 times.  However, it could cause problems for the dynamic config manager, which has logic like this:
```go
		// As the endpoints have changed, recalculate the weights.
		newWeights := r.calculateServiceWeights(cfg.ServiceUnits)

		// Get the weight for this service unit.
		weight, ok := newWeights[id]
		if !ok {
			weight = 0
		}
```
Anyway, from a strict correctness perspective, I believe `calculateServiceWeights` needs the `r.numberOfEndpoints(key) > 0` check.  

Please add a unit test case to ensure we don't regress:
```diff
diff --git a/pkg/router/template/router.go b/pkg/router/template/router.go
index 1641b0c7..c6a65a79 100644
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -1287,6 +1287,20 @@ func (r *templateRouter) getActiveEndpoints(serviceUnits map[ServiceUnitKey]int3
 func (r *templateRouter) calculateServiceWeights(serviceUnits map[ServiceUnitKey]int32) map[ServiceUnitKey]int32 {
 	serviceUnitNames := make(map[ServiceUnitKey]int32)
 	// portion of service weight for each endpoint
+
+
+	// If there is only 1 service unit, then always set the weight 1 for all the endpoints.
+	// Scaling the weight to 256 is redundant and causes haproxy to allocate more memory on startup.
+	if len(serviceUnits) == 1 {
+		for key := range serviceUnits {
+			if r.numberOfEndpoints(key) > 0 {
+				serviceUnitNames[key] = 1
+			}
+		}
+		return serviceUnitNames
+	}
+
+
 	epWeight := make(map[ServiceUnitKey]float32)
 	// maximum endpoint weight
 	var maxEpWeight float32 = 0.0
diff --git a/pkg/router/template/router_test.go b/pkg/router/template/router_test.go
index 73e77f55..1d88c983 100644
--- a/pkg/router/template/router_test.go
+++ b/pkg/router/template/router_test.go
@@ -873,6 +873,16 @@ func TestCalculateServiceWeights(t *testing.T) {
 		serviceWeights  map[ServiceUnitKey]int32
 		expectedWeights map[ServiceUnitKey]int32
 	}{
+		{
+			name: "service with no endpoint",
+			serviceUnits: map[ServiceUnitKey][]Endpoint{
+				suKey1: {},
+			},
+			serviceWeights: map[ServiceUnitKey]int32{
+				suKey1: 100,
+			},
+			expectedWeights: map[ServiceUnitKey]int32{},
+		},
 		{
 			name: "equally weighted services with same number of endpoints",
 			serviceUnits: map[ServiceUnitKey][]Endpoint{
```

gcs278: Good point, didn't think about a single service with no endpoints and exposing that datapath. Agreed it doesn't functionally changing anything, but keep our code clean for the future.

Done.

gcs278:
/label px-approved
Parent Epic NE-709 has px not needed.

PR #379, opened 2022-04-01
Bug 2025624: Fix certificate reloader

suleymanakbas91:
Ingress router metrics endpoint was still serving old certificates after certificate rotation. This fixes the logic to reload the certificate when there is a change.


Miciah:
Thanks!
/lgtm

PR #380, opened 2022-04-04
Futz with maxconn

frobware:


frobware:
: [sig-network][Feature:Router] The HAProxy router should run even if it has no access to update status [Skipped:Disconnected] [Suite:openshift/conformance/parallel] 

This is interesting given the change to health checks in this PR.

/retest

PR #383, opened 2022-04-14
Bug 2076297: Fix gap in router's handling of graceful shutdowns.

gcs278:
If you quickly start and send a SIGTERM to the router, it will hang indefinitely.
  The call to WaitForCacheSync wasn't listening for a SIGTERM and wouldn't terminate
  if still trying to sync cache.


Miciah:
Very nice fix.  Thanks!
/lgtm

PR #386, opened 2022-04-21
Add timeout http-keep-alive for HAProxy template.

git001:
Should be backported upto 4.8


Miciah:
Thanks for the PR!  This looks like a new feature (new route annotation and configuration option), and we generally don't backport features.  Please note also that HAProxy 2.2.23 changes the effect of `timeout http-keep-alive`, so the behavior wouldn't necessarily be the same on different versions of OpenShift anyway:

>   * Some issues in the H2 multiplexers were fixed. First the GOAWAY frame is
>     no longer sent if SETTINGS were not sent. Then, as announced, the
>     "timeout http-keep-alive" and "timeout http-request" are now respected
>     and work as documented, so that it will finally be possible to force
>     such connections to be closed when no request comes even if they're
>     seeing control traffic such as PING frames. This can typically happen in
>     some server-to-server communications whereby the client application
>     makes use of PING frames to make sure the connection is still alive.

Source: [[ANNOUNCE] haproxy-2.2.23](https://www.mail-archive.com/haproxy@formilux.org/msg42232.html).

Would you be willing to make a PR to add documentation for the new annotation? Route annotations are documented here: https://github.com/openshift/openshift-docs/blob/main/modules/nw-route-specific-annotations.adoc

PR #391, opened 2022-05-25
WIP: haproxy-config.template: Use PROXY protocol v2

Miciah:
* `images/router/haproxy/conf/haproxy-config.template`: Use PROXY protocol v2 for the internal "be_sni" and "be_no_sni" backends and "fe_no_sni" and "fe_sni" frontends.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -282,7 +282,7 @@ frontend public_ssl
 # traffic
 ##########################################################################
 backend be_sni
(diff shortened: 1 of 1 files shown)


jcmoraisjr:
Looks a fair update and it is currently in a good timing (start of a new release). It marginally improves performance and should not have any backward incompatibility - HAProxy itself is handling the changed request, and a proxy-proto aware frontend handles both v1 and v2.

/lgtm

Miciah:
I still need to create a Jira tracker for this PR.

PR #414, opened 2022-08-29
WIP: NE-366: Allow setting log level of router binary

rfredette:


frobware:
It would be great if this could be done dynamically. On customer calls I'd like to temporarily increase the verbosity without restarting the router (because production).

Using something similar to (or in fact) https://github.com/spf13/viper would allow us to drop a config file in /tmp and `openshift-router` could re-read that on `SIGHUP`.

rfredette:
It looks like this failed the image build due to some mirrors not being accessible? Retrying to see if it was a temporary outage.

/test images

PR #421, opened 2022-09-26
OCPBUGS-1730: Bump vendored K8S libraries to 1.25.2

gcs278:
Update k8s.io/api, k8s.io/apimachinery, k8s.io/apiserver, and k8s.io/client-go to v0.25.2 which maps to the overall 1.25 kubernetes release. Also bump github.com/openshift/library-go for its 1.25 support. All other vendor updates were automatically included as a part of these updates.


gcs278:
nothing specific about router and `OVNKubernetesControllerDisconnectedSouthboundDatabase` issues
/retest

images/router/haproxy/conf/haproxy-config.template.listen
```diff
@@ -0,0 +1,819 @@
+{{/*
```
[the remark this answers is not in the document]
gcs278: Whoops I have no clue where this came from. Thanks for catching this, I don't think this is part of the vendor bump... Maybe it was just sitting in my repo when I did the bump.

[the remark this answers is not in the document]
gcs278: Ah yea now I know. I created this file to test an alternative haproxy configuration. I didn't catch that it was still hanging out.

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/fortytw2/leaktest v1.3.0
 	github.com/fsnotify/fsnotify v1.4.9
 	github.com/gocarina/gocsv v0.0.0-20190927101021-3ecffd272576
-	github.com/google/go-cmp v0.5.5
+	github.com/google/go-cmp v0.5.8
 	github.com/haproxytech/config-parser/v4 v4.0.0-rc1
 	github.com/openshift/api v0.0.0-20211201130627-34f305c3af47
```
brandisher, line 14: Does this need a bump as well? It looks like it was done for cluster-ingress-operator in [#829](https://github.com/openshift/cluster-ingress-operator/pull/829)

gcs278: Sorry just saw this comment. Hmmm, that's strange. I didn't explicitly do the bump of openshift/api in cluster-ingress-operator and nor did I do it here. Not sure why this didn't bump, but maybe it's a dependency chain that openshift/router is not triggering?
Do you think I should bump openshift/api to it's K8S API bump regardless? I'm not sure.

brandisher: I'd be inclined to say "no" only because if the router doesn't need the bump, then we probably don't gain anything by bumping it aside from keeping the future delta of the API changes smaller. WDYT @Miciah ?

gcs278:
Lots of errors, installation probably went wrong.
/retest

brandisher:
Aside from the one pending (albiet non-blocking) question [here](https://github.com/openshift/router/pull/421#discussion_r982819177) these changes look good to me and I don't see any issues. If we do indeed need to bump `openshift/api` it can be done in another PR.

/lgtm

gcs278:
@Miciah this should be ready for you to review and approve

go.mod
Miciah: Bumping openshift/api isn't necessary.  There are no changes in it that would directly affect openshift/router.

PR #438, opened 2022-12-25
NE-1141: Adds logic for setting and deleting headers via Ingress Operator CR and Route Object.

miheer:
`images/router/haproxy/conf/haproxy-config.template`  Sets the
http-response and http-request stanzas to set or delete custom HTTP
Response/Request headers.
`pkg/cmd/infra/router/template.go` This consists the logic to get the
value from the env var from router deployment and use it for
setting/deleting HTTP response/request headers.
`pkg/router/template/plugin.go` Adds new fields in the
TemplatePluginConfig struct to set or delete HTTP request/reponse
headers. These are needed to set the values we get from router
deployment env var.
`pkg/router/template/router.go` This consists of the logic extracting
header values from the route object and then set/delete
HTTP request/reponse headers.
`pkg/router/template/router_test.go` This consists of units test cases
to the logic to extract headers from route object and set/delete HTTP
request/reponse headers.
`pkg/router/template/types.go` This adds two fields HTTPResponseHeaders
and HTTPRequestHeaders of type []HTTPHeader in the ServiceAliasConfig
which is used in the haproxy-config.template to add the set/delete
http-reponse and http-request stanzas for the header values fetched from
the route object.
This also add a struct called HTTPHeader which helps setting the HTTP
response/request values in the http-reponse/http-request stanzas in the
haproxy-config.template.

Jira Number https://issues.redhat.com/browse/NE-982


gcs278:
Note: I've created https://github.com/openshift/router/pull/451 for the K8S rebase, better for us to do it as a separate PR. This PR will need rebasing after it is submitted.

gcs278:
https://github.com/openshift/router/pull/451 has merged, which updated the K8S API versions to 1.26.

pkg/router/template/router.go
```diff
@@ -1376,6 +1430,7 @@ func configsAreEqual(config1, config2 *ServiceAliasConfig) bool {
 		config1.RoutingKeyName == config2.RoutingKeyName &&
 		config1.IsWildcard == config2.IsWildcard &&
 		config1.VerifyServiceHostname == config2.VerifyServiceHostname &&
+		reflect.DeepEqual(config1.HTTPResponseHeaders, config2.HTTPResponseHeaders) &&
```
frobware, line 1433: Shouldn't we also be checking for request headers too?

pkg/router/template/types.go
```diff
@@ -74,6 +74,12 @@ type ServiceAliasConfig struct {
 
 	// ActiveEndpoints is a count of the route endpoints that are part of a service unit with a non-zero weight
 	ActiveEndpoints int
+
+	// For route based setting of http response headers
+	HTTPResponseHeaders []HTTPHeader
+
+	// For route based setting of http response headers
```
frobware, line 81: Typo: should be "request" headers.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -612,6 +661,23 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
           {{- end }}{{/* hsts header */}}
         {{- end }}{{/* is "edge" or "reencrypt" */}}
 
+        {{- if matchValues (print $cfg.TLSTermination) "edge" "reencrypt" }}
```
candita, line 664: Why do you only need to check for `TLSTermination` type here? Why not in the other places?

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +294,77 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+				// must be a valid token.
+				if !validTokenRE.MatchString(string(headerName)) {
+					return captureHeaders, fmt.Errorf("invalid HTTP header name: %v", headerName)
+				}
+				headerValue, err := b64.StdEncoding.DecodeString(parts[1])
+				if err != nil {
+					return captureHeaders, fmt.Errorf("failed to decode base64 encoding3: %v", headerValue)
+				}
+				sanitizedHeaderValue := sanitizeHeaderValue(string(headerValue))
+				action, err := b64.StdEncoding.DecodeString(parts[2])
```
candita, line 330: It would be best to make sure the `action` is either set or delete, just in case someone misconfigured.

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +294,77 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+	return "'" + santizeHeaderValue + "'"
+}
+
+func parseHeadersToBeSetOrDeleted(in string) ([]templateplugin.HTTPHeader, error) {
+	var captureHeaders []templateplugin.HTTPHeader
+	var capture templateplugin.HTTPHeader
+	/*encodedIn, err := b64.StdEncoding.DecodeString(in)
+	if err != nil {
+		return captureHeaders, fmt.Errorf("failed to decode base64 encoding1: %v", encodedIn)
+	}*/
```
candita, line 308: Remove the commented section if not needed.

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +294,77 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
 	return captureHeaders, nil
 }
 
+func sanitizeHeaderValue(headerValue string) string {
```
candita, line 297: Is it haproxy or something else that doesn't allow unescaped `'` in the header value?

candita: Why did you need to do it here as opposed to where you constructed the value in the ingress operator?

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +294,77 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+	if len(in) > 0 {
+		for _, header := range strings.Split(in, ",") {
+			parts := strings.Split(header, ":")
+			if len(parts) < 2 {
+				return captureHeaders, fmt.Errorf("invalid HTTP header input specification: %v", header)
+			}
+			if len(parts) == 3 {
+				headerName, err := b64.StdEncoding.DecodeString(parts[0])
+				if err != nil {
+					return captureHeaders, fmt.Errorf("failed to decode base64 encoding2: %v", headerName)
```
candita, line 318:
```suggestion
					return captureHeaders, fmt.Errorf("failed to decode base64 encoding: %v", parts[0])
```

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +294,77 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+				}
+				// RFC 2616, section 4.2, states that the header name
+				// must be a valid token.
+				if !validTokenRE.MatchString(string(headerName)) {
+					return captureHeaders, fmt.Errorf("invalid HTTP header name: %v", headerName)
+				}
+				headerValue, err := b64.StdEncoding.DecodeString(parts[1])
+				if err != nil {
+					return captureHeaders, fmt.Errorf("failed to decode base64 encoding3: %v", headerValue)
```
candita, line 327:
```suggestion
					return captureHeaders, fmt.Errorf("failed to decode base64 encoding: %v", parts[1])
```

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +294,77 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+				}
+				sanitizedHeaderValue := sanitizeHeaderValue(string(headerValue))
+				action, err := b64.StdEncoding.DecodeString(parts[2])
+				if err != nil {
+					return captureHeaders, fmt.Errorf("failed to decode base64 encoding4: %v", action)
```
candita, line 332:
```suggestion
					return captureHeaders, fmt.Errorf("failed to decode base64 encoding: %v", parts[2])
```

pkg/cmd/infra/router/template.go
```diff
@@ -393,6 +471,18 @@ func (o *TemplateRouterOptions) Complete() error {
 	}
 	o.CaptureHTTPResponseHeaders = captureHTTPResponseHeaders
 
+	httpResponseHeaders, err := parseHeadersToBeSetOrDeleted(o.HTTPResponseHeadersString)
+	if err != nil {
+		return err
+	}
+	o.HTTPResponseHeaders = httpResponseHeaders
```
candita, line 478: `httpResponseHeaders` can be nil here, do you still want to set the `o.HTTPResponseHeaders` to nil if so?

pkg/router/template/router.go
```diff
@@ -922,16 +932,27 @@ func (r *templateRouter) createServiceAliasConfig(route *routev1.Route, backendK
 			activeServiceUnits++
 		}
 	}
+	var httpResponseHeadersList []HTTPHeader
+	var httpRequestHeadersList []HTTPHeader
+	if route.Spec.HTTPHeaders.Actions.Response != nil && len(route.Spec.HTTPHeaders.Actions.Response) != 0 {
```
candita, line 937: What happens if `route.Spec.HTTPHeaders.Actions` was omitted?

pkg/router/template/router.go
```diff
@@ -990,6 +1011,35 @@ func (r *templateRouter) createServiceAliasConfig(route *routev1.Route, backendK
...
+				Name:   value.Name,
+				Action: string(value.Action.Type),
+			}
+			httpHeadersList = append(httpHeadersList, deleteValue)
+		}
+	}
+	return httpHeadersList
+}
+
+func sanitizeHeaderValue(headerValue string) string {
```
candita, line 1038: This is already declared in template.go, why not reuse it?

pkg/router/template/router.go
```diff
@@ -990,6 +1011,35 @@ func (r *templateRouter) createServiceAliasConfig(route *routev1.Route, backendK
 	return &config
 }
 
+func getHeadersList(httpHeaderList []routev1.RouteHTTPHeader) []HTTPHeader {
+	var httpHeadersList []HTTPHeader
```
candita, line 1015: How do you ensure that these headers don't override the headers that the cluster admin set in the ingress controller?  What if the cluster admin doesn't want users to set request/response headers to something different than what the admin set?

images/router/haproxy/conf/haproxy-config.template
candita: Also, I understand that the cfg.HTTP*Headers are from Route.spec.  I think you may need to find some way to merge the header manipulation instructions given by IngressSpec with those given in the Route spec.  What if the cluster admin adds a header and the Route.spec deletes it?

candita (changes requested):
Overall, I think there is a design flaw in allowing the user to overwrite the configuration of the admin.  Have you discussed this with other team members? 

There are also some nits.

pkg/router/template/router.go
miheer: It can't be omitted the API is not structured that way. I already told you it is not a pointer.

pkg/router/template/router.go
miheer: The ingress controller updates the front end sections only and the route updates the plain backend section in the haproxy template. Both have different use cases.

miheer:
>> Also, I understand that the cfg.HTTP*Headers are from Route.spec. I think you may need to find some way to merge the header manipulation instructions given by IngressSpec with those given in the Route spec. What if the cluster admin adds a header and the Route.spec deletes it?

The ingress controller updates the front end sections only and the route updates the plain backend section in the haproxy template. Both have different use cases.

candita:
@miheer please add a description to document this change, with Jira number, etc.

images/router/haproxy/conf/haproxy-config.template
candita: I think you need to do something like what was done for hsts, see https://github.com/openshift/router/blob/master/images/router/haproxy/conf/haproxy-config.template#L610-L611

pkg/cmd/infra/router/template.go
miheer: Yes you need to perform escaping and quoting. Details are mentioned in the EP.

pkg/cmd/infra/router/template.go
miheer: sorry my bad

images/router/haproxy/conf/haproxy-config.template
miheer: we have already discussed this that route owner will be able to over-ride the header if same header was redefined.

pkg/cmd/infra/router/template.go
miheer: I am not sure because we allow both actions to be added at the same time. Also both actions have different way to provide a value.

miheer:
> Overall, I think there is a design flaw in allowing the user to overwrite the configuration of the admin. Have you discussed this with other team members?
> 
> There are also some nits.

As discussed the route owner would be to over-ride the header value set by admin via ingress controller.

pkg/router/template/router.go
candita:
```suggestion
	if route.Spec.HTTPHeaders.Actions.Response != "" {
```

images/router/haproxy/conf/haproxy-config.template
candita: Well, what if this template adds a header, and Route.spec deletes it?  Such as those in https://github.com/openshift/router/pull/438/files#diff-21bd64a4ad8a7927e9d59e99e09628efc884fe6d6291e50eb33c5deb3d2a79f0R354-R364
Shouldn't those be off-limits?

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -233,6 +233,22 @@ frontend public
   acl secure_redirect base,map_reg_int(/var/lib/haproxy/conf/os_route_http_redirect.map) -m bool
   redirect scheme https if secure_redirect
 
+    {{- range $idx, $http_request_header := .HTTPRequestHeaders }}
```
candita, line 236: I don't really understand why we need this on `frontend public`.  Can you help me understand?

candita: Also, why don't we forbid adding "Proxy" header, which we explicitly delete in https://github.com/openshift/router/pull/438/files#diff-21bd64a4ad8a7927e9d59e99e09628efc884fe6d6291e50eb33c5deb3d2a79f0R221 ?  We also make sure the "Host" is lower case, and we aren't disallowing the user to change that?

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -348,6 +364,22 @@ frontend fe_sni
   http-request set-header X-SSL-Client-DER       %{+Q}[ssl_c_der,base64]
     {{- end }}
 
+    {{- range $idx, $http_request_header := .HTTPRequestHeaders }}
```
candita, line 367: Again, there are so many headers we explicitly set here, why would we allow the user to change those?

miheer: This section is set by the admin via ingress controller so I think this shall be his responsibility  to set the headers correctly.

images/router/haproxy/conf/haproxy-config.template
miheer: Route owner can modify the header at the app level only as he can't modify the front end sections in the haproxy.template. In short the scope for the route owner to set this is at backend level i.e for a specific route which points to a specific backend.

pkg/router/template/router.go
miheer: response is an array

images/router/haproxy/conf/haproxy-config.template
miheer: or may be we can add this code just before https://github.com/openshift/router/pull/438/files#diff-21bd64a4ad8a7927e9d59e99e09628efc884fe6d6291e50eb33c5deb3d2a79f0R221 ?

images/router/haproxy/conf/haproxy-config.template
miheer: other option is to add this code just before https://github.com/openshift/router/pull/438/files#diff-21bd64a4ad8a7927e9d59e99e09628efc884fe6d6291e50eb33c5deb3d2a79f0R354

images/router/haproxy/conf/haproxy-config.template
miheer: Also another option is we can add this code just before set-header stanzas to avoid over-ride of headers set by template.

candita: Whatever you do here should be reflected in the enhancement proposal.  I don't think this last option is any better, because the user may still try to set a header that they are not permitted to modify/delete.  IMO, there should be a validation response that tells the user when they cannot modify/delete a specific header, as there is already for HSTS, but the whole list should be there.  If not, then explain why not in the enhancement proposal.

If we don't care about the user experience in trying to modify/delete headers that they shouldn't be permitted to, then put that in the enhancement proposal.

miheer: added in EP

images/router/haproxy/conf/haproxy-config.template
candita, line 1: Why don't you set the headers on `frontend public_ssl` ?

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,77 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+	var capture templateplugin.HTTPHeader
+	encodedIn, err := url.PathUnescape(in)
+	if err != nil {
+		return captureHeaders, fmt.Errorf("failed to decode percent encoding: %v", encodedIn)
+	}
+	if len(in) > 0 {
+		for _, header := range strings.Split(in, ",") {
+			parts := strings.Split(header, ":")
+			if len(parts) < 2 {
+				return captureHeaders, fmt.Errorf("invalid HTTP header input specification: %v", header)
```
candita, line 312: You should have a unit test for anything that could produce an error.  I think you need three more unit tests, as there is only one: `TestPercentDecodingForHeaders`.

pkg/cmd/infra/router/template_test.go
```diff
@@ -0,0 +1,41 @@
...
+			expectedValue: []templateplugin.HTTPHeader{
+
+				{Name: "X-Frame-Options", Value: "'DENY'", Action: "Set"},
+				{Name: "X-XSS-Protection", Value: "'1;mode=block'", Action: "Set"},
+				{Name: "x-forwarded-client-cert", Value: "'%{+Q}[ssl_c_der,base64]'", Action: "Set"},
+				{Name: "X-Frame-Options", Action: "Delete"},
+				{Name: "X-XSS-Protection", Action: "Delete"},
+			},
+			expectError: false,
+		},
```
candita, line 28: As mentioned above, you should add test cases for any of the new errors you measure.

pkg/router/template/router.go
```diff
@@ -123,7 +123,9 @@ type templateRouter struct {
 	// haveClientCA specifies if the user provided their own CA for client auth in mTLS
 	haveClientCA bool
 	// haveCRLs specifies if the crl file has been generated for client auth
-	haveCRLs bool
+	haveCRLs            bool
+	httpResponseHeaders []HTTPHeader
```
candita, line 127: Can you add comments for these two new fields?

pkg/router/template/router.go
```diff
@@ -191,7 +195,9 @@ type templateData struct {
 	// HaveClientCA specifies if the user provided their own CA for client auth in mTLS
 	HaveClientCA bool
 	// HaveCRLs specifies if the crl file is present
-	HaveCRLs bool
+	HaveCRLs            bool
+	HTTPResponseHeaders []HTTPHeader
```
candita, line 199: Comments here too.

candita:
I have a few questions, and this needs more unit test, but otherwise:
/lgtm

images/router/haproxy/conf/haproxy-config.template
Miciah: The connection is still encrypted in `public_ssl`, so it is not possible to examine or modify the HTTP session there.  The `fe_sni` and `fe_no_sni` frontends terminate TLS.

Here's my attempt at a diagram showing the transitions between cleartext HTTP and TLS:
```mermaid
graph LR;
  client-->|HTTP|public;
  public-->|HTTP|be_http["be_http:namespace:application<br/>(non-TLS route)"];
  client-->|TLS|public_ssl;
  fe_sni-->|HTTP|be_edge["be_edge_http:namespace:application<br/>(edge-terminated TLS)"];
  fe_no_sni-->|HTTP|be_edge;
  public_ssl-->|TLS|be_sni;
  public_ssl-->|TLS|be_no_sni;
  be_sni-->|TLS|fe_sni;
  be_no_sni-->|TLS|fe_no_sni;
  fe_sni-->|TLS|be_reencrypt["be_secure_http:namespace:application<br/>(reencrypt TLS)"];
  fe_sni-->|TLS|be_reencrypt;
  public_ssl-->|TLS|be_tcp["be_tcp:namespace:application<br/>(passthrough TLS)"];
  classDef cleartext fill:#1f1
  classDef encrypted fill:#f11,color:#fff;
  class public,fe_sni,fe_no_sni,be_http,be_edge cleartext;
  class public_ssl,be_sni,be_no_sni,be_reencrypt,be_tcp encrypted;
```

pkg/cmd/infra/router/template.go
candita: You don't necessarily need to test all possible paths, but at least test for 1. token errors and 2. format errors that produce len(parts) < 2 and 3. error on decode.

pkg/cmd/infra/router/template_test.go
```diff
@@ -0,0 +1,41 @@
...
+	}
```
Miciah, line 29: I agree with Candace's comment about this test.  You have a list of test cases, but you only have one test case in that list.  The test case that you do have verifies that the function parses valid input.  You need to add some test cases with invalid input, so that your test verifies that `parseHeadersToBeSetOrDeleted` behaves correctly when given invalid input.  There are several places in `parseHeadersToBeSetOrDeleted` that return errors, so you should have some test cases where `expectError` is true.

pkg/cmd/infra/router/template_test.go
candita: Some place where `expectError: true`.

pkg/router/template/router.go
```diff
@@ -1007,6 +1027,35 @@ func (r *templateRouter) createServiceAliasConfig(route *routev1.Route, backendK
...
+			}
+			httpHeadersList = append(httpHeadersList, deleteValue)
+		}
+	}
+	return httpHeadersList
+}
+
+func sanitizeHeaderValue(headerValue string) string {
+	santizeHeaderValue := strings.ReplaceAll(headerValue, "'", `'\''`)
+	return "'" + santizeHeaderValue + "'"
```
Miciah, line 1056:
```suggestion
	sanitizeHeaderValue := strings.ReplaceAll(headerValue, "'", `'\''`)
	return "'" + sanitizeHeaderValue + "'"
```

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,77 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
 	return captureHeaders, nil
 }
 
+func sanitizeHeaderValue(headerValue string) string {
+	santizeHeaderValue := strings.ReplaceAll(headerValue, "'", `'\''`)
+	return "'" + santizeHeaderValue + "'"
```
Miciah, line 298:
```suggestion
	sanitizeHeaderValue := strings.ReplaceAll(headerValue, "'", `'\''`)
	return "'" + sanitizeHeaderValue + "'"
```

pkg/cmd/infra/router/template.go
candita: So, a test for the actual value Set or Delete, not a typo like Seet or Del.

pkg/cmd/infra/router/template.go
miheer: Hi @Miciah ! I dont get what is changed here.

pkg/router/template/router.go
miheer: Hi @Miciah ! I dont get what is changed here.

pkg/cmd/infra/router/template.go
frobware: Missing `i` in san**i**tize.

pkg/cmd/infra/router/template_test.go
miheer: ack added tests to check error messages.

pkg/router/template/router_test.go
```diff
@@ -382,6 +386,72 @@ func TestCreateServiceAliasConfig(t *testing.T) {
 				CACertificate:            "ghi",
 				DestinationCACertificate: "jkl",
 			},
+			HTTPHeaders: routev1.RouteHTTPHeaders{
```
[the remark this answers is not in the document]
miheer: done

images/router/haproxy/conf/haproxy-config.template
miheer: I have addressed this in the "Risks and Mitigations" section of the EP https://github.com/openshift/enhancements/pull/1296.

pkg/cmd/infra/router/template.go
miheer: yes added.

candita:
/retitle NE-1141: Adds logic for setting and deleting headers via Ingress Operator CR and Route Object.

pkg/router/template/router.go
candita: s a n t i z e is a typo

pkg/router/template/types.go
```diff
@@ -236,6 +242,18 @@ type CaptureHTTPHeader struct {
 	MaxLength int
 }
 
+// HTTPHeader specifies an HTTP header that should be set/replace/append
```
candita, line 245:
```suggestion
// HTTPHeader specifies an HTTP header that should be set or deleted.
```

pkg/router/template/types.go
```diff
@@ -236,6 +242,18 @@ type CaptureHTTPHeader struct {
...
+type HTTPHeader struct {
+	// Name specifies an HTTP header name.
+	Name string
+
+	// Value specifies the header value.
+	Value string
+
+	// Action specifies the action to be performed.
+	Action string
```
candita, line 254: Better to set this up with a `type ActionType string` 
```suggestion
	Action ActionType
}
// ...
type ActionType string
// ..
const (
        // ...
        ActionTypeDelete ActionType = "Delete"
        // ...
       ActionTypeSet ActionType = "Set"
)
```

pkg/router/template/types.go
```diff
@@ -74,6 +74,12 @@ type ServiceAliasConfig struct {
 
 	// ActiveEndpoints is a count of the route endpoints that are part of a service unit with a non-zero weight
 	ActiveEndpoints int
+
+	// For route based setting of http response headers
```
candita, line 78: nit: upper case all the `http` references

pkg/router/template/router_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/md5"
 	"fmt"
+	"github.com/google/go-cmp/cmp"
```
candita, line 7: Reorder this where it belongs.

pkg/router/template/router_test.go
```diff
@@ -360,6 +361,9 @@ func TestCreateServiceAliasConfig(t *testing.T) {
 	serviceName := "TestService"
 	serviceWeight := int32(0)
 
+	var headerNameXFrame string = "X-Frame-Options"
+	var headerNameXSS string = "X-XSS-Protection"
+
```
candita, line 366: These tests should be in their own function, `TestCustomHeaders` or something like that.

pkg/router/template/router.go
```diff
@@ -124,6 +124,10 @@ type templateRouter struct {
 	haveClientCA bool
 	// haveCRLs specifies if the crl file has been generated for client auth
 	haveCRLs bool
+	// httpResponseHeaders allows users to set/delete custom HTTP response
```
candita, line 127:
```suggestion
	// httpResponseHeaders allows users to set or delete custom HTTP response headers.
```

pkg/router/template/router.go
```diff
@@ -124,6 +124,10 @@ type templateRouter struct {
...
+	httpResponseHeaders []HTTPHeader
+	// httpRequestHeaders allows users to set/delete custom HTTP request.
```
candita, line 129:
```suggestion
	// httpRequestHeaders allows users to set or delete custom HTTP request headers.
```

pkg/router/template/router.go
candita: Same suggestion as above.

pkg/router/template/router_test.go
candita: Never mind, I see that you altered the `createServiceAlias` function to do this.  Hm.

pkg/cmd/infra/router/template.go
```diff
@@ -126,6 +126,10 @@ type TemplateRouter struct {
 	CaptureHTTPCookie                   *templateplugin.CaptureHTTPCookie
 	HTTPHeaderNameCaseAdjustmentsString string
 	HTTPHeaderNameCaseAdjustments       []templateplugin.HTTPHeaderNameCaseAdjustment
+	HTTPResponseHeadersString           string
```
candita, line 129: I'd prefer if there were different names for the ingress controller headers vs the route headers, especially when you need to handle these and overrides in the template.  Could you rename the variables to have prefixes `Controller` or `Route` where applicable?

pkg/cmd/infra/router/template.go
```diff
@@ -182,6 +186,8 @@ func (o *TemplateRouter) Bind(flag *pflag.FlagSet) {
...
 	flag.StringVar(&o.CaptureHTTPCookieString, "capture-http-cookie", env("ROUTER_CAPTURE_HTTP_COOKIE", ""), "Name and maximum length of HTTP cookie that should be captured for logging.  The argument must have the following form: name:maxLength. Append '=' to the name to indicate that an exact match should be performed; otherwise a prefix match will be performed.  The value of first cookie that matches the name is captured.")
 	flag.StringVar(&o.HTTPHeaderNameCaseAdjustmentsString, "http-header-name-case-adjustments", env("ROUTER_H1_CASE_ADJUST", ""), "A comma-delimited list of HTTP header names that should have their case adjusted. Each item must be a valid HTTP header name and should have the desired capitalization.")
+	flag.StringVar(&o.HTTPResponseHeadersString, "set-delete-http-response-header", env("ROUTER_HTTP_RESPONSE_HEADERS", ""), "A comma-delimited list of HTTP response header names and values that should be set/deleted.")
+	flag.StringVar(&o.HTTPRequestHeadersString, "set-delete-http-request-header", env("ROUTER_HTTP_REQUEST_HEADERS", ""), "A comma-delimited list of HTTP request header names and values that should be set/deleted.")
```
candita, line 190: I don't think these flags are discussed in the EP.  These flags would only be applicable to the controller headers, right?

pkg/cmd/infra/router/template.go
candita: If you already do it elsewhere, it shouldn't be needed here.

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,77 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+	var captureHeaders []templateplugin.HTTPHeader
+	var capture templateplugin.HTTPHeader
+	var err error
+	if len(in) == 0 {
+		return captureHeaders, fmt.Errorf("encoded header string not present.")
+	}
+	if len(in) > 0 {
+		for _, header := range strings.Split(in, ",") {
+			parts := strings.Split(header, ":")
+			if len(parts) < 2 {
```
candita, line 311:
```suggestion
                       num := len(parts)
			if num < 2 || num > 3 {
```

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,77 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+				return captureHeaders, fmt.Errorf("invalid HTTP header input specification: %v", header)
+			}
+			if len(parts) == 3 {
```
candita, line 314:
```suggestion
			else {
```

candita, line 314: As long as you check that len(parts) > 2 and <= 3 above
```suggestion
			} else {

```

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,77 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+					return captureHeaders, fmt.Errorf("failed to decode percent encoding: %v", parts[2])
+				}
+				capture = templateplugin.HTTPHeader{
+					Name:   headerName,
+					Value:  sanitizedHeaderValue,
+					Action: action,
+				}
+				captureHeaders = append(captureHeaders, capture)
+			}
+			if len(parts) == 2 {
```
candita, line 340: As long as you check that len(parts) is either 2 or 3 above
```suggestion
			} else {
```

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,77 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+				}
+				headerValue, err := url.QueryUnescape(parts[1])
+				if err != nil {
+					return captureHeaders, fmt.Errorf("failed to decode percent encoding: %v", parts[1])
+				}
+				sanitizedHeaderValue := sanitizeHeaderValue(headerValue)
+				action, err := url.QueryUnescape(parts[2])
+				if err != nil {
+					return captureHeaders, fmt.Errorf("failed to decode percent encoding: %v", parts[2])
+				}
```
candita, line 332: It looks like this section is identical to the section below and should be refactored into a function.

pkg/cmd/infra/router/template.go
Miciah: These fields in `TemplateRouter` have configuration that ultimately came from the IngressController object.  The route configuration is on the `ServiceAliasConfig` object.  See <https://github.com/openshift/enhancements/blob/master/enhancements/ingress/set-delete-http-headers.md#implementation-detailsnotesconstraints>.

pkg/cmd/infra/router/template.go
Miciah: `ROUTER_HTTP_REQUEST_HEADERS` and `ROUTER_HTTP_RESPONSE_HEADERS` are described under <https://github.com/openshift/enhancements/blob/master/enhancements/ingress/set-delete-http-headers.md#implementation-detailsnotesconstraints>.  The corresponding flags are not explicitly mentioned on the EP, but we usually add a flag for each environment variable that is used by the controller itself (as opposed to variables that are used in the template).

pkg/router/template/router_test.go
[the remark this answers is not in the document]
candita: I don't see that, have you pushed it?

pkg/router/template/router.go
```diff
@@ -253,10 +263,11 @@ func newTemplateRouter(cfg templateRouterCfg) (*templateRouter, error) {
...
 		httpHeaderNameCaseAdjustments: cfg.httpHeaderNameCaseAdjustments,
-
-		metricReload:        metricsReload,
-		metricReloadFailure: metricReloadFailure,
-		metricWriteConfig:   metricWriteConfig,
+		httpResponseHeaders:           cfg.httpResponseHeaders,
+		httpRequestHeaders:            cfg.httpRequestHeaders,
+		metricReload:                  metricsReload,
+		metricReloadFailure:           metricReloadFailure,
+		metricWriteConfig:             metricWriteConfig,
```
Miciah, line 270: Why did you delete the blank line?  If you keep the blank line, that keeps things logically grouped, and it makes the diff smaller:
```suggestion
		httpResponseHeaders:           cfg.httpResponseHeaders,
		httpRequestHeaders:            cfg.httpRequestHeaders,

		metricReload:        metricsReload,
		metricReloadFailure: metricReloadFailure,
		metricWriteConfig:   metricWriteConfig,
```

pkg/router/template/router_test.go
```diff
@@ -391,11 +461,14 @@ func TestCreateServiceAliasConfig(t *testing.T) {
...
 	}
+	httpResponseHeadersList := []HTTPHeader{{Name: "X-Frame-Options", Value: "'DENY'", Action: "Set"}, {Name: "X-XSS-Protection", Value: "'1;mode=block'", Action: "Set"},
+		{Name: "X-Frame-Options", Action: "Delete"}, {Name: "X-XSS-Protection", Action: "Delete"}}
+	httpRequestHeadersList := []HTTPHeader{{Name: "Accept", Value: "'text/plain,text/html'", Action: "Set"}, {Name: "x-client", Value: `'"abc"\ '\''def'\'''`, Action: "Set"}, {Name: "Accept-Encoding", Action: "Delete"}}
 
 	// Basic sanity, validate more fields as necessary
 	if config.Host != route.Spec.Host || config.Path != route.Spec.Path || !compareTLS(route, config, t) ||
 		config.PreferPort != route.Spec.Port.TargetPort.String() || !reflect.DeepEqual(expectedSUs, config.ServiceUnits) ||
-		config.ActiveServiceUnits != 0 {
+		!cmp.Equal(config.HTTPResponseHeaders, httpResponseHeadersList) || !cmp.Equal(config.HTTPRequestHeaders, httpRequestHeadersList) || config.ActiveServiceUnits != 0 {
```
Miciah, line 471: This line is rather long.  If you format the lines this way, you make both the diff and the patched code easier to read:
```suggestion
		config.ActiveServiceUnits != 0 ||
		!cmp.Equal(config.HTTPResponseHeaders, httpResponseHeadersList) ||
		!cmp.Equal(config.HTTPRequestHeaders, httpRequestHeadersList) {
```

pkg/router/template/types.go
Miciah: :+1:, and please also put the field name as the first word in the comment, and ensure you end the sentence with a period:
```suggestion
	// HTTPResponseHeaders has route-specific custom HTTP response headers.

File: images/router/haproxy/conf/haproxy-config.template [hunk 39]
```diff
@@ -624,6 +672,21 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
           {{- end }}{{/* hsts header */}}
         {{- end }}{{/* is "edge" or "reencrypt" */}}
 
+          {{- range $idx, $http_request_header := $cfg.HTTPRequestHeaders }}
```
candita, line 675: There is a block above:
`{{- if matchValues (print $cfg.TLSTermination) "edge" "reencrypt" }}`, and I think the code that looks at headers needs to be within that block, not outside of it, since this does not apply to passthrough termination.

pkg/cmd/infra/router/template.go
rfredette, line 311: There doesn't seem to be any handling for if `len(parts) > 3`. It looks like the entire malformed item will get silently dropped and the list of header actions will continue to be processed. That wouldn't cause a crash or anything severe, but if `len(parts) < 2` causes an error to be returned, I'd expect `len(parts) > 3` to also return an error.

This looks like a reasonable place to use `switch`/`case` instead of a series of `if` statements, and it would also be easy to get the behavior I'm suggesting. For example:
```go
switch len(parts) {
case 3:
    // parse header:value:action
case 2:
    // parse header:action
default:
    // return error
}
```

pkg/router/template/router.go
miheer: ok done.

pkg/router/template/router.go
miheer: ok done.

pkg/cmd/infra/router/template.go
miheer: ok implemented switch.

pkg/cmd/infra/router/template.go
miheer: implemented switch

pkg/cmd/infra/router/template.go
miheer: done

pkg/router/template/types.go
miheer: done

pkg/router/template/types.go
miheer: done.

pkg/router/template/types.go
miheer: used the Action Type of the API.

pkg/router/template/router.go
miheer: ok

pkg/router/template/router_test.go
miheer: ok

images/router/haproxy/conf/haproxy-config.template
miheer: @candita line https://github.com/openshift/router/blob/5e9cc11475d5c5b1b5fa31a1c0d069163a62cc10/images/router/haproxy/conf/haproxy-config.template#L528  already checks it.

pkg/cmd/infra/router/template.go
miheer: @rfredette nice suggestion. thanks!

pkg/router/template/router_test.go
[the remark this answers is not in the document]
miheer: I added some values.
Even though a malformed value is passed the router wont crash as it will be escaped and quoted.

pkg/cmd/infra/router/template.go
miheer: i am not sure. please share your ideas. I am now using switch.

images/router/haproxy/conf/haproxy-config.template
candita: Lines 662 and 669 check it again, so it looks like a mistake that it isn't checked here.

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,83 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+					}
+					headerValue, err := url.QueryUnescape(parts[1])
+					if err != nil {
+						return captureHeaders, fmt.Errorf("failed to decode percent encoding: %v", parts[1])
+					}
+					sanitizedHeaderValue := templateplugin.SanitizeHeaderValue(headerValue)
+					action, err := url.QueryUnescape(parts[2])
+					if err != nil {
+						return captureHeaders, fmt.Errorf("failed to decode percent encoding: %v", parts[2])
+					}
```
candita, line 336: Isn't this just going to be "Set" or "Delete"?  Why do you need to decode?  Why not check for those two types and report and error if there's something different?

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,83 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+				{
+					headerName, err := url.QueryUnescape(parts[0])
+					if err != nil {
+						return captureHeaders, fmt.Errorf("failed to decode percent encoding: %v", parts[0])
+					}
+					// RFC 2616, section 4.2, states that the header name
+					// must be a valid token.
+					if !validTokenRE.MatchString(headerName) {
+						return captureHeaders, fmt.Errorf("invalid HTTP header name: %v", headerName)
+					}
```
candita, line 354: This code is called twice, why not refactor into a function?

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,83 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+	if len(in) == 0 {
+		return captureHeaders, fmt.Errorf("encoded header string not present.")
+	}
+	if len(in) > 0 {
+		for _, header := range strings.Split(in, ",") {
+			parts := strings.Split(header, ":")
+			num := len(parts)
+			switch num {
+			default:
+				{
```
candita, line 314: You don't need brackets {} around each switch case's code block.

pkg/cmd/infra/router/template_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"fmt"
+	"github.com/google/go-cmp/cmp"
+	templateplugin "github.com/openshift/router/pkg/router/template"
+	"strings"
+	"testing"
+)
+
+func TestPercentDecodingForHeaders(t *testing.T) {
```
candita, line 11:
```suggestion
func TestParseHeadersToBeSetOrDeleted(t *testing.T) {
```

pkg/cmd/infra/router/template_test.go
```diff
@@ -0,0 +1,111 @@
...
+			expectError:        true,
+			expectErrorMessage: "invalid HTTP header input specification: ",
+		},
+		{
+			description:        "should not allow blank values for headers",
+			inputValue:         "",
+			expectedValue:      nil,
+			expectError:        true,
+			expectErrorMessage: "encoded header string not present",
+		},
```
candita, line 94: Add a test for an Action that is not allowed - like something misspelled, or even `Append`.

pkg/router/router_test.go
```diff
@@ -445,7 +559,12 @@ func TestConfigTemplate(t *testing.T) {
 		for _, expectation := range expectations {
 			t.Run(name, func(t *testing.T) {
 				if err := expectation.Match(parser); err != nil {
-					t.Fatalf(err.Error())
+					if content, err := ioutil.ReadFile(config); err != nil {
+						t.Error(err)
+					} else {
+						t.Log("haproxy.config:", string(content))
+					}
```
candita, line 566: We probably don't need to log the content when there is no error.

pkg/router/template/router.go
candita: Change `set/delete` to `set or delete`.

pkg/cmd/infra/router/template.go
miheer: yes created functions now

pkg/cmd/infra/router/template.go
miheer: Yes. I created functions.

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,83 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+					}
+					sanitizedHeaderValue := templateplugin.SanitizeHeaderValue(headerValue)
+					action, err := url.QueryUnescape(parts[2])
+					if err != nil {
+						return captureHeaders, fmt.Errorf("failed to decode percent encoding: %v", parts[2])
+					}
+					capture = templateplugin.HTTPHeader{
+						Name:   headerName,
+						Value:  sanitizedHeaderValue,
+						Action: routev1.RouteHTTPHeaderActionType(action),
```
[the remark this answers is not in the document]
miheer: i have written a function checkValidAction

pkg/cmd/infra/router/template_test.go
miheer: ok

miheer:
@candita changes in 1st iteration https://github.com/openshift/router/compare/82dc5cae9f2f38f59527c27381bda3f6dfee5bd1..c7d5bff14aa1db240e124f18e4a995b3d821c540
changes in 2nd iteration
https://github.com/openshift/router/compare/c7d5bff14aa1db240e124f18e4a995b3d821c540..4ff95586e1bc199bfb67b07e55822d6aa9c72c15

pkg/cmd/infra/router/template_test.go
miheer: done

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,83 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+	if len(in) > 0 {
+		for _, header := range strings.Split(in, ",") {
+			parts := strings.Split(header, ":")
+			num := len(parts)
+			switch num {
+			default:
+				return captureHeaders, fmt.Errorf("invalid HTTP header input specification: %v", header)
+			case 3:
+				{
+					err = checkValidHeaderName(parts[0])
```
candita, line 312: You should be running `url.QueryUnescape` on every `part[*]` before you validate it.  Here, you are only running it on `parts[1]`.

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,83 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+				}
+			}
+		}
+	}
+
+	return captureHeaders, err
+}
+
+func checkValidAction(action string) error {
+	if action != "Set" && action != "Delete" {
```
candita, line 356: Don't use strings here, use the types, something like this:
```suggestion
	if action != string(routev1.Set) && action != string(routev1.Delete) {
```

pkg/cmd/infra/router/template_test.go
```diff
@@ -0,0 +1,132 @@
...
+		inputValue         string
+		expectedValue      []templateplugin.HTTPHeader
+		expectError        bool
+		expectErrorMessage string
+	}{
+		{
+			description: "should percent decode the single header value",
+			inputValue:  "Accept:text%2Fplain%2C+text%2Fhtml:Set",
+			expectedValue: []templateplugin.HTTPHeader{
+				{Name: "Accept", Value: "'text/plain, text/html'", Action: "Set"},
```
candita, line 23:
```suggestion
				{Name: "Accept", Value: "'text/plain, text/html'", Action: string (routev1.Set)},
```

candita: Please make these changes throughout.

pkg/cmd/infra/router/template_test.go
```diff
@@ -0,0 +1,132 @@
...
+		{
+			description: "should percent decode the multiple header values with simple non encoded strings",
+			inputValue:  "X-Frame-Options:DENY:Set",
+			expectedValue: []templateplugin.HTTPHeader{
+				{Name: "X-Frame-Options", Value: "'DENY'", Action: "Set"},
+			},
+			expectError: false,
+		},
+		{
+			description:        "When Action is `Set` it should not percent decode the incorrectly encoded value part Eg: percent is not followed by two hexadecimal digits",
```
candita, line 62:
```suggestion
			description:        "when Action is `Set` it should report an error on incorrectly encoded value part (percent is not followed by two hexadecimal digits)",
```

candita: Lower case first letter on the test case descriptions, everywhere, please.

pkg/cmd/infra/router/template_test.go
```diff
@@ -0,0 +1,132 @@
...
+			expectedValue:      nil,
+			expectError:        true,
+			expectErrorMessage: "encoded header string not present",
+		},
+		{
+			description:        "Should fail with incorrect header name when HTTP header input specification has 3 parts",
+			inputValue:         "X-Frame-Options**!=:DENY:Set",
+			expectedValue:      nil,
+			expectError:        true,
+			expectErrorMessage: "invalid HTTP header name: X-Frame-Options**!",
```
candita, line 100: Why doesn't this header name have the `=` at the end like the inputValue?

pkg/cmd/infra/router/template_test.go
```diff
@@ -0,0 +1,132 @@
...
+			expectErrorMessage: "invalid HTTP header input specification: X-Frame-Options:DENY:Set:Invalid",
+		},
+	}
+
+	for _, tc := range testCases {
+		switch gotValue, err := parseHeadersToBeSetOrDeleted(tc.inputValue); {
+		case !cmp.Equal(gotValue, tc.expectedValue):
+			t.Errorf(" expected %s, got %s", tc.expectedValue, gotValue)
+		case err == nil && tc.expectError:
+			t.Errorf("%s: expected error, got nil", tc.description)
```
candita, line 123:
```suggestion
			t.Errorf("%s: expected error %q, got nil", tc.description, tc.expectErrorMessage)
```

pkg/cmd/infra/router/template_test.go
```diff
@@ -0,0 +1,132 @@
...
+		case err != nil && !tc.expectError:
+			t.Errorf("%s: expected success, got error: %v", tc.description, err)
+		case err != nil && len(tc.expectErrorMessage) != 0:
+			if !strings.Contains(fmt.Sprint(err), tc.expectErrorMessage) {
+				t.Errorf("the error message %s returned does not match expected return message %s", err, tc.expectErrorMessage)
+			}
+		}
```
candita, line 130: Do this like https://github.com/openshift/library-go/pull/1449/files#diff-51cbcb9e560e512d7dba0cd93dab386055814a27e2c7903ad93bce92c58fd18bR1680, something more like:
```suggestion
                var actualErrorMessage string
                if err != nil {
                    actualErrorMessage = err.Error()
                }
		switch {
			case tc.expectedErrorMessage == "" && actualErrorMessage != "":
				t.Fatalf("unexpected error: %v", actualErrorMessage)
			case tc.expectedErrorMessage != "" && actualErrorMessage == "":
				t.Fatalf("got nil, expected %v", tc.expectedErrorMessage)
			case !strings.Contains(actualErroMessage, tc.expectedErrorMessage):
				t.Fatalf("unexpected error: %v, expected: %v", actualErrorMessage, tc.expectedErrorMessage)
			}
```

pkg/cmd/infra/router/template_test.go
```diff
@@ -0,0 +1,132 @@
...
+			expectedValue:      nil,
+			expectError:        true,
+			expectErrorMessage: "invalid HTTP header input specification: X-Frame-Options:DENY:Set:Invalid",
+		},
+	}
+
+	for _, tc := range testCases {
+		switch gotValue, err := parseHeadersToBeSetOrDeleted(tc.inputValue); {
+		case !cmp.Equal(gotValue, tc.expectedValue):
+			t.Errorf(" expected %s, got %s", tc.expectedValue, gotValue)
```
candita, line 121:
```suggestion
		gotValue, err := parseHeadersToBeSetOrDeleted(tc.inputValue); 
```

pkg/cmd/infra/router/template_test.go
```diff
@@ -0,0 +1,132 @@
+package router
+
+import (
+	"fmt"
+	"github.com/google/go-cmp/cmp"
+	templateplugin "github.com/openshift/router/pkg/router/template"
```
candita, line 6: Please locate these imports correctly.

pkg/cmd/infra/router/template_test.go
candita: I recommended some changes in your loop over the testCases.  https://github.com/openshift/router/pull/438/files#diff-a391430739b2cebfb45c7a0eb8c8b80910ba610cf34d7e9d80b08d005a368124R120 in particular, should not be the first case evaluated or you won't be able to compare the expected error message.

candita: I don't think your error messages are being checked.

pkg/router/router_test.go
```diff
@@ -408,6 +418,110 @@ func TestConfigTemplate(t *testing.T) {
 				},
 			},
 		},
+		"Simple global HTTP request header": {
+			mustCreateWithConfig{
```
candita, line 422: Are these run?  Why don't they have the `mustCreate` stanza like the other tests?

pkg/router/router_test.go
candita: Is this for debugging?  Please remove.

pkg/router/template/router.go
```diff
@@ -1393,6 +1447,8 @@ func configsAreEqual(config1, config2 *ServiceAliasConfig) bool {
 		config1.RoutingKeyName == config2.RoutingKeyName &&
 		config1.IsWildcard == config2.IsWildcard &&
 		config1.VerifyServiceHostname == config2.VerifyServiceHostname &&
+		reflect.DeepEqual(config1.HTTPResponseHeaders, config2.HTTPResponseHeaders) &&
+		reflect.DeepEqual(config1.HTTPRequestHeaders, config2.HTTPRequestHeaders) &&
```
candita, line 1451: This is something that should probably be unit tested.  Is it possible?

pkg/router/template/router_test.go
```diff
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/sets"
 
+	"github.com/google/go-cmp/cmp"
```
candita, line 16: nit - I wish this had openshift/api before any k8s.io imports.

pkg/router/router_test.go
```diff
@@ -456,16 +575,30 @@ type expectation interface {
 	Apply(h *harness) error
 }
 
+// mustCreate represents a route that gets created in a unit test.
```
candita, line 578: Thanks for adding comments on these!

pkg/router/template/router_test.go
miheer: I am not sure what exactly you want here

miheer: can please specify in detail ?

pkg/router/template/router.go
miheer: I am not sure how we could test it. Just passing two equal and unequal  HTTPResponseHeaders and 
HTTPRequestHeaders  ServiceAliasConfig object ?

pkg/router/router_test.go
miheer: This will be set by ingress controller so setting field in route is not required

pkg/cmd/infra/router/template_test.go
miheer: I am not sure what you want on this. We need switch.

pkg/cmd/infra/router/template_test.go
miheer: ack made the changes

pkg/cmd/infra/router/template_test.go
miheer: yes

pkg/cmd/infra/router/template_test.go
miheer: ok

pkg/cmd/infra/router/template.go
miheer: done

pkg/cmd/infra/router/template_test.go
miheer: done

pkg/cmd/infra/router/template_test.go
miheer: done

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,99 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+			}
+		}
+	}
+
+	return captureHeaders, err
+}
+
+func checkValidAction(action string) error {
+	if action != string(routev1.Set) && action != string(routev1.Delete) {
+		return fmt.Errorf("invalid action %v", action)
```
candita, line 373: nit:
```suggestion
		return fmt.Errorf("invalid action %s", action)
```

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,99 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+	} else {
+		return nil
+	}
+}
+
+func checkValidHeaderName(headerName string) error {
+	// RFC 2616, section 4.2, states that the header name
+	// must be a valid token.
+	if !validTokenRE.MatchString(headerName) {
+		return fmt.Errorf("invalid HTTP header name: %v", headerName)
```
candita, line 383: nit:
```suggestion
		return fmt.Errorf("invalid HTTP header name: %s", headerName)
```

pkg/cmd/infra/router/template_test.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			switch actualValue, actualErrorMessage := parseHeadersToBeSetOrDeleted(tc.inputValue); {
+			case !cmp.Equal(actualValue, tc.expectedValue):
+				t.Errorf(" expected %s, got %s", tc.expectedValue, actualValue)
+			case tc.expectErrorMessage == "" && actualErrorMessage != nil && actualErrorMessage.Error() != "":
+				t.Fatalf("unexpected error: %v", actualErrorMessage)
+			case tc.expectErrorMessage != "" && actualErrorMessage != nil && actualErrorMessage.Error() == "":
```
candita, line 128: Here you should be looking for the case where actualErrorMessage is nil or blank:
```suggestion
			case tc.expectErrorMessage != "" && (actualErrorMessage == nil || actualErrorMessage.Error() == ""):
```

pkg/router/template/router.go
candita: I think there's going to be a problem here if the headers get out of order.  Some evidence that it works would be helpful.

pkg/router/template/router_test.go
candita: I wish L16+17 were above L12.  Not a big deal.

pkg/cmd/infra/router/template_test.go
miheer: actualErrorMessage == nil

then nil.Error will crash

pkg/cmd/infra/router/template.go
miheer: ok

pkg/cmd/infra/router/template.go
miheer: done

pkg/router/template/router_test.go
miheer: ok done

pkg/router/router_test.go
Miciah: I asked Miheer to add this.  It is only printed when the test fails, and it is helpful for diagnosing the cause of the failure (at least it was useful to me).

pkg/router/template/router.go
Miciah: @candita, is the unit test in https://github.com/miheer/router/pull/2 satisfactory?

pkg/router/router_test.go
miheer: I also find it useful

pkg/cmd/infra/router/template_test.go
miheer: ok done

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,99 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+					}
+					err = checkValidHeaderName(headerName)
+					if err != nil {
+						return captureHeaders, err
+					}
+					headerValue, err := url.QueryUnescape(parts[1])
+					if err != nil {
+						return captureHeaders, fmt.Errorf("failed to decode percent encoding: %v", parts[1])
+					}
+					sanitizedHeaderValue := templateplugin.SanitizeHeaderValue(headerValue)
```
candita, line 324: Why does the header value have to be sanitized?  If it has a valid reason, why isn't the header name or action sanitized?

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,99 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+		return fmt.Errorf("invalid action %s", action)
+	} else {
+		return nil
+	}
+}
+
+func checkValidHeaderName(headerName string) error {
+	// RFC 2616, section 4.2, states that the header name
+	// must be a valid token.
+	if !validTokenRE.MatchString(headerName) {
```
candita, line 382: The validTokenRE is the same now as the regex for header name in https://github.com/openshift/library-go/pull/1449/files#diff-bfc6ba95cbc94986901ce1d0e3b74ee966499eab85a96e1d1ff763d2d74a7eeeR64

How are you going to make sure changes between these get propagated?  I think you probably need to choose one or the other.

pkg/cmd/infra/router/template_test.go
candita: Not done yet.

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,99 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+			}
+		}
+	}
+
+	return captureHeaders, err
+}
+
+func checkValidAction(action string) error {
+	if action != string(routev1.Set) && action != string(routev1.Delete) {
+		return fmt.Errorf("invalid action %s", action)
```
candita, line 373: Add a semi-colon between the error message and the error value, as customary.  Unit tests will need to be updated too.
```suggestion
		return fmt.Errorf("invalid action: %s", action)
```

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,99 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
 	return captureHeaders, nil
 }
 
+func parseHeadersToBeSetOrDeleted(in string) ([]templateplugin.HTTPHeader, error) {
+	var captureHeaders []templateplugin.HTTPHeader
+	var capture templateplugin.HTTPHeader
+	var err error
+	if len(in) == 0 {
+		return captureHeaders, fmt.Errorf("encoded header string not present.")
```
candita, line 301: This is the only error message with a period at the end.  Please remove.
```suggestion
		return captureHeaders, fmt.Errorf("encoded header string not present")
```

images/router/haproxy/conf/haproxy-config.template
miheer: https://github.com/openshift/router/pull/438/files#diff-21bd64a4ad8a7927e9d59e99e09628efc884fe6d6291e50eb33c5deb3d2a79f0L684

is the end of the block which allows the non-edge or non-re-encrypt, edge and re-encrypt route except passthrough.

pkg/cmd/infra/router/template.go
miheer: Header name goes through a regex check so a user can't provide an invalid value.
Header action also only 2 values are allowed via api so this check happens at api level.

pkg/cmd/infra/router/template.go
miheer: This point is valid. Both check the same chars. However I can add a regex check here to maintain consistency.

miheer: I will put a comment here.

images/router/haproxy/conf/haproxy-config.template
candita: Ok, I see lines 662 and 669 are checking for edge, reencrypt, whereas line 528 checks for none, edge, reencrypt.

pkg/cmd/infra/router/template.go
candita: The SanitizeHeaderValue is escaping single quotes, which we allow in the header name per the regex.  Why don't we allow single quotes in the header value?

ShudiLi:
/label qe-approved
If needed, I can list the test cases what I have done, thanks.

miheer:
> /label qe-approved If needed, I can list the test cases what I have done, thanks.

OK. If you can add them in a google doc and share us the link then that will be great.

pkg/cmd/infra/router/template.go
```diff
@@ -287,6 +293,102 @@ func parseCaptureHeaders(in string) ([]templateplugin.CaptureHTTPHeader, error)
...
+		return nil
+	}
+}
+
+func checkValidHeaderName(headerName string) error {
+	// RFC 2616, section 4.2, states that the header name
+	// must be a valid token.
+	// Any changes made to regex of header name in route type in openshift/api and `validation.go` file in library-go shall be reflected here and
+	// vice-versa
+	permittedHeaderNameRE := regexp.MustCompile("^[-!#$%&'*+.0-9A-Z^_`a-z|~]+$")
```
Miciah, line 384: Don't recompile the regexp every time the function is called.  Don't use the word "shall".  End your sentences with a period.  Remember to add godoc. 
```suggestion
// permittedHeaderNameRE is a compiled regexp to match an HTTP header name
// as specified in RFC 2616, section 4.2.
// Any changes made to regex of header name in route type in openshift/api and `validation.go` file in library-go must be reflected here and
// vice versa.
var permittedHeaderNameRE = regexp.MustCompile("^[-!#$%&'*+.0-9A-Z^_`a-z|~]+$")

// checkValidHeaderName verifies that the given HTTP header name is valid.
func checkValidHeaderName(headerName string) error {
```

pkg/cmd/infra/router/template.go
Miciah: `SanitizeHeaderValue` doesn't actually sanitize, but rather it quotes the string (for this reason, I've suggested changing "sanitize" to "quote" a few times).  I agree, it certainly _seems_ like the header name can container `'` and therefore needs to be quoted in the configuration file, just as the value needs to be quoted.

PR #439, opened 2023-01-04
OCPBUGS-5073: Count endpoints per port for weights and enabling active health checks

gcs278:
Previously, you could have a service with two different ports and openshift-router would count both, regardless of what `route.spec.port.targetPort` was configured to. This change uses the route's targetPort when counting endpoints. This 
impacts weights for alternateBackends as well a bug in which a service with two ports could enable backend health checks, despite only having one server.

- `pkg/router/template/router.go`: Added the port parameter to getActiveEndpoints, calculateServiceWeights, and numberOfEndpoints.
- `pkg/router/template/router_test.go`: Updated unit testing for calculateServiceWeights to provide port.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -538,10 +538,10 @@ func (r *templateRouter) writeConfig() error {
 
 		// calculate the server weight for the endpoints in each service
 		// called here to make sure we have the actual number of endpoints.
-		cfg.ServiceUnitNames = r.calculateServiceWeights(cfg.ServiceUnits)
+		cfg.ServiceUnitNames = r.calculateServiceWeights(cfg.ServiceUnits, cfg.PreferPort)
 
 		// Calculate the number of active endpoints for the route.
-		cfg.ActiveEndpoints = r.getActiveEndpoints(cfg.ServiceUnits)
+		cfg.ActiveEndpoints = r.getActiveEndpoints(cfg.ServiceUnits, cfg.PreferPort)
 
 		cfg.Status = ServiceAliasConfigStatusSaved
 		r.state[k] = cfg
@@ -757,7 +757,7 @@ func (r *templateRouter) dynamicallyAddRoute(backendKey ServiceAliasConfigKey, r
 	oldEndpoints := []Endpoint{}
 
 	// As the endpoints have changed, recalculate the weights.
-	newWeights := r.calculateServiceWeights(backend.ServiceUnits)
+	newWeights := r.calculateServiceWeights(backend.ServiceUnits, backend.PreferPort)
 	for key := range backend.ServiceUnits {
 		if service, ok := r.findMatchingServiceUnit(key); ok {
 			newEndpoints := endpointsForAlias(*backend, service)
@@ -819,7 +819,7 @@ func (r *templateRouter) dynamicallyReplaceEndpoints(id ServiceUnitKey, service 
 		newEndpoints := endpointsForAlias(cfg, service)
 
 		// As the endpoints have changed, recalculate the weights.
-		newWeights := r.calculateServiceWeights(cfg.ServiceUnits)
+		newWeights := r.calculateServiceWeights(cfg.ServiceUnits, cfg.PreferPort)
 
 		// Get the weight for this service unit.
 		weight, ok := newWeights[id]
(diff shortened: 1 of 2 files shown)


gcs278:
/hold
I need to test more and have discussion on potential edge cases this could impact.

pkg/router/template/router.go
```diff
@@ -1068,11 +1068,15 @@ func (r *templateRouter) removeRouteInternal(route *routev1.Route) {
...
+func (r *templateRouter) numberOfEndpoints(id ServiceUnitKey, port string) int32 {
 	var eps = 0
 	svc, ok := r.findMatchingServiceUnit(id)
 	if ok && len(svc.EndpointTable) > eps {
-		eps = len(svc.EndpointTable)
+		for _, ep := range svc.EndpointTable {
+			if ep.Port == port {
+				eps += 1
+			}
+		}
```
Miciah, line 1079: Does this do the right thing when the route's `spec.port` is omitted (in which case the router uses all ports)?  `endpointsForAlias` checks for empty `PreferPort`, and I expect you need to do likewise here: https://github.com/openshift/router/blob/80242039851150ba43a7fdca541d3e03ddc3ba76/pkg/router/template/template_helper.go#L180-L192

gcs278: Thanks, I knew I was missing something, fixed and added more unit tests.

gcs278:
Not related
/retest

pkg/router/template/router_test.go
```diff
@@ -972,14 +990,104 @@ func TestCalculateServiceWeights(t *testing.T) {
 				suKey1: 1,
 			},
 		},
+		{
+			name:      "a single service with multiple endpoints, but no endpoint ports match the route port",
+			routePort: "9090",
```
frobware, line 995: I changed `routePort` to "port" but the test failed. Shouldn't the symbolic form also pass?

pkg/router/template/router_test.go
```diff
@@ -991,20 +1099,22 @@ func TestCalculateServiceWeights(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			for suKey, eps := range tc.serviceUnits {
+				router.CreateServiceUnit(suKey)
+				router.AddEndpoints(suKey, eps)
+			}
+			endpointWeights := router.calculateServiceWeights(tc.serviceWeights, tc.routePort)
+			if !reflect.DeepEqual(endpointWeights, tc.expectedWeights) {
+				t.Errorf("test %s: expected endpointWeights to be %v, got %v", tc.name, tc.expectedWeights, endpointWeights)
+			}
+			// Remove endpoints and service units so the same sample template router
```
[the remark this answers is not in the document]
gcs278: Ah yes I agree, good time to fix this. Done.

pkg/router/template/router_test.go
gcs278: In this specific case, it will fail if you change it to "port" because the test matches "8080" or "port" and this test case is testing *not* matching either of those, hence why it's currently "9090". So your changing it to actually match, when we are testing for it to not match.

frobware: OK, thanks. I went through and fiddled with the inputs getting failures where I understood them and non-failures when substituting <port-number> for "port" -- except this specific case which is why I was seeking some clarification. All good.

gcs278:
/unhold
I reviewed again, trying to look for edge cases, and can't think of any situation this would break.

pkg/router/template/router_test.go
```diff
@@ -991,20 +1097,18 @@ func TestCalculateServiceWeights(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			router := NewFakeTemplateRouter()
+
+			for suKey, eps := range tc.serviceUnits {
+				router.CreateServiceUnit(suKey)
+				router.AddEndpoints(suKey, eps)
+			}
+			endpointWeights := router.calculateServiceWeights(tc.serviceWeights, tc.routePort)
+			if !reflect.DeepEqual(endpointWeights, tc.expectedWeights) {
+				t.Errorf("test %s: expected endpointWeights to be %v, got %v", tc.name, tc.expectedWeights, endpointWeights)
```
Miciah, line 1109: You don't need to include the test name in the error message; using `t.Run` gets you that for free.

gcs278: Done

Miciah:
Thanks!
/approve

PR #447, opened 2023-02-07
OCPBUGS-6661: Manage mTLS CRLs

rfredette:
This PR makes the router parse the certificates supplied in `ROUTER_MUTUAL_TLS_AUTH_CA`, download CRLs from any CRL distribution points that are found, and write them to `/crls/crls.pem`. That file is supplied to haproxy to use for mTLS client auth.

This PR requires that the router deployment include an `EmptyDir` mount for the `/crls` directory

### DIFF
--- a/images/router/haproxy/Dockerfile.rhel8
+++ b/images/router/haproxy/Dockerfile.rhel8
@@ -4,7 +4,7 @@ RUN INSTALL_PKGS="haproxy22 rsyslog procps-ng util-linux" && \
     rpm -V $INSTALL_PKGS && \
     yum clean all && \
     mkdir -p /var/lib/haproxy/router/{certs,cacerts,whitelists} && \
-    mkdir -p /var/lib/haproxy/{conf/.tmp,run,bin,log} && \
+    mkdir -p /var/lib/haproxy/{conf/.tmp,run,bin,log,mtls} && \
     touch /var/lib/haproxy/conf/{{os_http_be,os_edge_reencrypt_be,os_tcp_be,os_sni_passthrough,os_route_http_redirect,cert_config,os_wildcard_domain}.map,haproxy.config} && \
     setcap 'cap_net_bind_service=ep' /usr/sbin/haproxy && \
     chown -R :0 /var/lib/haproxy && \
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -9,6 +9,8 @@
 {{- $dynamicConfigManager := .DynamicConfigManager }}
 {{- $router_ip_v4_v6_mode := env "ROUTER_IP_V4_V6_MODE" "v4" }}
 {{- $router_disable_http2 := env "ROUTER_DISABLE_HTTP2" "false" }}
+{{- $haveClientCA := .HaveClientCA }}
+{{- $haveCRLs := .HaveCRLs }}
 
 
 {{- /* A bunch of regular expressions.  Each should be wrapped in (?:) so that it is safe to include bare */}}
@@ -290,8 +292,8 @@ frontend fe_sni
     {{- "" }} crt-list /var/lib/haproxy/conf/cert_config.map accept-proxy
     {{- with (env "ROUTER_MUTUAL_TLS_AUTH") }}
       {{- "" }} verify {{. }}
-    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CA") }} ca-file {{. }} {{ else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
-    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CRL") }} crl-file {{. }} {{ end }}
+	{{- if $haveClientCA }} ca-file /var/lib/haproxy/mtls/latest/ca-bundle.pem {{ else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
+	{{- if $haveCRLs }} crl-file /var/lib/haproxy/mtls/latest/crls.pem {{ end }}
     {{- end }}
   mode http
 
@@ -376,8 +378,8 @@ frontend fe_no_sni
   bind unix@/var/lib/haproxy/run/haproxy-no-sni.sock ssl crt {{ firstMatch ".+" .DefaultCertificate "/var/lib/haproxy/conf/default_pub_keys.pem" }} accept-proxy
     {{- with (env "ROUTER_MUTUAL_TLS_AUTH") }}
       {{- "" }} verify {{. }}
-    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CA") }} ca-file {{. }} {{ else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
-    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CRL") }} crl-file {{. }} {{ end }}
+	{{- if $haveClientCA }} ca-file /var/lib/haproxy/mtls/latest/ca-bundle.pem {{ else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
+	{{- if $haveCRLs }} crl-file /var/lib/haproxy/mtls/latest/crls.pem {{ end }}
     {{- end }}
   mode http
 
--- a/pkg/router/crl/crl.go
+++ b/pkg/router/crl/crl.go
@@ -1,0 +1,529 @@
+package crl
+
+import (
+	"bytes"
+	"crypto/x509"
+	"encoding/asn1"
+	"encoding/hex"
+	"encoding/pem"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	logf "github.com/openshift/router/log"
+	"github.com/openshift/router/pkg/util"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+var log = logf.Logger.WithName("crl")
+
+const (
+	// crlFilePermissions is the permission bits used for the CRL file.
+	crlFilePermissions = 0644
+	// stagingDirPermissions is the permission bits used for staging directories
+	stagingDirPermissions = 0755
+	// crlFallbackTime is how long to wait before retrying if nextUpdate can't be determined.
+	crlFallbackTime = 5 * time.Minute
+	// errorBackoffTime is how long to wait before retrying if a generic error happens during CRL refresh.
+	errorBackoffTime = 5 * time.Minute
+	// mtlsBaseDirectory is the directory where all crl temp directories and symlinks will live.
+	mtlsBaseDirectory = "/var/lib/haproxy/mtls"
+	// crlBasename is the name of the crl file
+	crlBasename = "crls.pem"
+	// caBundleBasename is the name of the CA bundle file copied from the CA bundle configmap.
+	caBundleBasename = "ca-bundle.pem"
+	// dummyCRL is a placeholder CRL so that HAProxy can start serving non-mTLS traffic while CRLs are downloaded. This
+	// CRL is for a CA cert/key that are intentionally not included, and was generated with an expiration (nextUpdate)
+	// at 1:00AM GMT on Jan 1, 2000 so that in the extremely unlikely case that someone is able to generate a matching
+	// cert and key, HAProxy will still reject the connection due to the expired CRL.
+	dummyCRL = `-----BEGIN X509 CRL-----
+MIIBzzCBuAIBATANBgkqhkiG9w0BAQsFADBhMQswCQYDVQQGEwJVUzELMAkGA1UE
+CAwCTkMxEDAOBgNVBAcMB1JhbGVpZ2gxDDAKBgNVBAoMA09TNDEMMAoGA1UECwwD
+RW5nMRcwFQYDVQQDDA5QbGFjZWhvbGRlciBDQRcNMDAwMTAxMDAwMDAwWhcNMDAw
+MTAxMDEwMDAwWqAjMCEwHwYDVR0jBBgwFoAUbaqyU8VAswFgefsu6pOdvqK1nfgw
+DQYJKoZIhvcNAQELBQADggEBAD8W+OmWHp9Pg7914rA5QOk+pUCZ4F7++fbmGPpc
+9gdNOxkeCrZ3sdBeEs0P3+tSf8dLcpI5PKEbL+bC3wrIM3yzsD+mIZkvV/FGhgE1
+s7b6IA/8FYsmNWIjgAWBAp13zh0AH3qhpI01tm+cQETz6r249TWQ+p04pEA89+XT
+7CE99nHd8yNDOESs1xZreSFkIF/Hmm8y4I0o/+8wpjA9e3PJ7O25ZB2OGX4FufMf
+tVa0xfWd9czWFqM1DjU3ME0mVi6lr38AhUDoG6sFbHk+TfzTp4ykVUpXIHu4bJTG
+DPfV3SE277EvsrsGFYIsxWgXskITjzb9no9fnodd/jG46tw=
+-----END X509 CRL-----
+`
+)
+
+var (
+	// mtlsLatestSymlink is the fully qualified path to the symlink used in the template.
+	mtlsLatestSymlink = filepath.Join(mtlsBaseDirectory, "latest")
+	// mtlsNextSymlink is the fully qualified path to the staging symlink, used to atomically replace the existing
+	// mtlsLatestSymlink.
+	mtlsNextSymlink = filepath.Join(mtlsBaseDirectory, "next")
+	// CRLFilename is the fully qualified path to the currently in use crl file.
+	CRLFilename = filepath.Join(mtlsLatestSymlink, crlBasename)
+	// CABundleFilename is the fully qualified path to the currently in use CA bundle.
+	CABundleFilename = filepath.Join(mtlsLatestSymlink, caBundleBasename)
+)
+
+// authorityKeyIdentifier is a certificate's authority key identifier.
+type authorityKeyIdentifier struct {
+	KeyIdentifier []byte `asn1:"optional,tag:0"`
+}
+
+// authorityKeyIdentifierOID is the ASN.1 object identifier for the authority key identifier extension.
+var authorityKeyIdentifierOID = asn1.ObjectIdentifier{2, 5, 29, 35}
+
+// InitMTLSDirectory creates an initial directory for HAProxy to use to complete startup and serve non-mTLS traffic
+// while CRLs are being downloaded in the background. Returns an error if any of the filesystem operations fail.
+func InitMTLSDirectory(caBundleFilename string) error {
+	stagingDirectory, err := makeStagingDirectory()
+	if err != nil {
+		return err
+	}
+	// Copy the CA bundle from the configmap mount location to the staging directory
+	if err := util.CopyFile(caBundleFilename, filepath.Join(stagingDirectory, caBundleBasename)); err != nil {
+		return err
+	}
+	// Write out the dummyCRL as a placeholder. With this, HAProxy will reject connections that require CRLs until after
+	// all CRL downloads are complete, but other traffic should be handled correctly.
+	if err := os.WriteFile(filepath.Join(stagingDirectory, crlBasename), []byte(dummyCRL), crlFilePermissions); err != nil {
+		return err
+	}
+	// At any time other than startup, we need to be careful overwrite the previous mtlsLatestSymlink in an atomic way,
+	// but at initialization, mtlsLatestSymlink shouldn't exist yet, so it should be safe to just directly create the
+	// symlink.
+	if err := os.Symlink(stagingDirectory, mtlsLatestSymlink); err != nil {
+		return err
+	}
+	return nil
+}
+
+// CABundleHasCRLs returns true if any of the certificates in caBundleFilename specify a CRL distribution point.
+// Returns an error if the CA Bundle could not be parsed.
+func CABundleHasCRLs(caBundleFilename string) (bool, error) {
+	clientCAData, err := os.ReadFile(caBundleFilename)
+	if err != nil {
+		return false, err
+	}
+	for len(clientCAData) > 0 {
+		block, data := pem.Decode(clientCAData)
+		if block == nil {
+			break
+		}
+		clientCAData = data
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			return false, fmt.Errorf("client CA bundle has an invalid certificate: %w", err)
+		}
+		if len(cert.CRLDistributionPoints) != 0 {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// ManageCRLs spins off a goroutine that ensures that any CRLs specified in caBundleFilename are downloaded and kept
+// up-to-date. It will automatically refresh expired CRLs and download missing CRLs when it receives a message on
+// caUpdateChannel (indicating the CA bundle has been updated), or when any existing CRL expires. Whenever either the CA
+// bundle or the CRL file has changed, updateCallback is called, with a boolean indicating whether crl-file needs to be
+// specified in the HAProxy config.
+func ManageCRLs(caBundleFilename string, caUpdateChannel <-chan struct{}, updateCallback func(bool)) {
+	go func() {
+		caUpdated := true
+		nextUpdate := time.Now()
+		updated := false
+		shouldHaveCRLs := false
+		for {
+			// All of the functions that could return an error also return data that needs to persist between iterations
+			// of the for loop, so we need to use var instead of := to declare err.
+			var err error
+			if caUpdated {
+				shouldHaveCRLs, err = CABundleHasCRLs(caBundleFilename)
+				if err != nil {
+					log.Error(err, "failed to parse CA bundle", "CA bundle filename", caBundleFilename)
+					// TODO: Should we back off if we can't parse the CA bundle, or just wait until the CA bundle is updated again?
+					nextUpdate = time.Now().Add(errorBackoffTime)
+					continue
+				}
+			}
+
+			nextUpdate, updated, err = updateCRLFile(caBundleFilename, caUpdated)
+			if err != nil {
+				log.Error(err, "failed to update CRLs")
+				nextUpdate = time.Now().Add(errorBackoffTime)
+				continue
+			}
+			// After successfully updating the CRL file, reset caUpdated
+			caUpdated = false
+			if updated {
+				updateCallback(shouldHaveCRLs)
+				updated = false
+			}
+
+			if nextUpdate.IsZero() {
+				log.V(4).Info("no nextUpdate. only watching for CA updates")
+				select {
+				case <-caUpdateChannel:
+					caUpdated = true
+				}
+			} else {
+				log.V(4).Info("nextUpdate is at " + nextUpdate.Format(time.RFC3339))
+				select {
+				case <-time.After(time.Until(nextUpdate)):
+				case <-caUpdateChannel:
+					caUpdated = true
+				}
+			}
+		}
+	}()
+}
+
+// updateCRLFile creates a new staging directory, updates CRLs, and updates mtlsLatestSymlink to point to the new
+// staging directory. Returns the next update time and a boolean for if anything changed. Returns an error if there was
+// an issue during the update.
+func updateCRLFile(caBundleFilename string, caUpdated bool) (time.Time, bool, error) {
+	stagingDirectory, err := makeStagingDirectory()
+	if err != nil {
+		log.Error(err, "failed to create staging directory")
+		return time.Time{}, false, err
+	}
+
+	defer reapStaleDirectories()
+
+	stagingCRLFilename := filepath.Join(stagingDirectory, crlBasename)
+
+	nextUpdate, crlsUpdated, err := writeCRLFile(caBundleFilename, CRLFilename, stagingCRLFilename)
+	if err != nil {
+		log.Error(err, "failed to update CRLs")
+		return time.Time{}, false, err
+	}
+
+	if caUpdated || crlsUpdated {
+		if err := commitCACRLUpdate(stagingDirectory, caBundleFilename, crlsUpdated); err != nil {
+			log.Error(err, "failed to commit CRL update")
+			return time.Time{}, false, err
+		}
+		return nextUpdate, true, nil
+	}
+	return nextUpdate, false, nil
+}
+
+// commitCACRLUpdate makes sure stagingDirectory contains up-to-date versions of both the CA bundle and the CRL file,
+// then updates the symlink so that HAProxy will reference the new versions on reload. Returns an error if any of the
+// file operations fail.
+func commitCACRLUpdate(stagingDirectory, caBundleFilename string, stagingCRLUpdated bool) error {
+	// Copy CA bundle to the new directory.
+	if err := util.CopyFile(caBundleFilename, filepath.Join(stagingDirectory, caBundleBasename)); err != nil {
+		return err
+	}
+	// If stagingCRLUpdated is true, then the CRL file has already been written to the new directory, so nothing needs
+	// to be done for it. However, if stagingCRLUpdated is false, then we need to copy the existing crl file to the new
+	// directory.
+	if !stagingCRLUpdated {
+		if err := util.CopyFile(CRLFilename, filepath.Join(stagingDirectory, crlBasename)); err != nil {
+			if os.IsNotExist(err) {
+				// Even if CRLFilename doesn't currently exist, a crl file may need to be supplied to HAProxy when this
+				// update is committed. Using the dummyCRL allows http traffic to be handled as normal, but essentially
+				// guarantees that any mTLS connections will fail until new CRLs are downloaded.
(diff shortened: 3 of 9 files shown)


pkg/router/crl/crl.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+	}
+
+	if len(subjectKeyIds) == 0 {
+		if !crlFileExists {
+			// Make sure the crl file exists, even if it's empty
+			return nextCRLUpdate, os.WriteFile(crlFilename, []byte(""), crlFilePermissions)
+		} else {
+			return nextCRLUpdate, nil
+		}
```
Miciah, line 137: If we reach this point and `subjectKeyIds` is empty, that means that we parsed the CAs, and none of them had a CRL distribution point, right?  So why not simply truncate `crlFilename` when `subjectKeyIds` is empty?
```suggestion
		nextCRLUpdate = time.Time{}
		return nextCRLUpdate, os.Truncate(crlFilename, 0)
```
Then `ManageCRLs` should take a zero value to mean that no update is needed until the CA is updated.  Maybe it would be better to return a Boolean from `updateCRLFile` to indicate if no automatic update is required because there are no CRLs to update.

pkg/router/crl/crl.go
```diff
@@ -0,0 +1,238 @@
...
+				Type:  "X509 CRL",
+				Bytes: asn1Data,
+			}
+			if err := pem.Encode(buf, block); err != nil {
+				return nextCRLUpdate, fmt.Errorf("failed to encode PEM for CRL for certificate key %s: %w", subjectKeyId, err)
+			}
+		}
+		crlData := buf.String()
+
+		return nextCRLUpdate, os.WriteFile(crlFilename, []byte(crlData), crlFilePermissions)
```
Miciah, line 158:
```suggestion
		return nextCRLUpdate, os.WriteFile(crlFilename, crlData.Bytes(), crlFilePermissions)
```

pkg/router/crl/crl.go
```diff
@@ -0,0 +1,238 @@
...
+			log.Info("retrieving CRL distribution point", "distribution point", distributionPoint)
+			crl, err := getHTTPCRL(distributionPoint)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("error getting %q: %w", distributionPoint, err))
+				continue
+			}
+			if crl.TBSCertList.NextUpdate.Before(now) {
+				log.Info("crl expired. trying next distribution point", "nextUpdate", crl.TBSCertList.NextUpdate.Format(time.RFC3339))
+				continue
+			}
```
Miciah, line 185: New feature?  * grin *.  

If there isn't another distribution point to try, then this log message is misleading.  However, if all distribution points have expired data, then you should track this condition as an error.
```go
	for i, distributionPoint := range distributionPoints {
	// ...
```
```suggestion
			if crl.TBSCertList.NextUpdate.Before(now) {
				if i+1 < len(distributionPonts) {
					log.Info("distribution point has an expired CRL; trying next distribution point", "nextUpdate", crl.TBSCertList.NextUpdate.Format(time.RFC3339))
				} else {
					errs = append(errs, fmt.Errorf("all reachable distribution points publish expired CRLs"))
				}
				continue
			}
```
Alternatively, instead of returning an error here, the caller could treat a nil CRL value as an error condition.

pkg/router/crl/crl.go
```diff
@@ -0,0 +1,238 @@
...
+		} else if crl != nil {
+			crls[subjectKeyId] = crl
+			subjectKeyIds = append(subjectKeyIds, subjectKeyId)
+			log.Info("new certificate revocation list", "subject key identifier", subjectKeyId, "next update", crl.TBSCertList.NextUpdate.Format(time.RFC3339))
+			if (nextCRLUpdate.IsZero() || crl.TBSCertList.NextUpdate.Before(nextCRLUpdate)) && crl.TBSCertList.NextUpdate.After(now) {
+				nextCRLUpdate = crl.TBSCertList.NextUpdate
+			}
+			updated = true
+		}
+	}
```
Miciah, line 129: Consider refactoring lines 87 through 129 into a new function.

pkg/router/template/router.go
```diff
@@ -380,6 +381,41 @@ func (r *templateRouter) watchVolumeMountDir(path string, reloadFn func()) error
 	return nil
 }
 
+// watchVolumeMountDir adds a watcher on path, which should be a secret or
+// configmap volume mount, and calls reloadFn when a change is detected.
+func (r *templateRouter) watchDir(path string, reloadFn func()) error {
```
Miciah, line 386:
```suggestion
// watchDir adds a watcher on path, which should be a secret or
// configmap volume mount, and calls reloadFn when a change is detected.
func (r *templateRouter) watchDir(path string, reloadFn func()) error {
```

pkg/router/template/router.go
```diff
@@ -428,9 +464,19 @@ func (r *templateRouter) writeDefaultCert() error {
 // watchMutualTLSCert watches the directory containing the certificates for
 // mutual TLS and reloads the router if the directory contents change.
 func (r *templateRouter) watchMutualTLSCert() error {
+	var caUpdateChannel chan struct{}
 	caPath := os.Getenv("ROUTER_MUTUAL_TLS_AUTH_CA")
+	// TODO: Determine the best place to put CRLs
+	crlPath := "/crls/crls.pem"
```
Miciah, line 470: Any reason not to add a subdirectory under `/var/lib/haproxy/router/`?  (Note that you would need to add such a subdirectory to the `mkdir` command in `Dockerfile.rhel8`.)  

I guess that might cause must-gather to include CRLs, which can be large, but it's probably worth that tradeoff if it makes diagnosing CRL issues easier.

pkg/router/template/router.go
```diff
@@ -428,9 +464,19 @@ func (r *templateRouter) writeDefaultCert() error {
...
+	os.Setenv("ROUTER_MUTUAL_TLS_AUTH_CRL", crlPath)
```
Miciah, line 471: This looks a bit hacky.  Could you add a new field to `templateData` and use that instead?

pkg/router/template/router.go
```diff
@@ -428,9 +464,19 @@ func (r *templateRouter) writeDefaultCert() error {
...
 	if len(caPath) != 0 {
+		caUpdateChannel = make(chan struct{})
```
Miciah, line 473: You might as well declare `caUpdateChannel` inside the `if` block that uses it.
```suggestion
	caPath := os.Getenv("ROUTER_MUTUAL_TLS_AUTH_CA")
	// TODO: Determine the best place to put CRLs
	crlPath := "/crls/crls.pem"
	os.Setenv("ROUTER_MUTUAL_TLS_AUTH_CRL", crlPath)
	if len(caPath) != 0 {
		caUpdateChannel := make(chan struct{})
```

pkg/router/template/router.go
```diff
@@ -428,9 +464,19 @@ func (r *templateRouter) writeDefaultCert() error {
...
+		if err := crl.ManageCRLs(caPath, crlPath, caUpdateChannel); err != nil {
+			log.V(0).Info("failed to download CRLs", "error", err)
```
Miciah, line 475:
```suggestion
			log.V(0).Error(err, "failed to download CRLs")
```

pkg/router/template/router.go
Miciah: Correction: You don't need to `mkdir` the subdirectory if it comes from a volume mount.

pkg/router/crl/crl.go
```diff
@@ -0,0 +1,238 @@
...
+	nextUpdate, err := updateCRLFile(caBundleFilename, crlFilename)
+	if err != nil {
+		return err
+	}
+	go func() {
+		for true {
+			if nextUpdate.IsZero() {
+				// TODO: Make this a longer interval (for example, hourly) or implement backoff logic?
+				log.V(1).Info("Couldn't determine next CRL update time. Falling back to updating every 5 minutes.")
+				nextUpdate = time.Now().Add(5 * time.Minute)
```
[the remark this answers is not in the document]
rfredette: I currently have it set to 5 minutes to aid in testing, since I manually update the CRL files, and sometimes I've missed the actual expiration time. I will add a constant with the appropriate fallback interval before this is ready to merge.

pkg/router/template/router.go
rfredette: Which approach do you think is better? It sounds like creating the crl directory with `mkdir` is probably the more straightforward approach, plus it cuts down on the changes necessary to the ingress-operator since no `EmptyDir` volume mount would be necessary.

pkg/router/template/router.go
rfredette: Agreed, this is definitely hacky. With this change the actual CRL path is going to be constant, so I can probably change the template to include a `crl-file /crls/crls.pem` (or wherever it's decided the CRL file will live) whenever `ROUTER_MUTUAL_TLS_AUTH_CA` is set. What do you think of that approach?

pkg/router/crl/crl.go
rfredette: Yes, new feature :) It's not technically required to try other distribution points if one fails (assuming I understand RFC-5820 correctly), but I think this will help make this more robust. I probably should have called it out in the PR description, though.

pkg/router/crl/crl.go
rfredette: Correct; if `subjectKeyIds` is empty, none of the certs in the CA bundle specify any CRL distribution points. I wrote it this way to avoid extra writes if the CRL file already exists, so that the fsnotify watch won't be triggered unnecessarily. Do you know if `os.Truncate` will always overwrite the specified file, or if it will only overwrite it if it's longer than the specified length?

Makefile
```diff
@@ -31,10 +33,14 @@ build:
 	$(GO_BUILD_RECIPE)
 
 images/router/*/Dockerfile: images/router/base/Dockerfile
-	imagebuilder -t registry.svc.ci.openshift.org/openshift/origin-v4.0:`basename $(@D)`-router -f images/router/`basename $(@D)`/Dockerfile .
+	imagebuilder -t registry.ci.openshift.org/openshift/origin-v4.0:`basename $(@D)`-router -f images/router/`basename $(@D)`/Dockerfile .
```
rfredette, line 36: These Makefile changes were part of what I needed to get local router builds working, just fyi.

I didn't actually intend to push my changes to the Makefile, but since the existing Makefile references `registry.svc.ci.openshift.org` which has been retired for quite some time, I don't think it's being used in the official builds anyway. 

I'm happy to revert these changes and come back to fixing it later in another PR if that's preferable.

images/router/haproxy/Dockerfile.rhel8
```diff
@@ -1,10 +1,10 @@
-FROM registry.ci.openshift.org/ocp/4.13:haproxy-router-base
+FROM registry.ci.openshift.org/ocp/4.13:base-router
```
rfredette, line 1: The actual "base" router image is named `base-router` when built from the Makefile, so I'm not sure where the `haproxy-router-base` tag comes from.

(Same as with the Makefile, I don't mind reverting this and coming back to it in another PR)

pkg/router/crl/crl.go
```diff
@@ -209,30 +388,111 @@ func getHTTPCRL(url string) (*pkix.CertificateList, error) {
...
+				log.V(2).Info(fmt.Sprintf("Removing %s because it's stale", newestDirName))
+				err := os.RemoveAll(filepath.Join(mtlsBaseDirectory, newestDirName))
+				if err != nil {
+					log.Error(err, fmt.Sprintf("Failed to remove %s", newestDirName))
 				}
-				subjectKeyId := hex.EncodeToString(authKeyId.KeyIdentifier)
-				crlForKeyId[subjectKeyId] = crl
 			}
+			newestDirName = entryName
 		}
```
Miciah, line 482: Are you sure you want this `else` branch?  I thought the idea was to use the timestamp of `mtlsLatestSymlink`'s target (i.e. `mostRecentTimestamp`) as the boundary:

* Delete things older than `mostRecentTimestamp` (i.e. the timestamp of `mtlsLatestSymlink`'s target).
* Keep `mtlsLatestSymlink`, its target, and anything newer than `mostRecentTimestamp`.

Using `mostRecentTimestamp = mtlsLatestSymlink` as an immutable boundary makes it easy to reason about what gets deleted.  But now you're moving the boundary, which makes this code more difficult to reason about, and it isn't clear why you need to change the boundary while you're using it.  

In particular, it seems to me like you could end up in trouble if somehow you have multiple newer tmpdirs and the newest one isn't the next one that you'll use as the target of `mtls-latest`:

| filename          | mtime |
|-------------------|-------|
| `mtls-latest`     |     1 |
| `mtls-latest.new` |     2 |
| `tmpdir1`         |     4 |
| `tmpdir2`         |     3 |

When you process `tmpdir1`, you set `mostRecentTimestamp` to 4, and then when you process `tmpdir2`, you end up deleting it, but maybe `tmpdir2` was to be the next target of `mtls-latest`!  Is that not possible with this logic?

pkg/router/crl/crl.go
```diff
@@ -209,30 +388,111 @@ func getHTTPCRL(url string) (*pkix.CertificateList, error) {
...
 	}
-	return crlForKeyId, nil
+	log.V(4).Info("Cleanup done")
+	return nil
+}
+
+// newMTLSDirectory creates a new temp directory for writing the CA bundle and CRL file, and returns the full path to
+// that directory. This directory will be cleaned up by the mTLS directory reaper once a newer version is present
+func newMTLSDirectory() (string, error) {
+	subdir := fmt.Sprintf("mtls-%s", time.Now().Format(time.RFC3339))
```
Miciah, line 491: If anyone ever tries to port OpenShift router to MS Windows, they will probably be annoyed by this code.  * grin *.  Would `fmt.Sprintf("mtls-%d", time.Now().Unix())` be better?  Or simply use `os.MkdirTemp(mtlsBaseDirectory, "mtls-")`?

pkg/router/crl/crl.go
```diff
@@ -11,80 +11,275 @@ import (
...
+// authorityKeyIdentifierOID is the ASN.1 object identifier for the authority key identifier extension.
 var authorityKeyIdentifierOID = asn1.ObjectIdentifier{2, 5, 29, 35}
 
-func ManageCRLs(caBundleFilename, crlFilename string, caUpdateChannel <-chan struct{}) error {
-	nextUpdate, err := updateCRLFile(caBundleFilename, crlFilename)
-	if err != nil {
-		return err
+// ManageCRLs reads the file at caBundleFilename, ensures that the file at crlFilename contains all specified crls, and
+// updates any crls that have expired. ManageCRLs does this synchronously initially, then spins off a goroutine to
+// update CRLs when necessary in the future. Returns a boolean that's true if any of the certificates in the CA bundle
```
Miciah, line 63: Might as well mention both goroutines.
```suggestion
// updates any CRLs that have expired. ManageCRLs does this synchronously initially, then spins off a goroutine to
// update CRLs when necessary in the future and another goroutine to reap stale CRLs. Returns a Boolean that's
// true if any of the certificates in the CA bundle
```

pkg/util/util.go
```diff
@@ -0,0 +1,19 @@
...
+	in, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer in.Close()
+	out, err := os.Create(dest)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
```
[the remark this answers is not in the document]
frobware: Yes, please handle errors from Close(). That can and will fail if there's buffered output and the file system is full.

pkg/router/crl/crl.go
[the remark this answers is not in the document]
frobware: If the failure was transient should we wait for another day before downloading again? I could see a case where we do a number of retries before falling back to downloading once a day. Otherwise (and assuming a transient download failure) the only way to get a new list (in a hurry) is to delete the pod.

pkg/router/crl/crl.go
```diff
@@ -0,0 +1,211 @@
...
+	var errs []error
+	for _, distributionPoint := range distributionPoints {
+		// The distribution point is typically a URL with the "http" scheme.  "https" is generally not used because the
+		// certificate list is signed, and because using TLS to get the certificate list could introduce a circular
+		// dependency (cannot use TLS without the revocation list, and cannot get the revocation list without using
+		// TLS).
+		//
+		// TODO Support ldap.
+		switch {
+		case strings.HasPrefix(distributionPoint, "http:"):
```
frobware, line 352: Maybe use `url.Parse(s)` and check the scheme. The string may begin with "http:" but the remainder could be bogus.

pkg/router/template/router.go
```diff
@@ -438,6 +439,9 @@ func (r *templateRouter) watchMutualTLSCert() error {
 			log.V(0).Info("failed to establish watch on mTLS certificate directory", "error", err)
 			return nil
 		}
+		if err := crl.UpdateCRLFile(caPath, "/crls-test"); err != nil {
```
frobware: Why does the path end in "-test"? Is this a temporary location?

pkg/router/template/router.go
```diff
@@ -439,8 +439,11 @@ func (r *templateRouter) watchMutualTLSCert() error {
 			log.V(0).Info("failed to establish watch on mTLS certificate directory", "error", err)
 			return nil
 		}
-		if err := crl.UpdateCRLFile(caPath, "/crls-test"); err != nil {
-			log.V(0).Info("failed to download CRLs", "error", err)
+		crlDir := os.Getenv("ROUTER_MUTUAL_TLS_AUTH_CRL_DIR")
+		if len(crlDir) != 0 {
+			if err := crl.UpdateCRLFile(caPath, crlDir+"/crls.pem"); err != nil {
+				log.V(0).Info("failed to download CRLs", "error", err)
```
frobware: s/Info/Error? Seems more fatal than informational.

Makefile
frobware: > These Makefile changes were part of what I needed to get local router builds working, just fyi.

Then let's not commit them. I'd rather not mix make/build machinery as part of the fix if we don't have to.

pkg/router/template/router.go
rfredette: Yes, it was just a temp location. The CRLs are now being stored in temp dirs in `/var/lib/haproxy/mtls`.

pkg/router/crl/crl.go
rfredette: I haven't made this change yet, but it does look like `url.Parse` would be a good choice here

pkg/router/crl/crl.go
rfredette: Thanks for mentioning `os.MkdirTemp`; I wasn't familiar with that, but it fits this use case very well. Using that, `newMTLSDirectory` basically boiled down to `return os.MkdirTemp(...)`, so I removed the function and now it just calls `MkdirTemp` directly

pkg/router/crl/crl.go
Miciah: `getHTTPCRL` uses `http.Get`, and I expect `http.Get` will return an error if the URL is invalid, in which case using `url.Parse` here just for the extra error checking would be superfluous.  Really all we care about is whether the scheme is "http:" as opposed to "ldap:", so it makes sense to me to keep it simple.

pkg/router/crl/crl.go
```diff
@@ -0,0 +1,441 @@
...
+			continue
+		}
+		if err := os.RemoveAll(fullPath); err != nil {
+			log.Error(err, "Failed to remove stale directory", "directory", fullPath)
+			errs = append(errs, err)
+		}
+	}
+	log.V(4).Info("Cleanup done")
+	if len(errs) != 0 {
+		return kerrors.NewAggregate(errs)
```
Miciah, line 438: It usually isn't necessary to log _and_ return an error.  In this case, the caller logs any error that it gets, so the router will end up logging each error twice.  Is there any reason not simply to log errors in `reapStaleDirectories` and eliminate its return value?

pkg/router/template/router.go
Miciah: The comment and function name don't match.  (Did this get fixed earlier and then unfixed?)

pkg/router/template/router.go
```diff
@@ -430,22 +474,23 @@ func (r *templateRouter) writeDefaultCert() error {
 func (r *templateRouter) watchMutualTLSCert() error {
 	caPath := os.Getenv("ROUTER_MUTUAL_TLS_AUTH_CA")
 	if len(caPath) != 0 {
-		reloadFn := func() {
-			log.V(0).Info("reloading to get updated client CA", "name", caPath)
+		r.haveClientCA = true
+		caUpdateChannel := make(chan struct{})
+		crlReloadFn := func(haveCRLs bool) {
+			r.haveCRLs = haveCRLs
+			log.V(0).Info("reloading to get updated client CA CRL", "name", crl.CRLFilename)
```
Miciah, line 481:
```suggestion
			log.V(0).Info("reloading to get updated client CA CRL", "name", crl.CRLFilename, "have CRLs", haveCRLs)
```
Or is that redundant with the logging that `managedURLs` already does when `nextUpdate` is zero?

pkg/util/util.go
```diff
@@ -0,0 +1,39 @@
...
+	if err != nil {
+		return err
+	}
+	_, err = out.ReadFrom(in)
+	if err != nil {
+		if err := out.Close(); err != nil {
+			log.Error(err, "Failed to close output file", "filename", dest)
+			return err
+		}
+		return err
```
Miciah, line 32: I think you want to log the error from `Close` and return the error from `ReadFrom`, right?  No need to log _and_ return the same error (which in this case will ultimately result in the error getting logged twice when the caller logs the returned error).  
```suggestion
		if err := out.Close(); err != nil {
			log.Error(err, "Failed to close output file", "filename", dest)
		}
		return err
```

pkg/util/util.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	if err := out.Close(); err != nil {
+		log.Error(err, "Failed to close output file", "filename", dest)
+		return err
+	}
+	return err
```
Miciah, line 38: Hm, maybe this?
```suggestion
	if err != nil {
		if err := out.Close(); err != nil {
			log.Error(err, "Failed to close output file", "filename", dest)
		}
		return err
	}
	return out.Close()
```

Miciah (commented):
I like the latest changes.  The cleanup of stale mTLS config looks much more straightforward now.  Nice improvement!  

However, several comments from my previous review don't appear to have been addressed.  Can you make sure that past review comments haven't been overlooked?  It's fine if you decide to reject some of the suggestions, but please reply in those cases so that I know a comment wasn't inadvertently missed.

pkg/router/crl/crl.go
```diff
@@ -11,80 +11,275 @@ import (
...
+
+// commitCACRLUpdate makes sure stagingDirectory contains up-to-date versions of both the CA bundle and the CRL file,
+// then updates the symlink so that HAProxy will reference the new versions on reload
+func commitCACRLUpdate(stagingDirectory, caBundleFilename string, crlUpdated bool) error {
+	// Copy CA bundle to the new directory
+	if _, err := os.Stat(caBundleFilename); err == nil {
+		if err := util.CopyFile(caBundleFilename, filepath.Join(stagingDirectory, caBundleBasename)); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
rfredette: We should return an error if `os.Stat` fails, although in revisiting this, I'm not sure `os.Stat` is necessary here. `CopyFile` should return an error if `caBundleFilename` points to a non-existant file or a file that `openshift-router` can't access

pkg/router/crl/crl.go
```diff
@@ -11,80 +11,275 @@ import (
...
 }
 
+// authorityKeyIdentifierOID is the ASN.1 object identifier for the authority key identifier extension.
 var authorityKeyIdentifierOID = asn1.ObjectIdentifier{2, 5, 29, 35}
 
-func ManageCRLs(caBundleFilename, crlFilename string, caUpdateChannel <-chan struct{}) error {
-	nextUpdate, err := updateCRLFile(caBundleFilename, crlFilename)
-	if err != nil {
-		return err
+// ManageCRLs reads the file at caBundleFilename, ensures that the file at crlFilename contains all specified crls, and
```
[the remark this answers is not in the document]
rfredette: The godoc is out of date; crlFilename was a function parameter in an earlier iteration. I'll fix this in my next update

pkg/router/crl/crl.go
```diff
@@ -0,0 +1,441 @@
...
+
+const (
+	// crlFilePermissions is the permission bits used by the CRL file.
+	crlFilePermissions = 0644
+	// crlFallbackTime is how long to wait before retrying if nextUpdate can't be determined.
+	crlFallbackTime = 5 * time.Minute
+	// errorBackoffTime is how long to wait before retrying if a generic error happens during CRL refresh.
+	errorBackoffTime = 5 * time.Minute
+	// mtlsBaseDirectory is the directory where all the files and temp directories created here will live.
+	mtlsBaseDirectory = "/var/lib/haproxy/mtls/"
```
frobware, line 33: Ryan, I think we discussed this last week--as you use path.Join() we can drop the trailing / character.

pkg/router/crl/crl.go
rfredette: I think it makes sense to eliminate the return value, and log any errors that occur instead. `updateCRLFile` already only calls `reapStaleDirectories` in a `defer` call, so returning an error and logging it in the `defer` doesn't add anything. I missed this in my latest change, but I'll update it shortly.

pkg/router/crl/crl.go
rfredette: After talking with Andy, I switched back to using a date-based ID in the mtls directory names. I think it makes the directory names easier to remember when debugging, and it should still result in unique directory names.

rfredette:
I plan to remove the build changes I made (in Makefile and the haproxy Dockerfile.rhel8), set the retry timeout to a final value, probably larger than 5 minutes, and squash the commits before this merges. It still needs some cleanup in the tags, but just in case:

/hold

pkg/router/crl/crl.go
```diff
@@ -0,0 +1,529 @@
...
+}
+
+// getHTTPCRL gets a certificate revocation list using the provided HTTP URL.
+func getHTTPCRL(url string) (*x509.RevocationList, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, fmt.Errorf("http.Get failed: %w", err)
+	}
+	defer resp.Body.Close()
+	crlBytes, err := ioutil.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
rfredette: Looks like all the `ioutil` functionality has been moved to the `io` package. I'll update these calls to use the new functions.

It does look like there are other places in the router codebase where `ioutil` functions are used, so in a separate PR we should probably also change those over to `io`.

pkg/router/crl/crl.go
```diff
@@ -0,0 +1,529 @@
...
+		nextUpdate := time.Now()
+		shouldHaveCRLs, err := ShouldHaveCRLs(caBundleFilename)
+		if err != nil {
+			log.Error(err, "failed to parse CA bundle", "CA bundle filename", caBundleFilename)
+			// TODO: Should we back off if we can't parse the CA bundle, or just wait until the CA bundle is updated again?
+			nextUpdate = time.Now().Add(errorBackoffTime)
+		}
+		for {
+			updated := false
+			if nextUpdate.IsZero() {
```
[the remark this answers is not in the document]
rfredette: I think I tested this before, but I couldn't be sure, so I ran it again: https://go.dev/play/p/_1F4jY_QdYf
Looks like if the `duration` in `time.After(duration)` is negative, it fires immediately

pkg/router/crl/crl.go
```diff
@@ -0,0 +1,529 @@
...
+	}
+	return false, nil
+}
+
+// ManageCRLs spins off a goroutine that ensures that any CRLs specified in caBundleFilename are downloaded and kept
+// up-to-date. It will automatically refresh expired CRLs and download missing CRLs when it receives a message on
+// caUpdateChannel (indicating the CA bundle has been updated), or when any existing CRL expires. Whenever either the CA
+// bundle or the CRL file has changed, updateCallback is called, with a boolean indicating whether crl-file needs to be
+// specified in the HAProxy config.
+func ManageCRLs(caBundleFilename string, caUpdateChannel <-chan struct{}, updateCallback func(bool)) {
```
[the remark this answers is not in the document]
rfredette: Most (if not all) of the functions I've added either deal with downloading data from another source, or read and write files in fixed locations on the filesystem, so isolating them to run in a unit test scenario would take some work. If I'm able to get good enough coverage with e2e tests, what do you think about me circling back to figure out unit tests in a later PR?

pkg/router/crl/crl.go
```diff
@@ -0,0 +1,529 @@
...
+	// but at initialization, mtlsLatestSymlink shouldn't exist yet, so it should be safe to just directly create the
+	// symlink.
+	if err := os.Symlink(stagingDirectory, mtlsLatestSymlink); err != nil {
+		return err
+	}
+	return nil
+}
+
+// ShouldHaveCRLs returns true if any of the certificates in caBundleFilename specify a CRL distribution point
+func ShouldHaveCRLs(caBundleFilename string) (bool, error) {
```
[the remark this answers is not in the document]
rfredette: My thought when naming this was that it's returning whether or not there will be CRLs once the download happens, but now that I revisit it, I think it would be better to name it `CABundleHasCRLs` or something similar to better describe what it's actually checking

candita:
Assigning Miciah and Andy due to their current level of feedback.  Feel free to remove, and add to PR Scrub instead.
/assign @Miciah 
/assign @frobware

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -290,8 +292,8 @@ frontend fe_sni
     {{- "" }} crt-list /var/lib/haproxy/conf/cert_config.map accept-proxy
     {{- with (env "ROUTER_MUTUAL_TLS_AUTH") }}
       {{- "" }} verify {{. }}
-    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CA") }} ca-file {{. }} {{ else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
-    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CRL") }} crl-file {{. }} {{ end }}
+	{{- if $haveClientCA }} ca-file /var/lib/haproxy/mtls/latest/ca-bundle.pem {{ else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
+	{{- if $haveCRLs }} crl-file /var/lib/haproxy/mtls/latest/crls.pem {{ end }}
```
Miciah, line 296: You replaced spaces with tabs.  Please use spaces for consistency with the rest of the template.  

I'm wary of breaking OKD users, or people using custom router deployments (unsupported, but better not to break them if we don't need to).  If `ROUTER_MUTUAL_TLS_AUTH_CA` or `ROUTER_MUTUAL_TLS_AUTH_CRL` is set, we should still respect it.  Likewise for `ROUTER_MUTUAL_TLS_AUTH_CRL` below.  
```suggestion
    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CA") }} ca-file {{ . }} {{- else }}{{ if $haveClientCA }} ca-file /var/lib/haproxy/mtls/latest/ca-bundle.pem {{- else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{- end }}{{ end }}
    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CRL") }} crl-file {{ . }} {{- else }}{{ if $haveCRLs }} crl-file /var/lib/haproxy/mtls/latest/crls.pem {{- end }}{{ end }}
```
Likewise for `ROUTER_MUTUAL_TLS_AUTH_CRL` below.

rfredette: My router changes still use `ROUTER_MUTUAL_TLS_AUTH_CA` since the operator still needs to supply the CA bundle, so using `with (env "ROUTER_MUTUAL_TRLS_AUTH_CA")` by itself to select the old behavior isn't specific enough.

What do you think of something like this?
```suggestion
    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CRL") }} crl-file {{ . }}
        {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CA") }} ca-file {{ . }} {{- else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
    {{- else }}
        {{- if $haveClientCA }} ca-file /var/lib/haproxy/mtls/latest/ca-bundle.pem {{- else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
        {{- if $haveCRLs }} crl-file /var/lib/haproxy/mtls/latest/crls.pem {{ end }}
    {{- end}}
```

Miciah: That could work (though I think you also need to re-add `{{- with (env "ROUTER_MUTUAL_TLS_AUTH_CRL") }} crl-file {{. }} {{ end }}`).

rfredette: The fix I proposed does include `crl-file ...`, but maybe it's a bit confusing. I'll see if I can make it clearer.

rfredette: I've included a clearer version of this fix in #472.

rfredette:
Closing in favor of #472, which includes this fix as well as a fix for OCPBUGS-9464.

PR #472, opened 2023-04-20
 OCPBUGS-6661, OCPBUGS-9464: Handle mTLS CRLs, and fix accidental CRL duplication

rfredette:
This PR makes the router parse the certificates supplied in ROUTER_MUTUAL_TLS_AUTH_CA, download CRLs from any CRL distribution points that are found, and write them to a file that is supplied to HAProxy for mTLS certificate validation.

It also includes a fix for an issue where, when a CRL is distributed in a certificate that is not its issuer, the CRL would sometimes be missing or two copies of the CRL would be present.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -9,6 +9,8 @@
 {{- $dynamicConfigManager := .DynamicConfigManager }}
 {{- $router_ip_v4_v6_mode := env "ROUTER_IP_V4_V6_MODE" "v4" }}
 {{- $router_disable_http2 := env "ROUTER_DISABLE_HTTP2" "false" }}
+{{- $haveClientCA := .HaveClientCA }}
+{{- $haveCRLs := .HaveCRLs }}
 
 
 {{- /* A bunch of regular expressions.  Each should be wrapped in (?:) so that it is safe to include bare */}}
@@ -290,8 +292,13 @@ frontend fe_sni
     {{- "" }} crt-list /var/lib/haproxy/conf/cert_config.map accept-proxy
     {{- with (env "ROUTER_MUTUAL_TLS_AUTH") }}
       {{- "" }} verify {{. }}
-    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CA") }} ca-file {{. }} {{ else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
-    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CRL") }} crl-file {{. }} {{ end }}
+      {{- if (ne (env "ROUTER_MUTUAL_TLS_AUTH_CRL") "") }}
+        {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CA") }} ca-file {{. }} {{ else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
+        {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CRL") }} crl-file {{. }} {{ end }}
+      {{- else }}
+        {{- if $haveClientCA }} ca-file /var/lib/haproxy/mtls/latest/ca-bundle.pem {{ else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
+        {{- if $haveCRLs }} crl-file /var/lib/haproxy/mtls/latest/crls.pem {{ end }}
+      {{- end }}
     {{- end }}
   mode http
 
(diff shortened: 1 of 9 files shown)


pkg/router/crl/crl.go
```diff
@@ -0,0 +1,530 @@
...
+	var nextCRLUpdate time.Time
+	crls := make(map[string]*x509.RevocationList)
+	updated := false
+	now := time.Now()
+	for len(clientCAData) > 0 {
+		block, data := pem.Decode(clientCAData)
+		if block == nil {
+			break
+		}
+		clientCAData = data
```
frobware, line 335: In `clientCAData` are we only concerned with certificates?

```go
if block.Type != "CERTIFICATE" {
    return nil, time.Time{}, false, fmt.Errorf("client CA bundle contains a non-certificate PEM block")
}
```

pkg/router/template/router.go
```diff
@@ -430,23 +439,29 @@ func (r *templateRouter) writeDefaultCert() error {
...
-		if err := r.watchVolumeMountDir(filepath.Dir(crlPath), reloadFn); err != nil {
-			log.V(0).Info("failed to establish watch on mTLS certificate directory", "error", err)
+		crl.ManageCRLs(caPath, caUpdateChannel, crlReloadFn)
+		caReloadFn := func() {
+			// Send signal to CRL management goroutine that client CA has been changed
+			caUpdateChannel <- struct{}{}
+		}
+		if err := r.watchVolumeMountDir(filepath.Dir(caPath), caReloadFn); err != nil {
+			log.V(0).Error(err, "failed to establish watch on mTLS certificate directory")
 			return nil
```
frobware, line 465: Why does this not return the error?

frobware (changes requested):

pkg/router/template/router.go
frobware: @Miciah I noticed that `writeDefaultCert` doesn't return an error for its watch on calls to `watchVolumeMountDir()` either - do you know of any history for why this returns nil?

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -290,8 +292,13 @@ frontend fe_sni
     {{- "" }} crt-list /var/lib/haproxy/conf/cert_config.map accept-proxy
     {{- with (env "ROUTER_MUTUAL_TLS_AUTH") }}
       {{- "" }} verify {{. }}
-    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CA") }} ca-file {{. }} {{ else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
-    {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CRL") }} crl-file {{. }} {{ end }}
+      {{- if (ne (env "ROUTER_MUTUAL_TLS_AUTH_CRL") "") }}
+        {{- with (env "ROUTER_MUTUAL_TLS_AUTH_CA") }} ca-file {{ . }} {{- else }} ca-file /etc/ssl/certs/ca-bundle.trust.crt {{ end }}
```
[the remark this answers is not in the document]
rfredette: It's not an intentional change; I will update this to get rid of that drift.

pkg/router/crl/crl.go
rfredette: We should only care about certificates; any other PEM block type shouldn't matter. I think log & skip any non-certificate content is the way to go.

pkg/router/template/router.go
frobware: If we were to return an error I'd propose we do that change in 4.14+ only for the moment as we want the backports to discretely fix OCPBUGS-6661 and OCPBUGS-9464 only.

images/router/haproxy/conf/haproxy-config.template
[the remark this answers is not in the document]
rfredette: Done

pkg/router/crl/crl.go
rfredette: Done

pkg/router/crl/crl.go
```diff
@@ -0,0 +1,530 @@
...
+
+// getHTTPCRL gets a certificate revocation list using the provided HTTP URL. Returns an error if the CRL could not be
+// downloaded, or if parsing the CRL fails.
+func getHTTPCRL(url string) (*x509.RevocationList, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, fmt.Errorf("http.Get failed: %w", err)
+	}
+	defer resp.Body.Close()
+	crlBytes, err := io.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
rfredette: Done

PR #483, opened 2023-05-19
OCPBUGS-6958: Fix clipHAProxyTimeoutValue

candita:
Fix clipHAProxyTimeoutValue so that:
- a value larger than time.ParseDuration can handle is clipped to the HAProxy max timeout
- a value that cannot be properly parsed for other reasons is set to empty instead of being silently allowed

To check that `time.ParseDuration` is experiencing overflow, add a new `ParseDuration` so we can evaluate the errors that wouldn't have been explicitly returned by `time.ParseDuration`, e.g. invalid HAProxy time format syntax and integer overflows.

Add more unit tests.


pkg/router/template/template_helper.go
```diff
@@ -26,6 +26,9 @@ const (
 	certConfigMap = "cert_config.map"
 	// max timeout allowable by HAProxy
 	haproxyMaxTimeout = "2147483647ms"
+	// default timeout to use when the timeout value is unparseable
+	// for reasons other than it is too large
+	haproxyDefaultTimeout = "5s"
```
Miciah, line 31:
```suggestion
	// haproxyDefaultTimeout is the default timeout to use when the timeout value is unparseable
	// for reasons other than it is too large.
	haproxyDefaultTimeout = "5s"
```

pkg/router/template/template_helper.go
```diff
@@ -335,6 +338,7 @@ func clipHAProxyTimeoutValue(val string) string {
 	if err != nil {
 		return val
 	}
```
Miciah, line 340: BTW, it seems silly to reparse `haproxyMaxTimeout` every time `clipHAProxyTimeoutValue` is called.  Why not declare it as a package variable?
```go
// haproxyMaxTimeoutDuration is haproxyDefaultTimeout as a time.Duration value.
var haproxyMaxTimeoutDuration = func() time.Duration {
	d, err := time.ParseDuration(haproxyMaxTimeout)
	if err != nil {
		panic(err)
	}
	return d
}()
```

pkg/router/template/template_helper.go
```diff
@@ -347,9 +351,20 @@ func clipHAProxyTimeoutValue(val string) string {
 			return haproxyMaxTimeout
 		}
 	} else {
-		duration, err := time.ParseDuration(val)
+		var duration, err = templateutil.ParseDurationWithErrors(val)
```
Miciah, line 354: Why change `:=` to `var`?

Why not add support for the "d" unit to `templateutil.ParseDurationWithErrors`? Then `clipHAProxyTimeoutValue` wouldn't need to special-case the `val[endIndex] == 'd'` case, and it would properly handle values like "1d1s".

pkg/router/template/template_helper.go
```diff
@@ -347,9 +351,20 @@ func clipHAProxyTimeoutValue(val string) string {
...
 		if err != nil {
-			return val
+			// Here we need to check if the value is greater than what time.ParseDuration
+			// allows (value must be representable as uint64, e.g. approximately 2562047.79 hours
+			// or 9223372036854775807 nanoseconds).
+			if strings.Contains(err.Error(), "overflow") {
+				log.V(7).Info("Route annotation timeout exceeds maximum allowable by HAProxy, clipping to max")
+				return haproxyMaxTimeout
+			} else if strings.Contains(err.Error(), "missing unit") {
+				// ignore this error, because we have allowed this in the past
```
Miciah, line 363: What if `val` is "9999999999999999"?

pkg/router/template/util/util.go
```diff
@@ -94,3 +96,176 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+		d += v
+		if d > 1<<63 {
+			return 0, errors.New("time: invalid duration, overflow " + orig)
+		}
+	}
+	if neg {
+		return -time.Duration(d), nil
+	}
+	if d > 1<<63-1 {
+		return 0, errors.New("time: invalid duration, overflow " + orig)
```
Miciah, line 201: Any reason not simply to return `haproxyMaxTimeout` in case of overflow?  

If you prefer to return an error value here and handle overflow in `clipHAProxyTimeoutValue`, consider using a typed error value, something like this:
```go
// OverflowError represents an overflow error from ParseDurationWithErrors.
type OverflowError struct {
	error
}

func ParseDurationWithErrors(s string) (time.Duration, error) {
	// ...
	return 0, OverflowError{errors.New("time: invalid duration, overflow " + orig)}
	// ...
}

func clipHAProxyTimeoutValue(val string) string {
	// ...
	duration, err := templateutil.ParseDurationWithErrors(val)
	switch err.(type) {
	case nil:
	case templateutil.OverflowError:
		return haproxyMaxTimeout
	default:
		return haproxyDefaultTimeout
	}
	// ...
}

File: pkg/router/template/util/util.go [hunk 6]
```diff
@@ -94,3 +96,176 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+		}
+		x = y
+		scale *= 10
+	}
+	return x, scale, s[i:]
+}
+
+// UnitMap is used by ParseDurationWithErrors above, but made public because it
+// could be useful for other validation in the future.
+var UnitMap = map[string]uint64{
```
Miciah, line 262: `Unitmap` can always be made public later as needed.

ShudiLi:
From QE side, tested it with 4.13.0-0.ci.test-2023-05-22-014550-ci-ln-qsqqkkt-latest 
1.
% oc get clusterversion
version   4.13.0-0.ci.test-2023-05-22-014550-ci-ln-qsqqkkt-latest   True        False         10m     Cluster version is 4.13.0-0.ci.test-2023-05-22-014550-ci-ln-qsqqkkt-latest

2.
% oc -n openshift-ingress get pods
NAME                             READY   STATUS    RESTARTS   AGE
router-default-c9ccc9cc7-hcf2f   1/1     Running   0          32m
router-default-c9ccc9cc7-pj8b2   1/1     Running   0          32m
%

3. create pod, service and the route

4.
% oc annotate route service-unsecure haproxy.router.openshift.io/timeout=100000000000s  --overwrite 
route.route.openshift.io/service-unsecure annotated
% 

5.
% oc get route service-unsecure -oyaml | grep timeout
    haproxy.router.openshift.io/timeout: 100000000000s

6.
% oc -n openshift-ingress  rsh router-default-c9ccc9cc7-hcf2f cat haproxy.config | grep -A5 be_http:default:service-unsecure
backend be_http:default:service-unsecure
  mode http
  option redispatch
  option forwardfor
  balance random
  timeout server  2147483647ms

7.
 % oc -n openshift-ingress get pods
NAME                             READY   STATUS    RESTARTS   AGE
router-default-c9ccc9cc7-hcf2f   1/1     Running   0          52m
router-default-c9ccc9cc7-pj8b2   1/1     Running   0          52m
% 
% oc -n openshift-ingress logs router-default-c9ccc9cc7-hcf2f | grep -i error  
%

8.
% curl http://service-unsecure-default.apps.ci-ln-qsqqkkt-72292.origin-ci-int-gce.dev.rhcloud.com
Hello-OpenShift-1 http-8080
%

/label qe-approved
thanks

candita:
Assigning to @Miciah because he has provided feedback.
/assign @Miciah

candita:
@frobware has an implementation as well: https://github.com/frobware/haproxytime/tree/main

pkg/router/template/util/util.go
candita: It has its pros and cons.  If a user mistakenly typed "h" instead of "s", this could really cause a problem.  If they used a value that was only a small duration of time away from the max timeout, it would be fine.

pkg/router/template/util/util.go
```diff
@@ -94,3 +96,176 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+	neg := false
+
+	// Consume [-+]?
+	if s != "" {
+		c := s[0]
+		if c == '-' || c == '+' {
+			neg = c == '-'
+			s = s[1:]
+		}
+	}
```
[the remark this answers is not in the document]
candita: Hm, if people used a +/- in the past, and it had no noticeable effect, would it be ok for me to make these changes now?

pkg/router/template/util/util.go
```diff
@@ -94,3 +96,176 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+		pre := pl != len(s) // whether we consumed anything before a period
+
+		// Consume (\.[0-9]*)?
+		post := false
+		if s != "" && s[0] == '.' {
+			s = s[1:]
+			pl := len(s)
+			f, scale, s = leadingFraction(s)
+			post = pl != len(s)
+		}
```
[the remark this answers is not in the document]
candita: [durationToHAProxyTimespec](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/deployment.go#L1653) should remove the decimal points.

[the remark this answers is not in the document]
candita: Hm, but that's not for annotations, which is what this bug deals with.

pkg/router/template/template_helper.go
candita: i will add the missing "ms" that HAProxy tacks on when the unit is missing instead.

pkg/router/template/util/util.go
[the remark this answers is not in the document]
candita: I'll just go ahead and disallow it.

pkg/router/template/util/util.go
```diff
@@ -94,3 +96,176 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+		i := 0
+		for ; i < len(s); i++ {
+			c := s[i]
+			if c == '.' || '0' <= c && c <= '9' {
+				break
+			}
+		}
+		if i == 0 {
+			return 0, errors.New("time: missing unit in duration " + orig)
+		}
```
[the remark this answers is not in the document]
candita: Done.

pkg/router/template/util/util.go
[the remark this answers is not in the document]
candita: it does not allow a decimal point, removed the code supporting this and updated unit tests.

pkg/router/template/util/util.go
candita: I incorporated this suggestion, thanks.

candita:
e2e-agnostic failure:

failed: (55.3s) 2023-06-09T17:20:52 "[sig-api-machinery] Aggregator Should be able to support the 1.17 Sample API Server using the current Aggregator [Conformance] [Suite:openshift/conformance/parallel/minimal] 

Recent changes for Kube bump:
https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/test/e2e/apimachinery/aggregator.go#L92

candita:
Seems like a cluster install issue, and it's repeated with different errors for days.

/test e2e-agnostic

pkg/router/template/util/util.go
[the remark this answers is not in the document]
Miciah: On further thought, it makes sense to disallow it irrespective of what HAProxy allows as `$timeSpecPattern` already disallows it.  There's no need to have logic to parse something that would be rejected by the regexp anyway.

pkg/router/template/template_helper.go
Miciah: I was concerned that ignoring the error and passing "9999999999999999" through would result in an unparseable `haproxy.config`.  Looks like you handle this case properly now and cap "9999999999999999" to the maximum value now.

pkg/router/template/util/util.go
Miciah: > It has its pros and cons. If a user mistakenly typed "h" instead of "s", this could really cause a problem. If they used a value that was only a small duration of time away from the max timeout, it would be fine.

I meant, rather than having `ParseDurationWithErrors` (which is now named `ParseHAProxyDuration` with your latest changes) return an overflow error and then having `clipHAProxyTimeoutValue` handle the overflow error by using the max value, why not simply have the parse function return the max value in case of overflows?  If you prefer returning an overflow error and handling it in `clipHAProxyTimeoutValue`, that's fine, I was just thinking you could simplify the logic a little.

pkg/router/template/template_helper.go
```diff
@@ -321,7 +317,7 @@ func generateHAProxyMap(name string, td templateData) []string {
 
 // clipHAProxyTimeoutValue prevents the HAProxy config file
 // from using timeout values specified via the haproxy.router.openshift.io/timeout
-// annotation that exceed the maximum value allowed by HAProxy.
+// annotation that exceed the maximum value allowed by HAProxy, or by time.ParseDuration.
 // Return the parameter string instead of an err in the event that a
 // timeout string value is not parsable as a valid time duration.
```
Miciah, line 322:
```suggestion
// Return the empty string instead of an err in the event that a
// timeout string value is not parsable as a valid time duration.
```

pkg/router/template/template_helper.go
```diff
@@ -330,32 +326,29 @@ func clipHAProxyTimeoutValue(val string) string {
...
+			return ""
+		default:
+			// This is not used at the moment
+			log.V(7).Info("Invalid route annotation timeout, setting to 5s")
+			return templateutil.HaproxyDefaultTimeout
 		}
 	}
+
+	// Then check to see if the timeout is larger than what HAProxy allows
+	if templateutil.HaproxyMaxTimeoutDuration.Milliseconds() < duration.Milliseconds() {
```
Miciah, line 348:
```suggestion
	if templateutil.HaproxyMaxTimeoutDuration < duration {
```

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,199 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
 
 	return prefix
 }
+
+var haproxyDurationRE = regexp.MustCompile(`^(([0-9]+)(us|ms|s|m|h|d)?)+$`)
+
+// HaproxyMaxTimeoutDuration is haproxyMaxTimeout as a time.Duration value.
```
Miciah, line 110:
```suggestion
// HaproxyMaxTimeoutDuration is HaproxyMaxTimeout as a time.Duration value.
```

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,199 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+}()
+
+// HaproxyDefaultTimeoutDuration is haproxyDefaultTimeout as a time.Duration value.
+var HaproxyDefaultTimeoutDuration = func() time.Duration {
+	d, err := time.ParseDuration(HaproxyDefaultTimeout)
+	if err != nil {
+		panic(err)
+	}
+	return d
+}()
```
Miciah, line 126: `HaproxyDefaultTimeoutDuration` isn't used anywhere.

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,199 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+// InvalidInputError represents an error based on invalid input to ParseHAProxyDuration
+type InvalidInputError struct {
+	error
+}
+
+// ParseHAProxyDuration is similar to time.ParseDuration, but modified with meaningful
+// error messages for invalid input.
+// It parses a duration string and returns a suitable duration. A duration string is a
+// sequence of decimal numbers, each with optional fraction and a unit suffix,
+// such as "300ms", "1.5h" or "2h45m".
```
Miciah, line 145:
```suggestion
// ParseHAProxyDuration is similar to time.ParseDuration, but modified with meaningful
// error messages for invalid input, with support for decimal points removed, with
// support for the "d" unit for days added, and with "ms" as the default unit.
// It parses a duration string and returns a suitable duration. A duration string is a
// sequence of decimal numbers, each with optional fraction and a unit suffix,
// such as "300ms", "1h" or "2h45m". If the unit suffix is omitted, "ms" is used.
```

Miciah (commented):
BTW, as a follow-up, it might make sense to replace `clipHAProxyTimeoutValue (firstMatch $timeSpecPattern foo)` with `clipHAProxyTimeoutValue foo` in `haproxy-config.template` as an optimization.

frobware:
The PR description has "a value that cannot be properly parsed for other reasons is set to 5s instead of being silently allowed" - will we ever back port this change in behaviour?

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,199 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+
+// InvalidInputError represents an error based on invalid input to ParseHAProxyDuration
+type InvalidInputError struct {
+	error
+}
+
+// ParseHAProxyDuration is similar to time.ParseDuration, but modified with meaningful
+// error messages for invalid input.
+// It parses a duration string and returns a suitable duration. A duration string is a
+// sequence of decimal numbers, each with optional fraction and a unit suffix,
```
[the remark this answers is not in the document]
Miciah: > What does an optional fraction look like for days, hours, minutes, seconds, milliseconds?
> 
> Do we have test cases for said fractional values?

There are a couple test cases for fractional values:

https://github.com/openshift/router/blob/a4b3f6086cf16bf15e9c35a5afb27012cb4d3bf9/pkg/router/template/template_helper_test.go#L824-L828
https://github.com/openshift/router/blob/a4b3f6086cf16bf15e9c35a5afb27012cb4d3bf9/pkg/router/template/template_helper_test.go#L838-L842

Miciah:
> The PR description has "a value that cannot be properly parsed for other reasons is set to 5s instead of being silently allowed" - will we ever back port this change in behaviour?

Seems reasonable to me if requested.  Silently allowing values that cannot be parsed breaks reloads.  

BTW, @candita, can you make sure the information in the PR description makes it into the commit message?

pkg/router/template/util/util.go
candita: My intention was that it would be more easily reused if we left it up to the caller to interpret the error and choose what to do with it.

candita:
> The PR description has "a value that cannot be properly parsed for other reasons is set to 5s instead of being silently allowed" - will we ever back port this change in behaviour?

A value that cannot be properly parsed is actually set to empty.  The template can determine whether to add a default value when it is blank.

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,199 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+	if !haproxyDurationRE.MatchString(s) {
+		return 0, InvalidInputError{errors.New("invalid duration, no pattern match " + s)}
+	}
+
+	if s == "0" {
+		return 0, nil
+	}
+	if s == "" {
+		return 0, nil
+	}
```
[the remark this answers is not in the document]
candita: Removed.

candita:
> BTW, as a follow-up, it might make sense to replace `clipHAProxyTimeoutValue (firstMatch $timeSpecPattern foo)` with `clipHAProxyTimeoutValue foo` in `haproxy-config.template` as an optimization.

Created an issue for this.

pkg/router/template/util/util.go
candita: Removed

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,130 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+func ParseHAProxyDuration(s string) (time.Duration, error) {
+	orig := s
+	var d uint64
+
+	// Compare the input to the HAProxy duration regex.
+	if !haproxyDurationRE.MatchString(s) {
+		return 0, InvalidInputError{errors.New("invalid duration, no pattern match " + s)}
+	}
+	// Set the default HAProxy unit of "ms", in case a unit wasn't provided.
+	u := "ms"
```
Miciah, line 148: For correctness, `u` needs to be initialized inside the loop so that an explicit unit for one quantity doesn't determine the default unit for the subsequent quantity.  (Stylistically, `u` should also be declared inside the loop because it isn't used outside the loop.)  Since you moved the declaration and initialization outside the loop, this test case for `TestClipHAProxyTimeoutValue` now fails:
```go
		{
			value:    "24d1",
			expected: "24d1",
		},
```
The test failure indicates that `ParseHAProxyDuration` parses the value as 24 days and 1 day, which gets capped to the maximum value:
```
--- FAIL: TestClipHAProxyTimeoutValue (0.00s)
    template_helper_test.go:884: clipHAProxyTimeoutValue yielded incorrect result: expected 24d1 but got 2147483647ms
```

Having made this observation, a couple other questions come to mind that I should have raised during earlier reviews:  First, what do you think about adding the "ms" suffix when it is missing, so that the unit is obvious from reading the rendered configuration file?  So the test case would look like this:
```go
		{
			value:    "24d1",
			expected: "24d1ms",
		},
```
Second, and more importantly, should `clipHAProxyTimeoutValue`/`PArseHAProxyDuration` even allow multiple units?  `$timeSpecPattern` does not, and I believe HAProxy does not.  Even if we modify `$timeSpecPattern` to allow multiple units, we need to convert to a single unit for HAProxy, so the test case would look like this:
```go
		{
			value:    "24d1",
			expected: "2073600001ms",
		},
```
Likewise, the input value "10" would get written out in the configuration file as "10ms".

Sorry, I should have noticed this during earlier reviews.

ShudiLi:
Tested it again with 4.14.0-0.ci.test-2023-07-06-002348-ci-ln-ipvvdnt-latest
1.
 % oc get route unsec-httpbin  -oyaml 
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  annotations:
    haproxy.router.openshift.io/timeout: 100000000000s
    openshift.io/host.generated: "true"
  creationTimestamp: "2023-07-06T01:03:43Z"
  labels:
    name: unsec-httpbin
  name: unsec-httpbin
  namespace: default
2.
% oc -n openshift-ingress logs router-default-789d48b8bd-95nsp --tail=10
I0706 00:49:03.343814       1 router.go:635] template "msg"="router reloaded" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0706 00:49:10.108815       1 router.go:635] template "msg"="router reloaded" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0706 00:49:15.102144       1 router.go:635] template "msg"="router reloaded" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0706 00:49:20.169744       1 router.go:635] template "msg"="router reloaded" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0706 00:49:42.604517       1 router.go:635] template "msg"="router reloaded" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0706 00:49:47.583552       1 router.go:635] template "msg"="router reloaded" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0706 00:50:19.991770       1 router.go:635] template "msg"="router reloaded" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0706 00:51:44.157853       1 router.go:635] template "msg"="router reloaded" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0706 01:03:43.277989       1 router.go:635] template "msg"="router reloaded" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0706 01:08:27.738848       1 router.go:635] template "msg"="router reloaded" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
3.
% oc -n openshift-ingress rsh router-default-789d48b8bd-95nsp cat haproxy.config | grep -A5 unsec-httpbin
backend be_http:default:unsec-httpbin
  mode http
  option redispatch
  option forwardfor
  balance random
  timeout server  2147483647ms
--
4.
% curl http://unsec-httpbin-default.apps.ci-ln-ipvvdnt-72292.origin-ci-int-gce.dev.rhcloud.com/headers
{
  "headers": {
    "Accept": "*/*", 
    "Forwarded": "for=111.193.129.0;host=unsec-httpbin-default.apps.ci-ln-ipvvdnt-72292.origin-ci-int-gce.dev.rhcloud.com;proto=http", 
    "Host": "unsec-httpbin-default.apps.ci-ln-ipvvdnt-72292.origin-ci-int-gce.dev.rhcloud.com", 
    "User-Agent": "curl/7.88.1", 
    "X-Forwarded-Host": "unsec-httpbin-default.apps.ci-ln-ipvvdnt-72292.origin-ci-int-gce.dev.rhcloud.com"
  }
}

/label qe-approved
thanks

pkg/router/template/util/util.go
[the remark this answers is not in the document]
candita: These are not allowed by HAProxy, as Miciah pointed out in https://github.com/openshift/router/pull/483#discussion_r1199556982 so I removed the tests and hopefully removed all the processing code for decimal points.

pkg/router/template/util/util.go
candita: Thanks @Miciah.  I've removed the ability to have multiple units as part of the string, so 1d1m is not allowed any more.  I changed the regex so multiples are not allowed, and updated the comments and unit tests.  I also moved the initialization of "u" into the loop.  PTAL when you get a chance and let me know if this is now acceptable.

pkg/router/template/template_helper.go
```diff
@@ -321,41 +317,38 @@ func generateHAProxyMap(name string, td templateData) []string {
...
-		duration, err := time.ParseDuration(val)
-		if err != nil {
-			return val
-		}
-		if maxTimeout.Milliseconds() < duration.Milliseconds() {
-			log.V(7).Info("Route annotation timeout exceeds maximum allowable by HAProxy, clipping to max")
-			return haproxyMaxTimeout
+			return templateutil.HaproxyMaxTimeout
+		case templateutil.InvalidInputError:
+			log.V(7).Info("Route annotation timeout removed because input is invalid: " + err.Error())
```
frobware, line 338: Use structured logging; 

```suggestion
			log.V(7).Info("Route annotation timeout removed because input is invalid", "error", err.Error())
```

frobware (changes requested):
I have concerns about the coverage we have for detecting overflow. In particular, we have zero unit test coverage for the `leadingInt` function. That function appears to initially be the core at detecting if a string of digits would exceed `int64`, irrespective of any time unit.

Additionally the `ParseHAProxyDuration` function has multiple strategies when detecting a permissible input value; first with a regexp, then the function goes on to peek at individual characters, before calling `leadingInt`. It seems that the peeking at characters can be left to `leadingInt`. And, once we've called `leadingInt`, we go on to peek at characters to determine the units for the time duration value. We should be able to consume the "remainder" that is returned (or should be) from `leadingInt` for the unit, if any.

I used `go test -coverpkg=./... -coverprofile=/tmp/profile.cov ./...` and `go tool cover -html /tmp/profile.cov` to view the code coverage for all the packages in ~openshift/router.

pkg/router/template/template_helper.go
Miciah: For that matter, we can use `log.Error` here:
```suggestion
			log.Error(err, "route annotation timeout removed because input is invalid")
```

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,130 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+	}
+	if d > 1<<63-1 {
+		return HaproxyMaxTimeoutDuration, OverflowError{errors.New("invalid duration, overflow " + orig)}
+	}
+	return time.Duration(d), nil
+}
+
+// leadingInt consumes the leading [0-9]* from s, and comes from the time package,
+// a private function used by time.ParseDuration, and needed by ParseHAProxyDuration above.
+func leadingInt[bytes []byte | string](s bytes) (x uint64, rem bytes, err error) {
```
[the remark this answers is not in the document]
frobware: I exercised the test cases in my previous comment with:

```go
func leadingInt(s string) (x uint64, rem string, err error) {
	i := 0

	for ; i < len(s); i++ {
		c := s[i]
		if c < '0' || c > '9' {
			break
		}
		if x >= 1<<63/10 {
			// Potential for overflow on next iteration
			nextVal := x*10 + uint64(c) - '0'
			if nextVal > 1<<63-1 {
				overflowErr := NewOverflowError(errors.New("value too large to be represented as duration"))
				if i == len(s)-1 {
					return 0, "", overflowErr
				}
				return 0, s[i+1:], overflowErr
			}
		}
		x = x*10 + uint64(c) - '0'
	}
	return x, s[i:], nil
}
```

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,130 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+			return HaproxyMaxTimeoutDuration, OverflowError{errors.New("invalid duration, overflow " + orig)}
+		}
+		v *= unit
+		d += v
+		if d > 1<<63 {
+			return HaproxyMaxTimeoutDuration, OverflowError{errors.New("invalid duration, overflow " + orig)}
+		}
+	}
+	if d > 1<<63-1 {
+		return HaproxyMaxTimeoutDuration, OverflowError{errors.New("invalid duration, overflow " + orig)}
```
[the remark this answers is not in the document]
candita: I can't find an edge case that hits this.  I'm looking into it.  The deepest place I can produce an error  is the line `if v > 1<<63/unit {`, which hits when I input 922337203685477581

pkg/router/template/template_helper_test.go
```diff
@@ -804,11 +805,79 @@ func TestClipHAProxyTimeoutValue(t *testing.T) {
...
+		{
+			value:    "1d12h",
+			expected: "",
+			// Invalid input produces blank output
+		},
+		{
+			value:    "25d",
+			expected: templateutil.HaproxyMaxTimeout,
+			// Exceeds the HAProxy maximum
+		},
```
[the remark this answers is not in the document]
candita: Even this smaller value overflows.  I'm not finding any number that triggers the overflow errors in `leadingInt`, see https://github.com/openshift/router/pull/483#discussion_r1302335520
```suggestion
		{
			value:    "922337203685477581ms",
			expected: templateutil.HaproxyMaxTimeout,
			// Overflow error > 1<<63/10 + 1
		},
```

pkg/router/template/util/util.go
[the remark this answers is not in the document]
candita: I'll be looking into this further.

pkg/router/template/template_helper.go
```diff
@@ -321,41 +317,38 @@ func generateHAProxyMap(name string, td templateData) []string {
...
+			return ""
+		default:
+			// This is not used at the moment
+			log.V(7).Info("Invalid route annotation timeout, setting to 5s")
```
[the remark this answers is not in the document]
candita: ok

pkg/router/template/template_helper.go
```diff
@@ -321,41 +317,38 @@ func generateHAProxyMap(name string, td templateData) []string {
...
+			return templateutil.HaproxyDefaultTimeout
 		}
 	}
+
+	// Then check to see if the timeout is larger than what HAProxy allows
+	if templateutil.HaproxyMaxTimeoutDuration < duration {
+		log.V(7).Info("Route annotation timeout exceeds maximum allowable by HAProxy, clipping to max")
```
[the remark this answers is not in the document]
candita: ok

pkg/router/template/template_helper_test.go
```diff
@@ -804,11 +805,79 @@ func TestClipHAProxyTimeoutValue(t *testing.T) {
...
+			expected: "",
+			// Invalid input produces blank output
+		},
+		{
+			value:    "1d12h",
+			expected: "",
+			// Invalid input produces blank output
+		},
+		{
+			value:    "25d",
```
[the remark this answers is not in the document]
candita: removed

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,130 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+	orig := s
+	var d uint64
+
+	// Compare the input to the HAProxy duration regex.
+	if !haproxyDurationRE.MatchString(s) {
+		return 0, InvalidInputError{errors.New("invalid duration, no pattern match " + s)}
+	}
+
+	if s == "0" {
+		return 0, nil
```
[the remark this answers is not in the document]
candita: ok

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,130 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+	}
+	for s != "" {
+		// The next character must be [0-9].
+		if !('0' <= s[0] && s[0] <= '9') {
+			return 0, InvalidInputError{errors.New("invalid duration, not a number " + orig)}
```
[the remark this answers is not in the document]
candita: Added a test for "s" as input

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,130 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+		}
+
+		// Set the default HAProxy unit of "ms", in case a unit wasn't provided.
+		u := "ms"
+		// Consume unit.
+		i := 0
+		for ; i < len(s); i++ {
+			c := s[i]
+			if '0' <= c && c <= '9' {
+				break
```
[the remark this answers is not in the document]
candita: This happens when it gets to the unit letter/s.

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,130 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+			}
+		}
+		if i > 0 {
+			u = s[:i]
+		}
+		s = s[i:]
+		unit, ok := unitMap[u]
+		if !ok {
+			return 0, InvalidInputError{errors.New("unknown unit " + u + " in duration " + orig)}
```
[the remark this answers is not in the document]
candita: This checks if the unit string is in the map, and you're right we don't need to verify that since the regexp does.   I'll remove the check to make sure the unit exists.

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,130 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+func leadingInt[bytes []byte | string](s bytes) (x uint64, rem bytes, err error) {
+	i := 0
+	for ; i < len(s); i++ {
+		c := s[i]
+		if c < '0' || c > '9' {
+			break
+		}
+		if x > 1<<63/10 {
+			// overflow.
+			return 0, rem, OverflowError{errors.New("overflow")}
```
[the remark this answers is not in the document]
candita: Added a test for 1<<63/10 + 1,  "922337203685477581ms" as input

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,130 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+		}
+		x = x*10 + uint64(c) - '0'
+		if x > 1<<63 {
+			// overflow.
+			return 0, rem, OverflowError{errors.New("overflow")}
```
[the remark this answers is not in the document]
candita: I can't find an edge case for this either.

pkg/router/template/template_helper.go
candita: fixed for all of these

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,130 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+		if err != nil {
+			// overflow.
+			return 0, OverflowError{errors.New(err.Error() + orig)}
+		}
+
+		// Set the default HAProxy unit of "ms", in case a unit wasn't provided.
+		u := "ms"
+		// Consume unit.
+		i := 0
+		for ; i < len(s); i++ {
```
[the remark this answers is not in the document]
candita: I think this is a vestige of when I wanted to allow strings like "1h30m", which we don't do anymore.  I'll remove it.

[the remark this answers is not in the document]
candita: On second thought, that wasn't an easy fix, I'll need to look into that.

pkg/router/template/util/util.go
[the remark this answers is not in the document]
candita: I copied this from go's `ParseDuration` and didn't pay attention to that.  It's not needed and I will look into removing it.

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,199 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
+
+// OverflowError represents an overflow error from ParseHAProxyDuration.
+// OverflowError is returned if the value is greater than what time.ParseDuration
+// allows (value must be representable as uint64, e.g. approximately 2562047.79 hours
+// or 9223372036854775807 nanoseconds).
+type OverflowError struct {
+	error
+}
+
+// InvalidInputError represents an error based on invalid input to ParseHAProxyDuration
```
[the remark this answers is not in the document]
candita: I think I fixed all of them except this one, will get it in an upcoming update.

candita:
> time="2023-08-23T04:08:43Z" level=info msg="added interval: Aug 23 04:06:03.434 - 1s    W ns/openshift-etcd pod/etcd-ci-op-1k3di8y9-9c2de-7g9sn-master-0 node/ci-op-1k3di8y9-9c2de-7g9sn-master-0 uid/7235a9fd-dffa-4a0d-ad2a-f689689f74ad container/etcd src/podLog dropped internal Raft message since sending buffer is full (overloaded network)" pod=etcd-ci-op-1k3di8y9-9c2de-7g9sn-master-0

/test e2e-upgrade

frobware:
I think parsing and validating a duration value can be simplified to rely on `strconv.ParseInt()` to detect for overflow.

I created `duration_parser.go`

```go
package haproxytime

import (
	"errors"
	"regexp"
	"strconv"
	"time"
)

// MaxTimeout defines the maximum duration that can be represented in
// HAProxy's configuration.
const MaxTimeout = 2147483647 * time.Millisecond

var (
	// OverflowError is returned when the parsed value exceeds the
	// maximum allowed.
	OverflowError = errors.New("value out of range")

	// SyntaxError is returned when the input string doesn't match
	// HAProxy's duration format.
	SyntaxError = errors.New("invalid duration")

	durationRE = regexp.MustCompile(`^([0-9]+)(us|ms|s|m|h|d)?$`)
)

// ParseDuration takes a string representing a duration in HAProxy's
// specific format and converts it into a time.Duration value. The
// string can include an optional unit suffix, such as "us", "ms",
// "s", "m", "h", or "d". If no suffix is provided, milliseconds are
// assumed. The function returns an OverflowError if the value exceeds
// MaxTimeout, or a SyntaxError if the input string doesn't match the
// expected format.
func ParseDuration(input string) (time.Duration, error) {
	matches := durationRE.FindStringSubmatch(input)
	if matches == nil {
		return 0, SyntaxError
	}

	// Default unit is milliseconds, unless specified.
	unit := time.Millisecond

	numericPart := matches[1]
	unitPart := ""
	if len(matches) > 2 {
		unitPart = matches[2]
	}

	switch unitPart {
	case "us":
		unit = time.Microsecond
	case "ms":
		unit = time.Millisecond
	case "s":
		unit = time.Second
	case "m":
		unit = time.Minute
	case "h":
		unit = time.Hour
	case "d":
		unit = 24 * time.Hour
	}

	value, err := strconv.ParseInt(numericPart, 10, 32)
	if err != nil {
		// ParseInt is documented to return only ErrSyntax or
		// ErrRange when an error occurs. As we've already
		// covered the ErrSyntax case with the regex, we can
		// assume this is ErrRange.
		return 0, OverflowError
	}

	duration := time.Duration(value) * unit
	if duration > MaxTimeout {
		return 0, OverflowError
	}

	return duration, nil
}
```

and some test cases: 

```go
package haproxytime_test

import (
	"testing"
	"time"

	"github.com/openshift/router/pkg/router/template/util/haproxytime"
)

func TestParseHAProxyDuration(t *testing.T) {
	tests := []struct {
		input            string
		expectedDuration time.Duration
		expectedErr      error
	}{
		// Success Cases
		{"0", 0, nil},
		{"123us", 123 * time.Microsecond, nil},
		{"456ms", 456 * time.Millisecond, nil},
		{"789s", 789 * time.Second, nil},
		{"5m", 5 * time.Minute, nil},
		{"2h", 2 * time.Hour, nil},
		{"1d", 24 * time.Hour, nil},
		{"24d", 24 * 24 * time.Hour, nil},
		{"2147483646", haproxytime.MaxTimeout - time.Millisecond, nil},
		{"2147483647", haproxytime.MaxTimeout, nil},

		// Syntax Error Cases
		{"", 0, haproxytime.SyntaxError},
		{"invalid", 0, haproxytime.SyntaxError},
		{"+100", 0, haproxytime.SyntaxError},
		{"-100", 0, haproxytime.SyntaxError},
		{"/", 0, haproxytime.SyntaxError},
		{" spaces are invalid", 0, haproxytime.SyntaxError},

		// Overflow Error Cases
		{"25d", 0, haproxytime.OverflowError},
		{"9999999999", 0, haproxytime.OverflowError},
		{"1000000000000000000000000000", 0, haproxytime.OverflowError},
		{"1000000000000000000000000000h", 0, haproxytime.OverflowError},
		{"2147483648", 0, haproxytime.OverflowError},
	}

	for _, tc := range tests {
		t.Run(tc.input, func(t *testing.T) {
			duration, err := haproxytime.ParseDuration(tc.input)
			if duration != tc.expectedDuration {
				t.Errorf("expected duration %v, got %v", tc.expectedDuration, duration)
			}
			if tc.expectedErr != nil {
				if err == nil || err.Error() != tc.expectedErr.Error() {
					t.Errorf("expected error %v, got %v", tc.expectedErr, err)
				}
			} else if err != nil {
				t.Errorf("expected no error, got %v", err)
			}
		})
	}
}
```

I plumbed these changes into the following branch https://github.com/frobware/router/tree/OCPBUGS-6958-FixClipTimeout.

I didn't [need to] change any of the existing test cases for `TestClipHAProxyTimeoutValue`.

- New parsing code: https://github.com/openshift/router/commit/0477c2186592a4c190f6e8b2174cdc4371267ef0
- Updated logic to use `haproxytime.ParseDuration`: https://github.com/openshift/router/commit/0b21f1637ca6174febe4f701217bd1fad63e2ef3

HTH

Miciah:
@frobware,
> ```go
> 	value, err := strconv.ParseInt(numericPart, 10, 32)
> ```
This can truncate large but in-range microsecond values:
```
fmt.Println(ParseDuration("2147483648us"))
=> 0s value out of range
```
On the other hand, using `value, err := strconv.ParseInt(numericPart, 10, 64)` would make it possible to overflow with large day values:
```
fmt.Println(ParseDuration("9223372036854775807d"))
=> -24h0m0s <nil>
```
I think you need 64 bits plus an overflow check:
```go
	value, err := strconv.ParseInt(numericPart, 10, 64)
	if err != nil {
		return 0, OverflowError
	}

	duration := time.Duration(value) * unit
	if time.Duration(value) > duration {
		return 0, OverflowError
	}
```
```
fmt.Println(ParseDuration("2147483648us"))
=> 35m47.483648s <nil>
fmt.Println(ParseDuration("9223372036854775807d"))
=> 0s value out of range
```
By the way, `duration > MaxTimeout` isn't technically an overflow; it's an out-of-range condition.  `clipHAProxyTimeoutValue` already checks for this out-of-range condition, so does it make sense for `ParseDuration` to check it?

frobware:
> 2147483648us

Isn't this out of range? Max is 2147483647ms, so `2147483648us / 1000 = 2147483648ms`.

frobware:
> > 2147483648us
> 
> Isn't this out of range? Max is 2147483647ms, so `2147483648us / 1000 = 2147483648ms`.

Ignore me. **micro**seconds.

frobware:
> By the way, duration > MaxTimeout isn't technically an overflow; it's an out-of-range condition

frobware:
> clipHAProxyTimeoutValue already checks for this out-of-range condition, so does it make sense for ParseDuration to check it?

I could go either way; haproxytime.ParseDuration() does the check, or it just parses the duration and something else decides whether the parsed value is too large.

frobware:
> By the way, duration > MaxTimeout isn't technically an overflow

It overflows HAProxy's permissible upper bound.

frobware:
@candita @Miciah 

I pushed some new changes to my branch based on @Miciah's observations. 

1. **[Commit 1](https://github.com/frobware/router/commit/abf9ccb3664482e33d227489efff9559f26ba26d): Refactor `haproxytime.ParseDuration` to delegate `maxTimeout` responsibility to the caller**
    - Changed the function signature of `haproxytime.ParseDuration` to remove built-in maximum timeout checking.
    - Added new test cases to exercise boundary conditions around int64 limits for various time units.

2. **[Commit 2](https://github.com/frobware/router/commit/12d65de4b345c1a7f736e275231d109891646f9a): Initialise and Validate `haproxyMaxTimeout` at Start-Up**
    - Introduced a new global variable `haproxyMaxTimeout` to store the maximum timeout, initialising it in `init()`.
    - Added safeguard to panic if the initialisation fails, ensuring the system starts with a valid configuration.
    - Modified `clipHAProxyTimeoutValue` to compare against `haproxyMaxTimeout`.

I didn't cater for distinguishing between overflow and out-of-range errors in `ParseDuration`, largely because it makes the diff smaller. But it's not clear whether we would actually switch on the distinct error types when reporting the error.

frobware:
I also added a benchmark test to evaluate the cost of using regular expressions for parsing input values. The cost is never negligible: https://github.com/frobware/router/commit/332bc14b07549c03ae5c574d0036e534df1e88b1

candita:
/label tide/merge-method-squash

@Miciah @frobware this is ready for review.  PTAL when you get a chance.

pkg/router/template/util/haproxytime/duration_parser.go
```diff
@@ -0,0 +1,75 @@
...
+
+	durationRE = regexp.MustCompile(`^([0-9]+)(us|ms|s|m|h|d)?$`)
+)
+
+// ParseDuration takes a string representing a duration in HAProxy's
+// specific format and converts it into a time.Duration value. The
+// string can include an optional unit suffix, such as "us", "ms",
+// "s", "m", "h", or "d". If no suffix is provided, milliseconds are
+// assumed. The function returns an OverflowError if the value exceeds
+// MaxTimeout, or a SyntaxError if the input string doesn't match the
```
frobware, line 28: We moved the check for MaxTimeout to the caller so we should drop the "if the value exceeds MaxTimeout".

pkg/router/template/util/haproxytime/duration_parser.go
```diff
@@ -0,0 +1,75 @@
...
+		// ParseInt is documented to return only ErrSyntax or
+		// ErrRange when an error occurs. As we've already
+		// covered the ErrSyntax case with the regex, we can
+		// assume this is ErrRange.
+		return 0, OverflowError
+	}
+
+	if value > math.MaxInt64/int64(unit) {
+		return 0, OverflowError
+	}
```
frobware, line 71:
```suggestion
	// Check for overflow conditions before multiplying 'value' by
	// 'unit'. 'value' is guaranteed to be >= 0, as ensured by the
	// preceding regular expression.
	//
	// The maximum allowable 'value' is determined by dividing
	// math.MaxInt64 by the nanosecond representation of 'unit'.
	// This prevents overflow when 'value' is later multiplied by
	// 'unit'.
	//
	// Examples:
	//  1. If the 'unit' is time.Second (1e9 ns), then the maximum
	//     'value' allowed is math.MaxInt64 / 1e9.
	//  2. If the 'unit' is 24 * time.Hour (86400e9 ns), then the
	//     maximum 'value' allowed is math.MaxInt64 / 86400e9.
	//
	// Concrete examples with actual values:
	//   - No Overflow: "106751d" as input makes 'unit' 24 *
	//     time.Hour (86400e9 ns). The check ensures 106751 <=
	//     math.MaxInt64 / 86400e9. Specifically, 106751 <=
	//     9223372036854775807 / 86400000000 (106751 <= 106751).
	//     This is the maximum 'value' for days that won't cause
	//     an overflow.
	//
	//   - Overflow: Specifying "106752d" makes 'unit' 24 *
	//     time.Hour (86400e9 ns). The check finds 106752 >
	//     math.MaxInt64 / 86400e9. Specifically, 106752 >
	//     9223372036854775807 / 86400000000 (106752 > 106751),
	//     causing an overflow.
	if value > math.MaxInt64/int64(unit) {
		return 0, OverflowError
	}
```

frobware, line 71:
```suggestion
	// Check for overflow conditions before multiplying 'value' by 'unit'.
	// 'value' is guaranteed to be >= 0, as ensured by the preceding regular expression.
	//
	// The maximum allowable 'value' is determined by dividing math.MaxInt64
	// by the nanosecond representation of 'unit'. This prevents overflow when
	// 'value' is later multiplied by 'unit'.
	//
	// Examples:
	//  1. If the 'unit' is time.Second (1e9 ns), then the maximum
	//     'value' allowed is math.MaxInt64 / 1e9.
	//  2. If the 'unit' is 24 * time.Hour (86400e9 ns), then the
	//     maximum 'value' allowed is math.MaxInt64 / 86400e9.
	//  3. If the 'unit' is time.Microsecond (1e3 ns), then the maximum
	//     'value' allowed is math.MaxInt64 / 1e3.
	//
	// Concrete examples with actual values:
	//   - No Overflow (days): "106751d" as input makes 'unit' 24 * time.Hour (86400e9 ns).
	//     The check ensures 106751 <= math.MaxInt64 / 86400e9.
	//     Specifically, 106751 <= 9223372036854775807 / 86400000000 (106751 <= 106751).
	//     This is the maximum 'value' for days that won't cause an overflow.
	//
	//   - Overflow (days): Specifying "106752d" makes 'unit' 24 * time.Hour (86400e9 ns).
	//     The check finds 106752 > math.MaxInt64 / 86400e9.
	//     Specifically, 106752 > 9223372036854775807 / 86400000000 (106752 > 106751), causing an overflow.
	//
	//   - No Overflow (us): "9223372036854775us" makes 'unit' time.Microsecond (1e3 ns).
	//     The check ensures 9223372036854775 <= math.MaxInt64 / 1e3.
	//     Specifically, 9223372036854775 <= 9223372036854775807 / 1000 (9223372036854775 <= 9223372036854775).
	//     This is the maximum 'value' for microseconds that won't cause an overflow.
	if value > math.MaxInt64/int64(unit) {
		return 0, OverflowError
	}
```

pkg/router/template/util/util.go
```diff
@@ -94,3 +103,12 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
 
 	return prefix
 }
+
+// HaproxyMaxTimeoutDuration is HaproxyMaxTimeout as a time.Duration value.
+var HaproxyMaxTimeoutDuration = func() time.Duration {
+	d, err := time.ParseDuration(HaproxyMaxTimeout)
```
frobware, line 109: Any reason to not use haproxtime.ParseDuration for consistency here?

pkg/router/template/template_helper.go
```diff
@@ -321,41 +317,38 @@ func generateHAProxyMap(name string, td templateData) []string {
...
+			return ""
+		default:
+			// This is not used at the moment
+			log.Info("invalid route annotation timeout, setting to", "default", templateutil.HaproxyDefaultTimeout)
+			return templateutil.HaproxyDefaultTimeout
 		}
 	}
+
+	// Then check to see if the timeout is larger than what HAProxy allows
+	if templateutil.HaproxyMaxTimeoutDuration < duration {
```
frobware, line 353: Why wouldn't you express this as `if duration > templateutil.HaproxyMaxTimeoutDuration`. 

Obviously both work, but `duration > some-max` seems more straightforward.

pkg/router/template/template_helper_test.go
```diff
@@ -804,12 +814,85 @@ func TestClipHAProxyTimeoutValue(t *testing.T) {
...
+		},
+		{
+			value:    "9223372036855ms",
+			expected: templateutil.HaproxyMaxTimeout,
+			// Exceeds the time.ParseDuration maximum
+		},
+		{
+			value:    "2147483647ms",
+			expected: "2147483647ms",
+		},
```
frobware, line 846:
```suggestion
		{
			value:    "2147483647ms",
			expected: "2147483647ms",
		},
		{
			value:    "2147483648ms",
			expected: templateutil.HaproxyMaxTimeout,
		},
```

Specifically test for the HAProxy maximum.

pkg/router/template/template_helper_test.go
```diff
@@ -804,12 +814,85 @@ func TestClipHAProxyTimeoutValue(t *testing.T) {
...
 		},
+		{
+			value:    "2562047.99h",
+			expected: "",
+			// Invalid input produces blank output
+		},
+		{
+			value:    "100000000000s",
+			expected: templateutil.HaproxyMaxTimeout,
+			// Exceeds the time.ParseDuration maximum
```
frobware, line 864: We should now reference `haproxytime.ParseDuration`. Ditto for the other occurrences in this test.

pkg/router/template/template_helper_test.go
```diff
@@ -804,12 +814,85 @@ func TestClipHAProxyTimeoutValue(t *testing.T) {
...
+		},
+		{
+			value:    "9999999999999999",
+			expected: templateutil.HaproxyMaxTimeout,
+			// Exceeds the time.ParseDuration maximum and has no unit
```
frobware, line 869: This exceeds the HAProxy maximum; it does not overflow `[haproxy]time.ParseDuration`.

pkg/router/template/util/haproxytime/duration_parser.go
```diff
@@ -0,0 +1,75 @@
...
+	"errors"
+	"math"
+	"regexp"
+	"strconv"
+	"time"
+)
+
+var (
+	// OverflowError represents an overflow error from ParseDuration.
+	// OverflowError is returned if the value is greater than what time.ParseDuration
```
frobware, line 13: This implementation doesn't use `time.ParseDuration`.

pkg/router/template/util/haproxytime/duration_parser.go
```diff
@@ -0,0 +1,75 @@
...
+	// allows (value must be representable as uint64, e.g. 9223372036854775807 nanoseconds).
```
frobware, line 14: Type is signed. `int64`, not `uint64`.

pkg/router/template/template_helper.go
candita: Just following the original ordering, it can be changed.

pkg/router/template/util/haproxytime/duration_parser.go
candita: ok

pkg/router/template/util/haproxytime/duration_parser.go
candita: ok

pkg/router/template/util/util.go
candita: ok

pkg/router/template/template_helper_test.go
candita: ok

pkg/router/template/template_helper_test.go
candita: ok

pkg/router/template/template_helper_test.go
candita: ok

pkg/router/template/util/haproxytime/duration_parser.go
candita: ok

pkg/router/template/util/haproxytime/duration_parser.go
candita: ok

frobware:
/lgtm

@Miciah do you want to take another look?

Miciah (commented):
Looks good for the most part.  Excellent comments and test coverage!  I'm curious about the `durationRE` regexp; I don't think it will cause a problem since we check `$timeSpecPattern` before we even call `clipHAProxyTimeoutValue`, but the fact that the regexps don't match seems (a) incorrect and (b) indicative of some misunderstanding around the requirements.  The rest of my comments are minor stylistic suggestions.

pkg/router/template/util/util.go
```diff
@@ -94,3 +104,12 @@ func GenerateBackendNamePrefix(termination routev1.TLSTerminationType) string {
...
 }
+
+// HaproxyMaxTimeoutDuration is HaproxyMaxTimeout as a time.Duration value.
+var HaproxyMaxTimeoutDuration = func() time.Duration {
+	d, err := haproxytime.ParseDuration(HaproxyMaxTimeout)
+	if err != nil {
+		panic(err)
+	}
+	return d
+}()
```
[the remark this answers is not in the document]
frobware: My concern here is that this a delayed failure. Why not parse the value in init and assign to a variable if all is OK.potentially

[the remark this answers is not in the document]
Miciah: If I understand https://go.dev/ref/spec#Package_initialization and https://go.dev/ref/spec#Program_initialization correctly, package variables (including `HaproxyMaxTimeoutDuration`) get initialized when the program starts.  Using `init` would be a slightly more verbose way to do effectively the same thing.  (In fact, using `init` would cause the initialization to happen slightly later, per the specification of package initialization that I referenced: "The entire package is initialized by assigning initial values to all its package-level variables followed by calling all `init` functions in the order they appear in the source, possibly in multiple files, as presented to the compiler.)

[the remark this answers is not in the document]
frobware: ok. Thanks for the links.

pkg/router/template/util/haproxytime/duration_parser.go
```diff
@@ -0,0 +1,104 @@
...
+var (
+	// OverflowError represents an overflow error from ParseDuration.
+	// OverflowError is returned if the input value is greater than what ParseDuration
+	// allows (value must be representable as int64, e.g. 9223372036854775807 nanoseconds).
+	OverflowError = errors.New("overflow")
+
+	// SyntaxError represents an error based on invalid input to ParseDuration.
+	SyntaxError = errors.New("invalid duration")
+
+	durationRE = regexp.MustCompile(`^([0-9]+)(us|ms|s|m|h|d)?$`)
```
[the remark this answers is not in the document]
candita: I agree, this should match the one in `$timeSpecPattern`.  I'll fix it and the unit tests.

pkg/router/template/template_helper.go
```diff
@@ -321,41 +318,43 @@ func generateHAProxyMap(name string, td templateData) []string {
...
-	} else {
-		duration, err := time.ParseDuration(val)
-		if err != nil {
-			return val
-		}
-		if maxTimeout.Milliseconds() < duration.Milliseconds() {
-			log.V(7).Info("Route annotation timeout exceeds maximum allowable by HAProxy, clipping to max")
-			return haproxyMaxTimeout
+		switch err {
+		case haproxytime.OverflowError:
```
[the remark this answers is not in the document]
candita: ok

pkg/router/template/util/haproxytime/duration_parser.go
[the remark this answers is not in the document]
candita: I will leave the anchors and the capture group: `^([1-9][0-9]*)(us|ms|s|m|h|d)?$`

pkg/router/template/template_helper.go
```diff
@@ -321,41 +318,43 @@ func generateHAProxyMap(name string, td templateData) []string {
...
+			log.Info("route annotation timeout exceeds maximum allowable format, clipping to "+templateutil.HaproxyMaxTimeout, "input", val)
+			return templateutil.HaproxyMaxTimeout
+		case haproxytime.SyntaxError:
+			log.Error(err, "route annotation timeout removed because input is invalid", "input", val)
```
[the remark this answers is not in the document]
candita: ok

pkg/router/template/template_helper.go
```diff
@@ -321,41 +318,43 @@ func generateHAProxyMap(name string, td templateData) []string {
...
 func clipHAProxyTimeoutValue(val string) string {
 	// If the empty string is passed in,
 	// simply return the empty string.
 	if len(val) == 0 {
 		return val
 	}
-	endIndex := len(val) - 1
-	maxTimeout, err := time.ParseDuration(haproxyMaxTimeout)
+
+	// First check to see if the timeout will fit into a time.Duration
```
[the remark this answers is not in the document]
candita: ok

pkg/router/template/util/haproxytime/duration_parser.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+// ParseDuration takes a string representing a duration in HAProxy's
+// specific format and converts it into a time.Duration value. The
+// string can include an optional unit suffix, such as "us", "ms",
+// "s", "m", "h", or "d". If no suffix is provided, milliseconds are
+// assumed. The function returns an OverflowError if the value exceeds
+// the maximum allowable input, or a SyntaxError if the input string
```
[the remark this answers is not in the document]
candita: ok

pkg/router/template/util/haproxytime/duration_parser.go
```diff
@@ -0,0 +1,104 @@
...
+// doesn't match the expected format.
+func ParseDuration(input string) (time.Duration, error) {
+	matches := durationRE.FindStringSubmatch(input)
+	if matches == nil {
+		return 0, SyntaxError
+	}
+
+	// Default unit is milliseconds, unless specified.
+	unit := time.Millisecond
```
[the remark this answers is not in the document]
candita: ok

pkg/router/template/util/util.go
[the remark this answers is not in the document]
candita: done

pkg/router/template/template_helper_test.go
```diff
@@ -804,11 +814,91 @@ func TestClipHAProxyTimeoutValue(t *testing.T) {
...
+		},
+		{
+			value:    "100000000000s",
+			expected: templateutil.HaproxyMaxTimeout,
+			// Exceeds the haproxytime.ParseDuration maximum
+		},
+		{
+			value:    "9999999999999999",
+			expected: templateutil.HaproxyMaxTimeout,
+			// Exceeds the HAProxy maximum and has no unit
```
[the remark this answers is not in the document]
candita: I'll update the ordering and comments on these tests.  I also changed the value "2147483648ms" to "2147483648" to test a value that exceeds HAproxy's maximum but has no unit.

These values are testing different paths.  The value "9223372036854775807" exceeds the int64 maximum and has no unit, so it is testing that case.  The test case for "9999999999999999" overflows ParseDuration, but not int64, and has no unit, so it is testing that case.

Please take a look at the changes and let me know if this makes more sense.

pkg/router/template/util/haproxytime/duration_parser_test.go
```diff
@@ -0,0 +1,97 @@
+package haproxytime_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/openshift/router/pkg/router/template/util/haproxytime"
+)
+
+func TestParseDuration(t *testing.T) {
```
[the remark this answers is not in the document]
candita: ok

pkg/router/template/util/haproxytime/duration_parser_test.go
```diff
@@ -0,0 +1,97 @@
...
+	tests := []struct {
+		input            string
+		expectedDuration time.Duration
+		expectedErr      error
+	}{
+		// Syntax error test cases.
+		{" spaces are invalid", 0, haproxytime.SyntaxError},
```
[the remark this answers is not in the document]
candita: ok

pkg/router/template/util/haproxytime/duration_parser_test.go
```diff
@@ -0,0 +1,94 @@
...
+		{"-1us", 0, haproxytime.SyntaxError},
+		{"/", 0, haproxytime.SyntaxError},
+		{"123ns", 0, haproxytime.SyntaxError},
+		{"invalid", 0, haproxytime.SyntaxError},
+
+		// Validate default unit.
+		{"1", 1 * time.Millisecond, nil},
+
+		// Small values for each unit.
+		{"1us", 1 * time.Microsecond, nil},
```
frobware, line 33: Was there a reason to drop the boundary case (i.e., 0) for each of these units?

pkg/router/template/template_helper_test.go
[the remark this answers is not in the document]
frobware: > The test case for "9999999999999999" overflows ParseDuration, but not int64, and has no unit, so it is testing that case.

The underlying type of ParseDuration and int64 are one and the same thing.

pkg/router/template/util/haproxytime/duration_parser_test.go
candita: We don't allow zero values anymore, I changed the regex to match the haproxy template regex, per https://github.com/openshift/router/pull/483#discussion_r1355909022

candita:
> error adding pod e2e-k8s-sig-apps-daemonset-upgrade-2538_ds1-fbb6g to CNI network "multus-cni-network": plugin type="multus-shim" name="multus-cni-network" failed (add): CmdAdd (shim): CNI request failed with status 400:

/test e2e-upgrade

pkg/router/template/util/haproxytime/duration_parser_test.go
frobware: Ah, ok. Thanks.

Miciah: I'm missing something here.  If the function doesn't/shouldn't allow zero values anymore, doesn't that mean we should have test cases that verify that the function gives an error result for zero values?

candita: The haproxy parsing method tested zero values, but I removed those from ParseDuration that were returning success.  I just added one back into ParseDuration that returns failure, as it should now.

frobware:
I think it's totally reasonable that ParseDuration will parse 0 without reporting a syntax error. The semantic layering is wrong. Not accepting 0 is a property of the clip function; ParseDuration should be more the generic case. Let's modify the clip function so that it first checks to see of the leading character in the input is 0, and if so, it can handle that corner case. We should revert the changes we recently made to ParseDuration's RE.

/remove-lgtm
/lgtm cancel

frobware:
/hold
I don't know the magic incantation to cancel an existing LGTM.

candita:
> I think it's totally reasonable that ParseDuration will parse 0 without reporting a syntax error. The semantic layering is wrong. Not accepting 0 is a property of the clip function; ParseDuration should be the generic case.
> 
> Proposal:
> 
> * Revise the clip function to handle the corner case when the leading character in the input is '0'.
> * Additionally, revert the recent changes to the regular expression in ParseDuration and the test cases that asserted on leading 0.

@frobware the HAProxy template does not accept 0 as the leading character.  I thought we all agreed to make the regex pattern the same here as it is in HAProxy in https://github.com/openshift/router/pull/483#discussion_r1355909022.  Are you referring to something different?

frobware:
I previously gave this PR an `/lgtm`, but I've had reservations about the behaviour of `haproxytime.ParseDuration`—especially after we started removing test cases that involved parsing `0` as a duration. I'd like to present some counter-arguments for reconsideration.

#### Counter-Argument for Allowing `0` in `haproxytime.ParseDuration`

1. **Principle of Least Astonishment**: This principle suggests that our function should behave in a way that most users expect. Allowing `0` as a leading character aligns `haproxytime.ParseDuration` more closely with Go's native `time.ParseDuration`, which is likely familiar to many users.

2. **Modularity and Single Responsibility Principle**: A function should ideally do one thing well. `haproxytime.ParseDuration` should focus solely on parsing durations. HAProxy-specific constraints or project-specific requirements, like not allowing `0`, should be managed externally. The `clip` function already alters behaviour by capping the maximum duration to a value much less than what `time.ParseDuration` supports (i.e., `int64`), making it the appropriate place to handle such edge cases.

3. **Edge Cases in `clip` Function**: Given that the constraint on `0` is specific to our use case, this makes it even more fitting to manage this edge case in the `clip` function, maintaining the modularity and focus of `haproxytime.ParseDuration`.

4. **Consistency with Standard Libraries**: Allowing `0` makes our function more in line with the behaviour of similar functions in Go's standard libraries, thereby making it more intuitive to use and easier to understand.

5. **Ease of Future Maintenance**: Making haproxytime.ParseDuration less use-case-specific and more generalised will simplify future maintenance and adaptations. Additionally, disallowing 0 within `haproxytime.ParseDuration` would create a ripple effect: if another caller is introduced later that does permit 0, this edge case would be perpetuated, complicating the codebase and hindering maintainability.

6. **Leveraging Native Libraries**: If Go's `time.ParseDuration` supported day-level granularity, we would have reused it and worked around the `0` edge case externally, which further argues for handling such adjustments outside of `haproxytime.ParseDuration`.

Based on these considerations, I propose that we allow `0` as a leading character in `haproxytime.ParseDuration` and manage any use-case-specific requirements, like the constraint on `0` when clipping, in the `clip` function.

Miciah:
I understand the argument for consistency.  However, I discussed this with @candita, and currently, all values that get parsed using `haproxytime.ParseDuration` are guarded by `$timeSpecPattern`; if a use-case arises in the future that requires `haproxytime.ParseDuration` to allow zero values for some duration value, we can update the function then.  In the interest of avoiding continued churn, let's merge the PR in its current state.  
/approve
/lgtm

PR #488, opened 2023-06-01
RFE-4087: support for the haproxy.router.openshift.io/aws-cloud-front-id annotation

samuano76:
This PR enables the support for the haproxy.router.openshift.io/aws-cloud-front-id annotation.
If a route is annotated with haproxy.router.openshift.io/aws-cloud-front-id only connections with the HTTP header X-Aws-CFID matching the annotation value are permitted.

This allows to secure routes that are behind an AWS CloudFront as described in AWS docs:
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/add-origin-custom-headers.html

See the RFE https://issues.redhat.com/browse/RFE-4087

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -519,7 +519,11 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
         {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel")) }}
   timeout tunnel  {{ $value }}
         {{- end }}
-
+        {{- with $awsCloudFrontAnnotation := index $cfg.Annotations "haproxy.router.openshift.io/aws-cloud-front-id" }}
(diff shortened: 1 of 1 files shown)


candita:
/retitle RFE-4087: support for the haproxy.router.openshift.io/aws-cloud-front-id annotation

candita:
@samuano76 I'm checking with our product management leaders on these two PRs.

candita:
@samuano76 Thank you for submitting this pull request.  Unfortunately, we will need more robust work to be completed for the design of a solution, so I placing this on hold.

I checked with our team lead and it is not, something that we can do with the custom headers feature, because the "deny" operation is not implemented by the custom headers.  So please disregard my comment in https://github.com/openshift/router/pull/488#issuecomment-1581077799

The design that we would accept would need to adhere to the following constraints:
- provide enhancement proposal for review and discussion
- do not add a new annotation, but make it part of the cluster-ingress-operator API
- discussion and review on API changes
- avoid platform specific changes

/hold

samuano76:
Hi @candita , this would mean that enhancement in comment https://github.com/openshift/router/pull/488#issuecomment-1581077799 it's not good for deny a transaction basing on a particular custom header defined in Cluster-ingress-operator API through IngressController CR ?

PR #491, opened 2023-06-01
[release-4.12] OCPBUGS-14454, OCPBUGS-14455: Handle mTLS CRLs, and fix accidental CRL duplication

rfredette:
Manual cherry-pick of #472 

---

This PR makes the router parse the certificates supplied in ROUTER_MUTUAL_TLS_AUTH_CA, download CRLs from any CRL distribution points that are found, and write them to a file that is supplied to HAProxy for mTLS certificate validation.

It also includes a fix for an issue where, when a CRL is distributed in a certificate that is not its issuer, the CRL would sometimes be missing or two copies of the CRL would be present.


frobware:
```console
        cat: /var/lib/haproxy/mtls/latest/crls.pem: No such file or directory
```

This PR depends on https://github.com/openshift/router/pull/491.

PR #498, opened 2023-07-22
OCPBUGS-16623: properly handle weight=0

freedge:
fix regression introduced with [NE-822](https://issues.redhat.com//browse/NE-822) for services with a single endpoint having weight=0 that are wrongly sent traffic to.

Tested in Kind using example/example.yaml and changing the weight.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -1318,9 +1318,13 @@ func (r *templateRouter) calculateServiceWeights(serviceUnits map[ServiceUnitKey
 	// If there is only 1 service unit, then always set the weight 1 for all the endpoints.
 	// Scaling the weight to 256 is redundant and causes haproxy to allocate more memory on startup.
 	if len(serviceUnits) == 1 {
-		for key := range serviceUnits {
+		for key, weight := range serviceUnits {
+			var newWeight int32 = 1
+			if weight == 0 {
+				newWeight = 0
+			}
 			if r.numberOfEndpoints(key, port) > 0 {
-				serviceUnitNames[key] = 1
+				serviceUnitNames[key] = newWeight
 			}
 		}
 		return serviceUnitNames
--- a/pkg/router/template/router_test.go
+++ b/pkg/router/template/router_test.go
@@ -1094,6 +1094,51 @@ func TestCalculateServiceWeights(t *testing.T) {
 			},
 			expectedWeights: map[ServiceUnitKey]int32{},
 		},
+		{
+			name:      "a single service with weight 0 with multiple endpoints",
+			routePort: "8080",
+			serviceUnits: map[ServiceUnitKey][]Endpoint{
+				suKey1: {ep1, ep2},
+			},
+			serviceWeights: map[ServiceUnitKey]int32{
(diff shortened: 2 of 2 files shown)


pkg/router/template/router.go
```diff
@@ -1318,9 +1318,13 @@ func (r *templateRouter) calculateServiceWeights(serviceUnits map[ServiceUnitKey
 	// If there is only 1 service unit, then always set the weight 1 for all the endpoints.
 	// Scaling the weight to 256 is redundant and causes haproxy to allocate more memory on startup.
 	if len(serviceUnits) == 1 {
-		for key := range serviceUnits {
+		for key, units := range serviceUnits {
```
gcs278, line 1321: So I understand why you used `units` here, because of the other existing parts of the code use `units`, but to me, it's confusing (not your fault, the code is already a bit confusing IMO). What's a unit? Looks like `units` is actually a weight here for the service.

I think we should avoid propagating confusing variable names:

```suggestion
		for key, weight := range serviceUnits {
```

pkg/router/template/router_test.go
```diff
@@ -1094,6 +1094,51 @@ func TestCalculateServiceWeights(t *testing.T) {
...
+			},
+			serviceWeights: map[ServiceUnitKey]int32{
+				suKey1: 0,
+			},
+			expectedWeights: map[ServiceUnitKey]int32{
+				suKey1: 0,
+			},
+		},
+		{
+			name:      "services with multiple units including one shut",
```
gcs278, line 1111: *nit* consistent naming pattern with other unit tests and not sure what a `unit` here is.

```suggestion
			name:      "services with one of weight 0",
```

pkg/router/template/router_test.go
```diff
@@ -1094,6 +1094,51 @@ func TestCalculateServiceWeights(t *testing.T) {
 			},
 			expectedWeights: map[ServiceUnitKey]int32{},
 		},
+		{
+			name:      "a single service with one endpoint with weight 0",
```
gcs278, line 1098: *nit* Endpoints don't have weights, services do.

```suggestion
			name:      "a single service with weight 0 with multiple endpoints",
```

pkg/router/template/router_test.go
```diff
@@ -1094,6 +1094,51 @@ func TestCalculateServiceWeights(t *testing.T) {
...
+				suKey1: 50,
+				suKey2: 0,
+			},
+			expectedWeights: map[ServiceUnitKey]int32{
+				suKey1: 256,
+				suKey2: 0,
+			},
+		},
+		{
+			name:      "services with multiple units all shut",
```
gcs278, line 1127:
```suggestion
			name:      "services with all weight 0",
```

gcs278 (commented):
Thanks for the PR and bug. I'm surprised we missed this use case and unit testing for weight=0 .

freedge:
thanks for the review @gcs278  is it fine if I close this pull request and let you work on a proper one? (for better efficiency!) Thanks!

gcs278:
@freedge Sure, I can take this over, no problem, thanks!

PR #499, opened 2023-07-25
OCPBUGS-16623: properly handle weight=0

gcs278:
Fix regression introduced with [NE-822](https://issues.redhat.com//browse/NE-822) for services with a single endpoint having weight=0 that are wrongly sent traffic to.
    
`pkg/router/template/router.go`: Fix logic for single endpoint services to use weight=0
`pkg/router/template/router_test.go`: Add unit tests for weight=0

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -1315,12 +1315,17 @@ func (r *templateRouter) getActiveEndpoints(serviceUnits map[ServiceUnitKey]int3
 func (r *templateRouter) calculateServiceWeights(serviceUnits map[ServiceUnitKey]int32, port string) map[ServiceUnitKey]int32 {
 	serviceUnitNames := make(map[ServiceUnitKey]int32)
 
-	// If there is only 1 service unit, then always set the weight 1 for all the endpoints.
+	// If there is only 1 service unit, then always set the weight 1
+	// for all the endpoints, except when the service weight is 0.
 	// Scaling the weight to 256 is redundant and causes haproxy to allocate more memory on startup.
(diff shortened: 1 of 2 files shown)


gcs278:
Replaces https://github.com/openshift/router/pull/498

pkg/router/template/router.go
```diff
@@ -1315,12 +1315,17 @@ func (r *templateRouter) getActiveEndpoints(serviceUnits map[ServiceUnitKey]int3
...
+			var newWeight int32 = 1
+			if weight == 0 {
+				newWeight = 0
+			}
 			if r.numberOfEndpoints(key, port) > 0 {
-				serviceUnitNames[key] = 1
+				serviceUnitNames[key] = newWeight
 			}
 		}
 		return serviceUnitNames
```
frobware, line 1331:
```suggestion
	if len(serviceUnits) == 1 {
		for key, weight := range serviceUnits {
			if r.numberOfEndpoints(key, port) > 0 {
				if weight == 0 {
					serviceUnitNames[key] = 0
				} else {
					serviceUnitNames[key] = 1
				}
			}
		}
		return serviceUnitNames
	}
```

frobware (commented):
LGTM, just a suggestion to drop the temporary when iterating over the service units.

pkg/router/template/router.go
gcs278: Thanks, I think this is clearer, done.

gcs278:
I don't think this will work, but curious if I can get at least get a Jira clone:
/cherry-pick release-4.12

PR #508, opened 2023-08-22
OCPBUGS-16707: UniqueHost: Fix incorrect identification of conflicting route

frobware:
When a route is rejected due to the host already being claimed, it was
always selecting the first route but the first route is not guaranteed
to being the conflicting route.

All routes for the host are retrieved using the 'RoutesForHost'
function and that function returns a slice of routes derived from a
map in the 'hostIndex' structure. Since maps do not guarantee order,
the first route was not necessarily the oldest.

The fix is to first sort the conflicting routes based on their
submission time, and then pick the newest.

Fixes: https://issues.redhat.com/browse/OCPBUGS-16707

### DIFF
--- a/pkg/router/controller/unique_host.go
+++ b/pkg/router/controller/unique_host.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	kapi "k8s.io/api/core/v1"
@@ -15,6 +16,7 @@ import (
 
 	"github.com/openshift/router/pkg/router"
 	"github.com/openshift/router/pkg/router/controller/hostindex"
+	"github.com/openshift/router/pkg/router/routeapihelpers"
 )
 
 // RouteHostFunc returns a host for a route. It may return an empty string.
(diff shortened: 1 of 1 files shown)


frobware:
Reproducer: https://github.com/frobware/haproxy-hacks/tree/master/OCPBUGS-16707

frobware:
There are no nodes that your pod can schedule to - check your requests, tolerations, and node selectors (0/16 nodes are available

/test all

frobware:
Too many days, ...
/retest

pkg/router/controller/unique_host.go
```diff
@@ -179,6 +181,9 @@ func (p *UniqueHost) HandleRoute(eventType watch.EventType, route *routev1.Route
 			// we were not added because another route is covering us
 			var owner *routev1.Route
 			if old, ok := p.index.RoutesForHost(host); ok && len(old) > 0 {
+				sort.SliceStable(old, func(i, j int) bool {
+					return !routeapihelpers.RouteLessThan(old[i], old[j])
+				})
```
rfredette, line 186: I'm not entirely sure this fixes the issue. `old` here seems to be all the routes with the same host, but I think we also need to take path into account to decide which route conflicts. Sorting all the routes works for [the reproducer script from your haproxy-hacks repo](https://github.com/frobware/haproxy-hacks/blob/master/OCPBUGS-16707/reproducer/repro.sh), but if I edit the script to create the routes in the order route2 > route1 > route3, then route3 reports that route1 is the conflicting route, even though it's actually route2 that's the problem:
```
route.route.openshift.io/route2 created
route.route.openshift.io/route1 created
route.route.openshift.io/route3 created
Name:                   route3
Namespace:              ocpbugs16707
Created:                3 seconds ago
Labels:                 <none>
Annotations:            <none>
Requested Host:         httpd-example-path-based-routes.apps.firstcluster.lab.upshift.rdu2.redhat.com
                        rejected by router default:  (host router-default.apps.ci-ln-xxi6972-72292.gcp-2.ci.openshift.org)HostAlreadyClaimed (3 seconds ago)
                          route route1 already exposes httpd-example-path-based-routes.apps.firstcluster.lab.upshift.rdu2.redhat.com and is older
Path:                   /api
TLS Termination:        <none>
Insecure Policy:        <none>
Endpoint Port:          http

Service:        api-service
Weight:         100 (100%)
Endpoints:      <error: endpoints "api-service" not found>
```

frobware: I pushed an update in https://github.com/openshift/router/commit/ca357ed0bd537bc7f01ce71e46a6b50f6f2b18a8.

PR #514, opened 2023-09-25
[WIP] OCPBUGS-1689: Router clear admitted status on label changes

gcs278:
The router clears route admitted status for routes no longer selected by route or namespace labels. This is achieved by modifying the handling of `Delete` events across the plugins. It also introduces the use of the `Error` event to identify rejected routes, instead of relying solely on the `Delete` event. Additionally, the Status plugin now utilizes these redefined Delete events to clear route statuses.

This is a simplified (and arguably more elegant) redesign of https://github.com/openshift/cluster-ingress-operator/pull/849, which added a new control loop in the ingress operator. This control loop watched every route and namespace update, attempting to detect label changes. If a namespace label changed, it would need to reconcile every route in that namespace.


gcs278:
/hold
unit test is failing

gcs278:
Needs rebase and understanding the unit test failure

pkg/router/controller/status.go
```diff
@@ -167,16 +183,22 @@ func performIngressConditionUpdate(action string, lease writerlease.Lease, track
...
 // a pointer to the current ingress record.
-func recordIngressCondition(route *routev1.Route, name, hostName string, condition routev1.RouteIngressCondition) (changed, created bool, at time.Time, latest, original *routev1.RouteIngress) {
+func recordIngressCondition(route *routev1.Route, name, hostName string, condition *routev1.RouteIngressCondition) (changed, created bool, at time.Time, latest, original *routev1.RouteIngress) {
 	for i := range route.Status.Ingress {
 		existing := &route.Status.Ingress[i]
 		if existing.RouterName != name {
 			continue
 		}
 
+		// If condition is nil, remove the ingress condition.
```
gcs278, line 196: This should be targeting a specific condition, not removing the entire ingress condition, unless perhaps there isn't any conditions left.

gcs278: I'll take that back. When a router unadmits a route, we should just clear the whole status.ingress object. So this is mostly fine.

PR #534, opened 2023-11-03
OCPBUGS-22739: Properly handle rewrite-target annotation

gcs278:
Process the `haproxy.router.openshift.io/rewrite-target` annotation value to prevent HaProxy configuration issues while maintaining API compatibility. Properly handle values with spaces, backslashes, and other special characters.

Add a new package, rewritetarget, to organize the functions that process the rewrite target annotation. Add a separate unit test package, rewritetarget_test, to enforce black-box testing principles.

- `images/router/haproxy/conf/haproxy-config.template`: Call processRewriteTarget function in template.
- `pkg/router/template/rewritetarget/rewritetarget.go`: New package with functions to process the rewrite target annotation.
- `pkg/router/template/rewritetarget/rewritetarget_test.go`: Unit test SanitizeInput function.
- `pkg/router/template/template_helper.go`: Expose SanitizeInput to as a Go Template function.

### DIFF
--- a/pkg/router/template/template_helper.go
+++ b/pkg/router/template/template_helper.go
@@ -19,6 +19,7 @@ import (
 	templateutil "github.com/openshift/router/pkg/router/template/util"
 	haproxyutil "github.com/openshift/router/pkg/router/template/util/haproxy"
 	"github.com/openshift/router/pkg/router/template/util/haproxytime"
+	"github.com/openshift/router/pkg/router/template/util/rewritetarget"
 )
 
 const (
@@ -411,4 +412,6 @@ var helperFunctions = template.FuncMap{
 
 	"clipHAProxyTimeoutValue": clipHAProxyTimeoutValue, //clips extrodinarily high timeout values to be below the maximum allowed timeout value
 	"parseIPList":             parseIPList,             //parses the list of IPs/CIDRs (IPv4/IPv6)
+
+	"processRewriteTarget": rewritetarget.SanitizeInput, //sanitizes `haproxy.router.openshift.io/rewrite-target` annotation
 }
--- a/pkg/router/template/util/rewritetarget/rewritetarget.go
+++ b/pkg/router/template/util/rewritetarget/rewritetarget.go
@@ -1,0 +1,159 @@
+package rewritetarget
+
+import (
+	"regexp"
+	"strings"
+)
+
+var (
+	// oneOrMorePercentSigns matches one or more literal percent
+	// signs.
+	oneOrMorePercentSigns = regexp.MustCompile(`%+`)
+
+	// oneOrMoreBackslashes matches one or more literal
+	// backslashes.
+	oneOrMoreBackslashes = regexp.MustCompile(`\\+`)
+)
+
+type runeResult struct {
+	value   string // The rune to append to the result.
+	escaped bool   // Whether the next rune is escaped.
+	skip    bool   // Whether to skip appending this rune.
+	stop    bool   // Whether to stop processing further runes.
+}
+
+type processFunc func(char rune, escaped bool) runeResult
+
+// processRunes iterates through the input string and calls processFn
+// for each rune. It uses the result of processFn to determine whether
+// to continue, skip the rune, or stop processing the input. It returns
+// a processed string.
+func processRunes(input string, processFn processFunc) string {
+	var sb strings.Builder
+
+	// Increase buffer capacity to improve performance by
+	// preallocating space for the string being built. The buffer
+	// is set to 125% of the input length to accommodate
+	// additional characters without the need for further
+	// allocation.
+	sb.Grow(len(input) + (len(input) / 4))
+
+	// escapeFlag tracks if the next rune is escaped.
+	escapeFlag := false
+
+	for _, char := range input {
+		result := processFn(char, escapeFlag)
+		escapeFlag = result.escaped
+		if result.stop {
+			break
+		}
+		if !result.skip {
+			sb.WriteString(result.value)
+		}
+	}
+
+	return sb.String()
+}
+
+// replacePercentSigns replaces each % to with %% if the current sequence
+// contains an odd number %. Previously, a single % produced a syntax error,
+// and users had to use %% to represent a literal %. To ensure compatibility
+// while resolving this problem, we must continue to replace single % with
+// double %%, but leave double %%  as is, so that existing users who manually
+// addressed the issue are not broken. Furthermore, even sequences, such as
+// %%%% or %%%%%% were also valid and should be left untouched, while odd
+// sequences such as %%% or %%%%% were invalid, and each % should be doubled
+// to ensure every % is escaped.
+func replacePercentSigns(val string) string {
+	return oneOrMorePercentSigns.ReplaceAllStringFunc(val, func(match string) string {
+		if len(match)%2 == 1 {
+			// For an odd count of %, replace each % with %%.
+			return strings.ReplaceAll(match, "%", "%%")
+		}
+		return match
+	})
+}
+
+// removeSingleBackslashes removes single backslashes \.
+func removeSingleBackslashes(val string) string {
+	return oneOrMoreBackslashes.ReplaceAllStringFunc(val, func(match string) string {
+		if len(match) == 1 {
+			return oneOrMoreBackslashes.ReplaceAllString(match, "")
+		}
+		return match
+	})
+}
+
+// convertDoubleBackslashes converts double backslashes \\ to single
+// backslashes \.
+func convertDoubleBackslashes(val string) string {
+	return strings.ReplaceAll(val, `\\`, `\`)
+}
+
+// newProcessHashCreator creates a function to process '#' runes in a
+// string. It returns a processFunc which will set
+// encounteredCommentMarker to true and stop processing if it
+// encounters a '#' that is not escaped.
+func newProcessHashCreator(encounteredCommentMarker *bool) processFunc {
+	return func(char rune, escaped bool) runeResult {
+		if char == '#' && !escaped {
+			*encounteredCommentMarker = true
+			return runeResult{stop: true}
+		}
+		return runeResult{value: string(char), escaped: char == '\\' && !escaped}
+	}
+}
+
+// processDoubleQuotes is a processFunc that processes double quote
+// characters. It skips the double quote if it's not escaped.
+func processDoubleQuotes(char rune, escaped bool) runeResult {
+	if char == '"' && !escaped {
+		return runeResult{skip: true}
+	}
+	return runeResult{value: string(char), escaped: char == '\\' && !escaped}
+}
+
+// processSingleQuotes is a processFunc that processes single quote
+// characters. It skips the single quote if it's not escaped. If it's
+// escaped, it adds an escaped single quote for later conversion.
+func processSingleQuotes(char rune, escaped bool) runeResult {
+	if char == '\'' && !escaped {
+		return runeResult{skip: true}
+	} else if char == '\'' && escaped {
+		// TRICKY: Because our later operation removes single \,
+		// we should double escape \\ it, so it will be converted
+		// to \ later.
+		return runeResult{value: `'\\''`}
+	}
+	return runeResult{value: string(char), escaped: char == '\\' && !escaped}
+}
+
+// SanitizeInput processes the `haproxy.router.openshift.io/rewrite-target`
+// annotation value for API compatibility while properly handling values
+// with spaces, backslashes, and other special characters. Because the
+// annotation value was initially introduced without being enclosed in
+// single quotes, certain characters were interpreted rather than being
+// treated literally. We later added the single quotes wrapping to resolve
+// OCPBUGS-22739. However, we must still maintain API compatibility after
+// this change: the annotation values MUST be interpreted to the same values
+// after updating to enclose the value in single quotes.
+func SanitizeInput(val string) string {
+	var encounteredCommentMarker bool
+
+	val = processRunes(val, newProcessHashCreator(&encounteredCommentMarker))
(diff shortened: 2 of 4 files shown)


pkg/router/template/template_helper.go
```diff
@@ -386,6 +386,115 @@ func parseIPList(list string) string {
...
+
+	// Convert percent sign % to become double percent signs %% if the sequence contains an odd number of
+	// percent signs.
+	// In the past, we didn't enclose the rewrite target value in single quotes to ensure haproxy interprets the values
+	// literally. As a result, a single % produced a syntax error, and users had to use %% to represent a literal %. To
+	// ensure compatibility while resolving this problem, we must continue to replace single % with double %%, but
+	// leave double %% as is, so that existing users who manually addressed the issue are not broken. Furthermore, even
+	// sequences, such as %%%% or %%%%%% were valid and should be left untouched, while odd sequences such as %%% or
+	// %%%%% were invalid, and should be doubled to ensure every % is escaped.
+	re := regexp.MustCompile(`%+`)
```
frobware, line 466: If this function will be called frequently then we should be somewhat wary of the time and cost to compile the RE. See: https://github.com/openshift/router/pull/268

gcs278: Nice catch. Might as well optimize it. Done.

Miciah: Using `cachedRegexpCompile` makes sense where the pattern is built at run-time, but in this case, you can just use a package variable to avoid the map lookup and error checking:
```go
var (
	// oneOrMorePercentSigns matches one or more literal percent signs.
	oneOrMorePercentSigns = regexp.MustCompile(`%+`)
)

Comment (gcs278) [reply pkg/router/template/template_helper.go]: Oh I see the intention of `cachedRegexpComplie` now. Done.

**Comment by frobware** [comment]
The changes LGTM but I think somebody else should also review otherwise I'd just be rubber-stamping my own suggestions.

**Comment by gcs278** [comment]
Image build failed again.
/test e2e-agnostic

**Comment by gcs278** [comment]
testplatform refreshed the RPM repo:
/test e2e-agnostic

**Comment by ShudiLi** [comment]
tested it with 4.15.0-0.ci.test-2023-11-14-012251-ci-ln-8wxgkwb-latest
`
1.
% oc get clusterversion
version   4.15.0-0.ci.test-2023-11-14-012251-ci-ln-8wxgkwb-latest   True        False         38m     Cluster version is 4.15.0-0.ci.test-2023-11-14-012251-ci-ln-8wxgkwb-latest

2.
% oc -n openshift-ingress rsh router-default-9d6f95d4c-9tmc
sh-4.4$ grep "http-request replace-path" haproxy-config.template
  http-request replace-path ^{{ $cfg.Path }}/?(.*)$ '{{ processRewriteTarget $pathRewriteTarget }}'
  http-request replace-path ^{{ $cfg.Path }}(.*)$ '{{ processRewriteTarget $pathRewriteTarget }}'
sh-4.4$

3.
% oc annotate route/service-unsecure haproxy.router.openshift.io/rewrite-target="/t t$%&*(){}[].,z#:\`_-|" --overwrite
route.route.openshift.io/service-unsecure annotate

4.
% oc get route service-unsecure -o=jsonpath="{.metadata.annotations.haproxy\.router\.openshift\.io/rewrite-target}"
/t t$%&*(){}[].,z#:`_-|
`

**Comment by ShudiLi** [comment]
/label qe-approved
thanks

**Comment by gcs278** [comment]
waiting on reviews
/remove-lifecycle stale

**Comment by gcs278** [comment]
I've asked @alebedev87 to take a look to help move this PR along (since Andy has a conflict of interest)

/assign @alebedev87

File: pkg/router/template/util/rewritetarget/rewritetarget_test.go [hunk 2]
```diff
@@ -0,0 +1,94 @@
...
+			output: `%%foo\1`,
+		},
+		{
+			name:   "double percent should be not changed",
+			input:  `%%foo`,
+			output: `%%foo\1`,
+		},
+		{
+			name:   "triple percent should be doubled entirely",
+			input:  `%%%foo`,
```
alebedev87, line 26: So here we think that the user put "3 percent signs without escaping" and not "1 escaped percent sign plus one unescaped", right?

pkg/router/template/util/rewritetarget/rewritetarget.go
```diff
@@ -0,0 +1,156 @@
...
+
+	val = processRunes(val, processHashCreator(&encounteredCommentMarker))
+	val = processRunes(val, processDoubleQuotes)
+	val = processRunes(val, processSingleQuotes)
+	val = replacePercentSigns(val)
+	val = removeSingleBackslashes(val)
+	val = convertDoubleBackslashes(val)
+
+	if !encounteredCommentMarker {
+		// Add the capture group.
```
alebedev87, line 151: May be a little more info of why we are adding the capture group.

alebedev87 (commented):
Overall, looks fine to me, just some small comments and questions.

pkg/router/template/util/rewritetarget/rewritetarget_test.go
```diff
@@ -0,0 +1,94 @@
+package rewritetarget_test
+
+import (
+	"github.com/openshift/router/pkg/router/template/util/rewritetarget"
```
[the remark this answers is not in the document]
frobware: It's not the same package. The tests are in their own package. i.e., `rewritetarget_test`.

pkg/router/template/util/rewritetarget/rewritetarget.go
```diff
@@ -0,0 +1,156 @@
...
+// after updating to enclose the value in single quotes.
+func SanitizeInput(val string) string {
+	var encounteredCommentMarker bool
+
+	val = processRunes(val, processHashCreator(&encounteredCommentMarker))
+	val = processRunes(val, processDoubleQuotes)
+	val = processRunes(val, processSingleQuotes)
+	val = replacePercentSigns(val)
+	val = removeSingleBackslashes(val)
+	val = convertDoubleBackslashes(val)
```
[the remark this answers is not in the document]
frobware: Given that there's differing function signatures you would have to do something like:

```go
func SanitizeInput(val string) string {
	type processStep func(string) string

	var encounteredCommentMarker bool

	fns := []processStep{
		func(val string) string { return processRunes(val, processHashCreator(&encounteredCommentMarker)) },
		func(val string) string { return processRunes(val, processDoubleQuotes) },
		func(val string) string { return processRunes(val, processSingleQuotes) },
		replacePercentSigns,
		removeSingleBackslashes,
		convertDoubleBackslashes,
	}

	for _, fn := range fns {
		val = fn(val)
	}

	if !encounteredCommentMarker {
		// The literal `\1` is appended to annotations without
		// comments to comply with HAProxy rewrite rule
		// expectations, which necessitate a capture group
		// reference (i.e., `\1`) for dynamic substitutions.
		val += `\1`
	}

	return val
}
```

Is this actually any better than just calling known functions names one by one?

pkg/router/template/util/rewritetarget/rewritetarget.go
frobware:
```go
		// The literal `\1` is appended to annotations without
		// comments to comply with HAProxy rewrite rule
		// expectations, which necessitate a capture group
		// reference (i.e., `\1`) for dynamic substitutions.
```

Any better?

pkg/router/template/util/rewritetarget/rewritetarget_test.go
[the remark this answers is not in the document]
alebedev87: Ah, right, I was looking at the directories. So, what's the idea behind? To test without the access to any private variables? Like if it was the user of the package?

I see this pattern is not quite uniformly implemented in `util` package (e.g. [haproxy](https://github.com/openshift/router/blob/4933f71131a3f4fc9f4ca0d93449569e315b74b2/pkg/router/template/util/haproxy/map_entry_test.go#L1)).

pkg/router/template/util/rewritetarget/rewritetarget.go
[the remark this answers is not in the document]
alebedev87: > Is this actually any better than just calling known functions names one by one?

I find it slightly more structured. But I agree that this may be subjective. So, I'm fine to drop it.

pkg/router/template/util/rewritetarget/rewritetarget_test.go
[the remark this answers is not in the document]
alebedev87: >     Encapsulation: It guarantees tests can only access exported package members, mirroring how external packages interact with rewritetarget. This approach reinforces the integrity of the package's encapsulation.
> 
>     API Design: Black-box testing inherently leads to clearer and more stable APIs by focusing on the package's usability rather than its internals.
> 
>     Resilience: Tests are less prone to break due to internal changes, as they're tied to the public API. This helps to reduce maintenance.
> 
>     User Perspective: It provides insights into how the package is used, highlighting potential areas for improvement from the end-user’s standpoint.

I see. Yes, I think it makes sense for `SanitizeInput` function.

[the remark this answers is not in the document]
gcs278: Right, Andy and I had this discussion in slack. It's an interesting perspective we should consider when writing unit tests.

pkg/router/template/util/rewritetarget/rewritetarget.go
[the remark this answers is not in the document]
gcs278: Sounds good, leaving as is.

pkg/router/template/util/rewritetarget/rewritetarget.go
gcs278: Done. Thanks Andy, you are making this too easy for me :smile:

pkg/router/template/util/rewritetarget/rewritetarget.go
```diff
@@ -0,0 +1,156 @@
...
+// backslashes \.
+func convertDoubleBackslashes(val string) string {
+	return strings.ReplaceAll(val, `\\`, `\`)
+}
+
+// processHashCreator creates a function to process '#' runes in a
+// string. It returns a processFunc which will set
+// encounteredCommentMarker to true and stop processing if it
+// encounters a '#' that is not escaped.
+func processHashCreator(encounteredCommentMarker *bool) processFunc {
```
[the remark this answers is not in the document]
gcs278: Yea that's fair. Done.

pkg/router/template/util/rewritetarget/rewritetarget.go
```diff
@@ -0,0 +1,156 @@
...
+			*encounteredCommentMarker = true
+			return runeResult{stop: true}
+		}
+		return runeResult{value: string(char), escaped: char == '\\' && !escaped}
+	}
+}
+
+// processDoubleQuotes is a processFunc that processes double quote
+// characters. It skips the double quote if it's not escaped.
+var processDoubleQuotes = func(char rune, escaped bool) runeResult {
```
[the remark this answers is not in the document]
gcs278: Yea I don't see any reason these need to be a var, I'll change.

gcs278:
@alebedev87 @frobware Code review changes from Andrey's comments. I also revised the git commit message.

pkg/router/template/util/rewritetarget/rewritetarget_test.go
alebedev87: Another interpretation of `%%%` input could be as a combination of `%%` + `%` inputs. Therefore, by default, we assume the user's intent is to do the right escaping (like in case of `%%` input) and we do the minimum necessary to ensure the correct syntax(== `%%%%`).

ShudiLi:
tested it with 4.15.0-0.ci.test-2024-03-14-121356-ci-ln-1wy2cjb-latest

1.
% oc get clusterversion
version   4.15.0-0.ci.test-2024-03-14-121356-ci-ln-1wy2cjb-latest   True        False         5m44s   Cluster version is 4.15.0-0.ci.test-2024-03-14-121356-ci-ln-1wy2cjb-latest

2.
% oc -n openshift-ingress rsh router-default-567f677657-7gr4q 
sh-5.1$  grep "http-request replace-path" haproxy-config.template
  http-request replace-path ^{{ $cfg.Path }}/?(.*)$ '{{ processRewriteTarget $pathRewriteTarget }}'
  http-request replace-path ^{{ $cfg.Path }}(.*)$ '{{ processRewriteTarget $pathRewriteTarget }}'
sh-5.1$

3.
% oc annotate route/service-unsecure haproxy.router.openshift.io/rewrite-target="/t t$%&*(){}[].,z#:\`_-|" --overwrite
route.route.openshift.io/service-unsecure annotate

4.
% oc get route service-unsecure -o=jsonpath="{.metadata.annotations.haproxy\.router\.openshift\.io/rewrite-target}" 
/t t$%&*(){}[].,z#:`_-|%

ShudiLi:
/label qe-approved
thanks

pkg/router/template/util/rewritetarget/rewritetarget_test.go
gcs278: Ah yes it's all coming back to me now. You are correct, we could assume mixed intention for escaped & unescaped, taking each `%` by itself. However, the doubling `%%%` to `%%%%%%` then gets interpreted back to `%%%` again, which is exactly what the author specified in the first place, keeping some consistency.

gcs278:
/hold

I want to give @Miciah a chance to look at this, as he helped me with the solution. We are in `acknowledge-critical-fixes-only`  anyways, so there isn't a rush until after shift week.

I'll unhold after shift week if I don't hear any issues.

gcs278:
Moving forward with this fix. I'm going to look into getting docs to help me with a doc update.  

/unhold

PR #536, opened 2023-11-07
OCPBUGS-14914: Set tunnel and server timeouts at backend level 

alebedev87:
This PR removes the tunnel and server timeouts in the global `default` section and sets them on all the backends. The route annotations for the timeouts continue to work as they did before.

[As suggested in the upstream issue](https://github.com/haproxy/haproxy/issues/2280#issuecomment-1711648863), the middle backends (`be_sni` and `be_no_sni`) are set with the maximum of all the route timeouts to avoid the following warning message:
```
I1130 10:49:26.557528       1 router.go:649] template "msg"="router reloaded" "output"="[NOTICE]   (18) : haproxy version is 2.6.13-234aa6d\n[NOTICE]   (18) : path to executable is /usr/sbin/haproxy\n[WARNING]  (18) : config : missing timeouts for backend 'be_sni'.\n   | While not properly invalid, you will certainly encounter various problems\n   | with such a configuration. To fix this, please ensure that all following\n   | timeouts are set to a non-zero value: 'client', 'connect', 'server'.\n[WARNING]  (18) : config : missing timeouts for backend 'be_no_sni'.\n   | While not properly invalid, you will certainly encounter various problems\n   | with such a configuration. To fix this, please ensure that all following\n   | timeouts are set to a non-zero value: 'client', 'connect', 'server'.\n
```

Test using [haproxy-timeout-tunnel](https://github.com/alebedev87/haproxy-timeout-tunnel) repository:
```
$ oc -n openshift-ingress get deploy router-default -o yaml | yq .spec.template.spec.containers[0].image 
quay.io/alebedev/openshift-router:12.4.110

$ oc -n openshift-ingress get deploy router-default -o yaml | yq .spec.template.spec.containers[0].env | grep -A1 TUNNEL
- name: ROUTER_DEFAULT_TUNNEL_TIMEOUT
  value: 5s

$ oc -n tunnel-timeout get route hello-openshift -o yaml | yq .metadata.annotations
haproxy.router.openshift.io/timeout-tunnel: 15s

$ ROUTE_HOST=$(oc -n tunnel-timeout get route hello-openshift --template='{{.spec.host}}')
$ time HOST=${ROUTE_HOST} PORT=443 SCHEME=wss ./websocket-cli.js
WebSocket Client Connected
echo-protocol Connection Closed

real	0m15.721s
user	0m0.126s
sys	0m0.021s

$ oc -n openshift-ingress exec -ti deploy/router-default -- cat /var/lib/haproxy/conf/haproxy.config | grep 'timeout tunnel '
  timeout tunnel  5s
  timeout tunnel  5s
  timeout tunnel  5s
  timeout tunnel  5s
  timeout tunnel  5s
  timeout tunnel  5s
  timeout tunnel  5s
  timeout tunnel  5s
  timeout tunnel  15s
```

The e2e test: https://github.com/openshift/cluster-ingress-operator/pull/1013.
Presentation which summarizes the issue: [link](https://docs.google.com/presentation/d/1vU1CH20lzFWDs8Z_iNaWBhPwdKM2iyS-Qb7zlMD9XTc) (Red Hat only).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,13 @@ GO_LDFLAGS ?=-ldflags "-s -w $(call version-ldflags,$(PACKAGE)/pkg/version) $(GO
 GO=GO111MODULE=on GOFLAGS=-mod=vendor go
 GO_BUILD_RECIPE=CGO_ENABLED=1 $(GO) build -o $(BIN) $(GO_GCFLAGS) $(GO_LDFLAGS) $(MAIN_PACKAGE)
 
+# Get all the packages with benchmark tests defined to prevent
+# "go test" from entering all existing packages when "./.." is used.
+BENCH_PKGS ?= $(shell \grep -lR '^func Benchmark' | xargs -I {} dirname {} | sed 's|^|./|' | paste -sd ' ')
+BENCH_TEST ?= .
+BENCH_TIME ?= 2s
+BENCH_COUNT ?= 5
+
 all: build
 
 build:
@@ -38,6 +45,11 @@ images/router/*/Dockerfile.rhel: images/router/base/Dockerfile.rhel
 
 check:
 	CGO_ENABLED=1 $(GO) test -race ./...
+
+.PHONY: bench
+bench:
+	@# -run flag is added to avoid running unit tests
+	CGO_ENABLED=1 $(GO) test -bench=$(BENCH_TEST) -run='^#' -benchtime=$(BENCH_TIME) -count=$(BENCH_COUNT) -benchmem $(BENCH_PKGS)
 
 .PHONY: verify
 verify:
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -9,6 +9,8 @@
 {{- $dynamicConfigManager := .DynamicConfigManager }}
 {{- $router_ip_v4_v6_mode := env "ROUTER_IP_V4_V6_MODE" "v4" }}
 {{- $router_disable_http2 := env "ROUTER_DISABLE_HTTP2" "false" }}
+{{- $routerDefaultServerTimeout := env "ROUTER_DEFAULT_SERVER_TIMEOUT" "30s" }}
+{{- $routerDefaultTunnelTimeout := env "ROUTER_DEFAULT_TUNNEL_TIMEOUT" "1h" }}
 {{- $haveClientCA := .HaveClientCA }}
 {{- $haveCRLs := .HaveCRLs }}
 
@@ -41,6 +43,9 @@
 
 {{- /* pathRewriteTargetPattern: Match path rewrite-Target */}}
 {{- $pathRewriteTargetPattern := `^/.*$` -}}
+
+{{- /* Maximum timeout among all the routes, required to be set on the middle backends to avoid warning message about missing server timeout.  */}}
+{{- $routerMaxServerTimeout := maxTimeoutFirstMatchedAndClipped .State "haproxy.router.openshift.io/timeout" $timeSpecPattern $routerDefaultServerTimeout }}
 
 global
   # Drop resource limit checks to mitigate https://issues.redhat.com/browse/OCPBUGS-21803 in HAProxy 2.6.
@@ -158,13 +163,9 @@ defaults
   timeout connect {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_CONNECT_TIMEOUT") "5s" }}
   timeout client {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_CLIENT_TIMEOUT") "30s" }}
   timeout client-fin {{ firstMatch $timeSpecPattern (env "ROUTER_CLIENT_FIN_TIMEOUT") "1s" }}
-  timeout server {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s" }}
   timeout server-fin {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_SERVER_FIN_TIMEOUT") "1s" }}
   timeout http-request {{ firstMatch $timeSpecPattern (env "ROUTER_SLOWLORIS_TIMEOUT") "10s" }}
   timeout http-keep-alive {{ firstMatch $timeSpecPattern (env "ROUTER_SLOWLORIS_HTTP_KEEPALIVE") "300s" }}
-
-  # Long timeout for WebSocket connections.
-  timeout tunnel {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_TUNNEL_TIMEOUT") "1h" }}
 
   {{- if isTrue (env "ROUTER_ENABLE_COMPRESSION") }}
   compression algo gzip
@@ -318,6 +319,7 @@ frontend public_ssl
 # traffic
 ##########################################################################
 backend be_sni
+  timeout server {{ $routerMaxServerTimeout }}
   server fe_sni unix@/var/lib/haproxy/run/haproxy-sni.sock weight 1 send-proxy
 
 frontend fe_sni
@@ -434,6 +436,7 @@ frontend fe_sni
 ##########################################################################
 # backend for when sni does not exist, or ssl term needs to happen on the edge
 backend be_no_sni
+  timeout server {{ $routerMaxServerTimeout }}
   server fe_no_sni unix@/var/lib/haproxy/run/haproxy-no-sni.sock weight 1 send-proxy
 
 frontend fe_no_sni
@@ -593,11 +596,11 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
           {{- end }}
   tcp-request content reject if !whitelist
         {{- end }}
-        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
-  timeout server  {{ $value }}
+        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout") $routerDefaultServerTimeout) }}
+  timeout server {{ $value }}
         {{- end }}
-        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel")) }}
-  timeout tunnel  {{ $value }}
+        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel") $routerDefaultTunnelTimeout) }}
+  timeout tunnel {{ $value }}
         {{- end }}
 
         {{- if isTrue (index $cfg.Annotations "haproxy.router.openshift.io/rate-limit-connections") }}
@@ -797,8 +800,11 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
           {{- end }}
   tcp-request content reject if !whitelist
         {{- end }}
-        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel") (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
-  timeout tunnel  {{ $value }}
+        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout") $routerDefaultServerTimeout) }}
+  timeout server {{ $value }}
+        {{- end }}
+        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel") (index $cfg.Annotations "haproxy.router.openshift.io/timeout") $routerDefaultTunnelTimeout) }}
+  timeout tunnel {{ $value }}
         {{- end }}
 
         {{- if isTrue (index $cfg.Annotations "haproxy.router.openshift.io/rate-limit-connections") }}
--- a/pkg/router/template/template_helper.go
+++ b/pkg/router/template/template_helper.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"sync"
 	"text/template"
+	"time"
 
 	routev1 "github.com/openshift/api/route/v1"
 	"github.com/openshift/router/pkg/router/routeapihelpers"
@@ -389,6 +390,38 @@ func parseIPList(list string) string {
 	return list
 }
 
+// maxTimeoutFirstMatchedAndClipped finds the maximum timeout managed by a given annotation among all the routes, matches it against a given pattern, and clips.
+// The goal is to get the maximum timeout among the ones set on the lowest layer backends, rather than the maximum of all values provided to the function.
+// For instance, if a route has a timeout annotation set, it will take precedence over the default timeout, even if the default timeout is greater.
+func maxTimeoutFirstMatchedAndClipped(aliases map[ServiceAliasConfigKey]ServiceAliasConfig, annotation, pattern string, values ...string) string {
+	var (
+		max         string
+		maxDuration time.Duration
+	)
+	// find max timeout in route annotations
+	for _, cfg := range aliases {
+		timeout := clipHAProxyTimeoutValue(firstMatch(pattern, cfg.Annotations[annotation]))
+		if timeout != "" {
+			// No error handling because clipHAProxyTimeoutValue returns
+			// a valid timeout or an empty string. The latter is already handled.
+			timeoutDuration, _ := time.ParseDuration(timeout)
+			if timeoutDuration > maxDuration {
+				max = timeout
+				maxDuration = timeoutDuration
+			}
+		}
+	}
+	// use values if no max was found in routes
+	if max == "" {
+		max = clipHAProxyTimeoutValue(firstMatch(pattern, values...))
+	}
+	// use max haproxy timeout if no max was found
+	if max == "" {
+		max = templateutil.HaproxyMaxTimeout
+	}
+	return max
+}
+
 var helperFunctions = template.FuncMap{
 	"endpointsForAlias":        endpointsForAlias,        //returns the list of valid endpoints
 	"processEndpointsForAlias": processEndpointsForAlias, //returns the list of valid endpoints after processing them
@@ -416,4 +449,6 @@ var helperFunctions = template.FuncMap{
 	"parseIPList":             parseIPList,             //parses the list of IPs/CIDRs (IPv4/IPv6)
 
 	"processRewriteTarget": rewritetarget.SanitizeInput, //sanitizes `haproxy.router.openshift.io/rewrite-target` annotation
+
+	"maxTimeoutFirstMatchedAndClipped": maxTimeoutFirstMatchedAndClipped, //finds the maximum timeout managed by a given annotation among all the routes, matches it against a given pattern, and clips
 }
--- a/pkg/router/template/template_helper_test.go
+++ b/pkg/router/template/template_helper_test.go
@@ -1067,3 +1067,213 @@ func TestParseIPList(t *testing.T) {
 		})
 	}
 }
+
+func Test_maxTimeoutFirstMatchedAndClipped(t *testing.T) {
+	testCases := []struct {
+		name       string
+		state      map[ServiceAliasConfigKey]ServiceAliasConfig
+		annotation string
+		pattern    string
+		values     []string
+		expected   string
+	}{
+		{
+			name: "Route timeout is maximum",
+			state: map[ServiceAliasConfigKey]ServiceAliasConfig{
+				"test:route1": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout": "5m",
+					},
+				},
+				"test:route2": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout": "35s",
+					},
+				},
+				"test:route3": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout-tunnel": "10m",
+					},
+				},
+				"test:route4": {},
+			},
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			values:     []string{"30s"},
+			expected:   "5m",
+		},
+		{
+			name: "Default timeout is maximum",
+			state: map[ServiceAliasConfigKey]ServiceAliasConfig{
+				"test:route1": {},
+				"test:route2": {},
+				"test:route3": {},
+			},
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			values:     []string{"30s"},
+			expected:   "30s",
+		},
+		{
+			name: "First default timeout is choosen",
+			state: map[ServiceAliasConfigKey]ServiceAliasConfig{
+				"test:route1": {},
+				"test:route2": {},
+				"test:route3": {},
+			},
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			values:     []string{"30s", "40s"},
+			expected:   "30s",
+		},
+		{
+			name: "One route timeout doesn't match pattern",
+			state: map[ServiceAliasConfigKey]ServiceAliasConfig{
+				"test:route1": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout": "5minutes",
+					},
+				},
+				"test:route2": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout": "35s",
+					},
+				},
+			},
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			values:     []string{"30s"},
+			expected:   "35s",
+		},
+		{
+			name: "No route timeout matches pattern",
+			state: map[ServiceAliasConfigKey]ServiceAliasConfig{
+				"test:route1": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout": "5minutes",
+					},
+				},
+				"test:route2": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout": "35seconds",
+					},
+				},
+			},
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			values:     []string{"30s"},
+			expected:   "30s",
+		},
+		{
+			name: "Route timeout clipped",
+			state: map[ServiceAliasConfigKey]ServiceAliasConfig{
+				"test:route1": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout": "999999999s",
+					},
+				},
+			},
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			values:     []string{"30s"},
+			expected:   "2147483647ms",
+		},
+		{
+			name: "Default timeout overflows but route takes precedence",
+			state: map[ServiceAliasConfigKey]ServiceAliasConfig{
+				"test:route1": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout": "30s",
+					},
+				},
+			},
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			values:     []string{"999999999s"},
+			expected:   "30s",
+		},
+		{
+			name:       "Empty state",
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			values:     []string{"30s"},
+			expected:   "30s",
+		},
+		{
+			name:       "Empty state and values",
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			expected:   "2147483647ms",
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := maxTimeoutFirstMatchedAndClipped(tc.state, tc.annotation, tc.pattern, tc.values...)
+			if got != tc.expected {
+				t.Errorf("Failure: expected %q, got %q", tc.expected, got)
+			}
+		})
+	}
+}
+
+func Benchmark_maxTimeoutFirstMatchedAndClipped(b *testing.B) {
+	testCases := []struct {
+		name       string
+		state      map[ServiceAliasConfigKey]ServiceAliasConfig
+		annotation string
+		pattern    string
+		values     []string
+	}{
+		{
+			name: "Input1",
+			state: map[ServiceAliasConfigKey]ServiceAliasConfig{
+				"test:route1": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout": "5m",
+					},
+				},
+			},
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			values:     []string{"30s"},
+		},
+		{
+			name: "Input5",
+			state: map[ServiceAliasConfigKey]ServiceAliasConfig{
+				"test:route1": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout": "5m",
+					},
+				},
+				"test:route2": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout": "35s",
+					},
+				},
+				"test:route3": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout-tunnel": "35s",
+					},
+				},
+				"test:route4": {
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout-tunnel": "10m",
+					},
+				},
+				"test:route5": {},
+			},
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			// valid value is the last one to force all the iterations
+			values: []string{"", "", "", "", "30s"},
+		},
+	}
+	for _, tc := range testCases {
+		b.ResetTimer()
+		b.Run(tc.name, func(b *testing.B) {
+			for n := 0; n < b.N; n++ {
+				maxTimeoutFirstMatchedAndClipped(tc.state, tc.annotation, tc.pattern, tc.values...)
+			}
+		})
+	}
+}


ShudiLi:
Tested it with 4.15.0-0.ci.test-2023-11-16-093712-ci-ln-dr45512-latest
`
1.
% oc get clusterversion
version   4.15.0-0.ci.test-2023-11-16-093712-ci-ln-dr45512-latest   True        False         50m     Cluster version is 4.15.0-0.ci.test-2023-11-16-093712-ci-ln-dr45512-latest

2.
% oc get route
NAME     HOST/PORT                                                                   PATH   SERVICES        PORT   TERMINATION   WILDCARD
myedge   myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com          unsec-server3   http   edge          None

3.
 % oc annotate route myedge haproxy.router.openshift.io/timeout-tunnel=70m
route.route.openshift.io/myedge annotate

4.
 %  oc -n openshift-ingress rsh router-default-d78ddc6f-bprk5
sh-4.4$ grep -A7 "timeout server " haproxy-config.template
  timeout server  {{ $value }}
        {{- else }}
  timeout server {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s" }}
        {{- end }}
        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel")) }}
  timeout tunnel  {{ $value }}
        {{- else }}
  timeout tunnel {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_TUNNEL_TIMEOUT") "1h" }}
        {{- end }}

sh-4.4$

5. let client get a large size file with speed 1000bytes/s, started time at about 2023-11-17 03:18:56  and terminated time about 2023-11-17 04:38:15, more than 1h
sh-4.4# curl  --no-check-certificate --limit-rate=1000  --delete-after https://myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com/oc -k
curl: option --no-check-certificate: is unknown
curl: try 'curl --help' or 'curl --manual' for more information
sh-4.4# wget --no-check-certificate --limit-rate=1000  --delete-after https://myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com/oc -k
--2023-11-17 03:18:56--  https://myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com/oc
Resolving myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com (myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com)... 146.148.109.119
Connecting to myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com (myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com)|146.148.109.119|:443... connected.
WARNING: The certificate of 'myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com' is not trusted.
WARNING: The certificate of 'myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com' hasn't got a known issuer.
HTTP request sent, awaiting response... 200 OK
Length: 121359408 (116M)
Saving to: 'oc.tmp'

oc.tmp                                                               0%[                                                                                                                                                               ] 239.88K  1000 B/s    in 4m 6s

--2023-11-17 04:38:15--  (try:20)  https://myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com/oc
Connecting to myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com (myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com)|146.148.109.119|:443... connected.
WARNING: The certificate of 'myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com' is not trusted.
WARNING: The certificate of 'myedge-default.apps.ci-ln-9r4m3fk-72292.origin-ci-int-gce.dev.rhcloud.com' hasn't got a known issuer.
HTTP request sent, awaiting response... 206 Partial Content
Length: 121359408 (116M), 116746470 (111M) remaining
Saving to: 'oc.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp'

oc.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp.tmp   3%[++++++                                                                                                                                                         ]   4.62M  1000 B/s    in 3m 55s

2023-11-17 04:42:09 (1000 B/s) - Read error at byte 4847481/121359408 (The TLS connection was non-properly terminated.). Giving up.

sh-4.4#

6. check the captured packets and can see the session is terminated after more than 1h
tcpdump: listening on eth0, link-type EN10MB (Ethernet), capture size 262144 bytes
03:19:09.684312 IP (tos 0x0, ttl 64, id 2327, offset 0, flags [DF], proto TCP (6), length 52)
    146.148.109.119.https > 10.129.2.17.43942: Flags [.], cksum 0x0cc4 (incorrect -> 0x0a65), ack 1908667883, win 506, options [nop,nop,TS val 1841440389 ecr 1595459286], length 0
03:19:09.684337 IP (tos 0x0, ttl 64, id 12471, offset 0, flags [DF], proto TCP (6), length 52)
    10.129.2.17.43942 > 146.148.109.119.https: Flags [.], cksum 0x0cc4 (incorrect -> 0x189e), ack 1, win 0, options [nop,nop,TS val 1595462614 ecr 1841433925], length 0
03:19:16.340278 IP (tos 0x0, ttl 64, id 2328, offset 0, flags [DF], proto TCP (6), length 52)
    146.148.109.119.https > 10.129.2.17.43942: Flags [.], cksum 0x0cc4 (incorrect -> 0xe364), ack 1, win 506, options [nop,nop,TS val 1841447045 ecr 1595462614], length 0
03:19:16.340316 IP (tos 0x0, ttl 64, id 12472, offset 0, flags [DF], proto TCP (6), length 52)
    10.129.2.17.43942 > 146.148.109.119.https: Flags [.], cksum 0x0cc4 (incorrect -> 0xfe9d), ack 1, win 0, options [nop,nop,TS val 1595469270 ecr 1841433925], length 0
...

04:41:53.345283 IP (tos 0x0, ttl 62, id 24165, offset 0, flags [DF], proto TCP (6), length 628)
    146.148.109.119.https > 10.129.2.17.45884: Flags [P.], cksum 0xccdd (correct), seq 235640:236216, ack 904, win 505, options [nop,nop,TS val 927357442 ecr 1600426274], length 576
04:41:53.345307 IP (tos 0x0, ttl 62, id 24166, offset 0, flags [DF], proto TCP (6), length 628)
    146.148.109.119.https > 10.129.2.17.45884: Flags [.], cksum 0x0157 (correct), seq 236216:236792, ack 904, win 505, options [nop,nop,TS val 927357442 ecr 1600426274], length 576
04:41:53.345371 IP (tos 0x0, ttl 64, id 62346, offset 0, flags [DF], proto TCP (6), length 52)
`

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -583,9 +579,13 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
         {{- end }}
         {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
   timeout server  {{ $value }}
+        {{- else }}
+  timeout server {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s" }}
```
Miciah, line 583: This can be simplified:
```suggestion
        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout") (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s") }}
  timeout server  {{ $value }}
```

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -583,9 +579,13 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
...
         {{- end }}
         {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel")) }}
   timeout tunnel  {{ $value }}
+        {{- else }}
+  timeout tunnel {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_TUNNEL_TIMEOUT") "1h" }}
```
Miciah, line 588: This can be similarly simplified.

hack/Dockerfile.debug
```diff
@@ -1,5 +1,5 @@
-FROM centos:8
-RUN yum install -y https://github.com/frobware/haproxy-hacks/raw/master/RPMs/haproxy24-2.4.1-1.el8.x86_64.rpm
+FROM registry.access.redhat.com/ubi8/ubi
+RUN yum install -y https://github.com/frobware/haproxy-builds/raw/master/rhaos-4.14-rhel-8/haproxy26-2.6.13-2.rhaos4.14.el8.x86_64.rpm
```
Miciah, line 2: If we needed to backport the bug fix, would it be problematic to include this change?

alebedev87: It may. I'll do a dedicated PR for this change.

images/router/haproxy/conf/haproxy-config.template
alebedev87: Suggestion applied.

Miciah:
> Back to WIP to address the config warning about missing `timeout server` in the middle backends:
> 
> ```
> I1130 10:49:26.557528       1 router.go:649] template "msg"="router reloaded" "output"="[NOTICE]   (18) : haproxy version is 2.6.13-234aa6d\n[NOTICE]   (18) : path to executable is /usr/sbin/haproxy\n[WARNING]  (18) : config : missing timeouts for backend 'be_sni'.\n   | While not properly invalid, you will certainly encounter various problems\n   | with such a configuration. To fix this, please ensure that all following\n   | timeouts are set to a non-zero value: 'client', 'connect', 'server'.\n[WARNING]  (18) : config : missing timeouts for backend 'be_no_sni'.\n   | While not properly invalid, you will certainly encounter various problems\n   | with such a configuration. To fix this, please ensure that all following\n   | timeouts are set to a non-zero value: 'client', 'connect', 'server'.\n
> ```

I still see these errors in the most recent test results:

* [e2e-agnostic](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_router/536/pull-ci-openshift-router-master-e2e-agnostic/1731613996659773440) ([router-default-db7b6468-psrj9](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_router/536/pull-ci-openshift-router-master-e2e-agnostic/1731613996659773440/artifacts/e2e-agnostic/gather-extra/artifacts/pods/openshift-ingress_router-default-db7b6468-psrj9_router.log), [router-default-db7b6468-tlc5j](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_router/536/pull-ci-openshift-router-master-e2e-agnostic/1731613996659773440/artifacts/e2e-agnostic/gather-extra/artifacts/pods/openshift-ingress_router-default-db7b6468-tlc5j_router.log))
* [e2e-upgrade](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_router/536/pull-ci-openshift-router-master-e2e-upgrade/1730626233370152960) ([router-default-b69f7f7fb-79z8c](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_router/536/pull-ci-openshift-router-master-e2e-upgrade/1730626233370152960/artifacts/e2e-upgrade/gather-extra/artifacts/pods/openshift-ingress_router-default-b69f7f7fb-79z8c_router.log), [router-default-b69f7f7fb-gbq8b](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_router/536/pull-ci-openshift-router-master-e2e-upgrade/1730626233370152960/artifacts/e2e-upgrade/gather-extra/artifacts/pods/openshift-ingress_router-default-b69f7f7fb-gbq8b_router.log))
* [e2e-aws-serial](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_router/536/pull-ci-openshift-router-master-e2e-aws-serial/1730626233298849792) ([router-default-7b6688c47c-4d552](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_router/536/pull-ci-openshift-router-master-e2e-aws-serial/1730626233298849792/artifacts/e2e-aws-serial/gather-extra/artifacts/pods/openshift-ingress_router-default-7b6688c47c-4d552_router.log), [router-default-7b6688c47c-vz2kw](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_router/536/pull-ci-openshift-router-master-e2e-aws-serial/1730626233298849792/artifacts/e2e-aws-serial/gather-extra/artifacts/pods/openshift-ingress_router-default-7b6688c47c-vz2kw_router.log))

Did you mean remove the WIP label?

alebedev87:
> Did you mean remove the WIP label?

Yes, I removed WIP to start the discussion on what we want to do about this warning. Keeping it is one of the options. The other option I can see (complex and not elegant): to gather timeouts of all the routes and set all the middle backends with the server timeout greater than any of the route timeouts.

ShudiLi:
tested it with 4.15.0-0.test-2023-12-11-011641-ci-ln-z43xgmb-latest on a gcp cluster

`
1.
% oc get clusterversion
NAME      VERSION                                                AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.15.0-0.test-2023-12-11-011641-ci-ln-z43xgmb-latest   True        False         7h57m   Cluster version is 4.15.0-0.test-2023-12-11-011641-ci-ln-z43xgmb-latest

2.
% oc -n tunnel-timeout  get route hello-openshift -oyaml | grep 70m
    haproxy.router.openshift.io/timeout-tunnel: 70m

3.
% oc -n openshift-ingress-operator get ingresscontroller default -o yaml | grep -i clientTimeout
    clientTimeout: 7200s


4.
% oc -n openshift-ingress rsh router-default-868b6c666c-gbxvq
sh-4.4$ grep clipHAProxyTimeoutValue haproxy-config.template 
{{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (env "ROUTER_HARD_STOP_AFTER")) }}
        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout") (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s") }}
        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel") (env "ROUTER_DEFAULT_TUNNEL_TIMEOUT") "1h") }}
        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout") (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s") }}
        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel") (index $cfg.Annotations "haproxy.router.openshift.io/timeout") (env "ROUTER_DEFAULT_TUNNEL_TIMEOUT") "1h") }}
sh-4.4$ 

5. use WebSocket client to connect to the WebSocket server via the edge route
[root@localhost haproxy-timeout-tunnel]# ./run-cli.sh 
WebSocket Client Connected

6. The WebSocket server  is disconnected after 70m as expected(07:57:30 connected, 09:07:30 disconnected)
sh-4.4# Mon Dec 11 2023 07:57:30 GMT+0000 (Coordinated Universal Time) Connection accepted.

sh-4.4# netstat -ant
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN
tcp6       0      0 :::8080                 :::*                    LISTEN
tcp6       0      0 :::22                   :::*                    LISTEN
tcp6       0      0 10.131.0.16:8080        10.128.2.35:55898       ESTABLISHED
sh-4.4# Mon Dec 11 2023 09:07:30 GMT+0000 (Coordinated Universal Time) Peer 111.193.131.168 disconnected.

sh-4.4# netstat -ant
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN
tcp6       0      0 :::8080                 :::*                    LISTEN
tcp6       0      0 :::22                   :::*                    LISTEN
sh-4.4#
`

ShudiLi:
/label qe-approved
thanks @alebedev87 for testing this PR

candita:
@alebedev87 could you add an e2e test for this?

alebedev87:
The server timeout in the middle backends is addressed by setting it to the maximum timeout among all the routes. This was [one of the suggestions from the upstream issue](https://github.com/haproxy/haproxy/issues/2280#issuecomment-1711648863).

alebedev87:
@candita: yes, I plan to add a test case for this fix in [the cluster-ingress-operator's e2e](https://github.com/openshift/cluster-ingress-operator/tree/master/test/e2e).

ShudiLi:
Tested it again with 4.16.0-0.test-2023-12-25-011937-ci-ln-2g1gk12-latest, WebSocket Client was disconnected after about 30 seconds.

`
1.
sh-4.4# oc get clusterversion
NAME      VERSION                                                AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.16.0-0.test-2023-12-25-011937-ci-ln-2g1gk12-latest   True        False         124m    Cluster version is 4.16.0-0.test-2023-12-25-011937-ci-ln-2g1gk12-latest

2.
sh-4.4# ./run-cli.sh 
WebSocket Client Connected
echo-protocol Connection Closed

real	0m30.100s
user	0m0.081s
sys	0m0.018s
sh-4.4#
`

ShudiLi:
tested it with 4.16.0-0.test-2024-01-03-013347-ci-ln-mfg1zck-latest, the default clientTimeout is 30s, and the default erverTimeout is 300s, after configured clientTimeout and serverTimeout with 2h, websocket client was disconnected after 70m as expected

1.
% oc -n openshift-ingress-operator get ingresscontroller default -oyaml | grep -A3 tuningOptions:
  tuningOptions:
    clientTimeout: 7200s
    reloadInterval: 0s
    serverTimeout: 7200s

2.
%  oc -n tunnel-timeout get route hello-openshift -oyaml | grep -A1 annotations:                 
  annotations:
    haproxy.router.openshift.io/timeout-tunnel: 70m

3.
sh-4.4# ./run-cli.sh 
WebSocket Client Connected
echo-protocol Connection Closed

real	70m0.112s
user	0m0.091s
sys	0m0.017s
sh-4.4# 

4.
 % oc get clusterversion
NAME      VERSION                                                AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.16.0-0.test-2024-01-03-013347-ci-ln-mfg1zck-latest   True        False         4h16m   Cluster version is 4.16.0-0.test-2024-01-03-013347-ci-ln-mfg1zck-latest

ShudiLi:
Tested serverTimeout with 10s, 120s, 600s and the default, the function worked as expected in a created route
1. tuningOptions\serverTimeout: 600s
2. check haproxy.conf:
backend be_sni
  timeout server  10m
backend be_no_sni
  timeout server  10m

backend be_http:default:unsec-server3
  mode http
  option redispatch
  option forwardfor
  balance random
  timeout server  10m
  timeout tunnel  70m

  timeout check 5000ms
3. send traffic to a un-secure route, and check the captured packets on server, after 10m tcp reset packet is received
06:20:53.539042 IP (tos 0x0, ttl 64, id 28240, offset 0, flags [DF], proto TCP (6), length 483)
    10.129.2.17.39226 > 10.131.0.21.webcache: Flags [P.], cksum 0xdee3 (correct), seq 1:432, ack 1, win 504, options [nop,nop,TS val 1118141720 ecr 2226409851], length 431: HTTP, length: 431
        GET / HTTP/1.1
        user-agent: curl/7.61.1
        accept: */*
        host: unsec-server3-default.apps.shudi-416gcptime.qe.gcp.devcluster.openshift.com
        x-forwarded-host: unsec-server3-default.apps.shudi-416gcptime.qe.gcp.devcluster.openshift.com
        x-forwarded-port: 80
        x-forwarded-proto: http
        forwarded: for=10.131.0.22;host=unsec-server3-default.apps.shudi-416gcptime.qe.gcp.devcluster.openshift.com;proto=http
        x-forwarded-for: 10.131.0.22
06:30:53.540111 IP (tos 0x0, ttl 64, id 28241, offset 0, flags [DF], proto TCP (6), length 52)
    10.129.2.17.39226 > 10.131.0.21.webcache: Flags [R.], cksum 0x5f1d (correct), seq 432, ack 1, win 504, options [nop,nop,TS val 1118741720 ecr 2226409851], length 0
    
4.  set serverTimeout to defaut by remove tuningOptions\serverTimeout
5. check haproxy.conf
backend be_sni
  timeout server  5m
backend be_no_sni
  timeout server  5m


backend be_http:default:unsec-server3
  mode http
  option redispatch
  option forwardfor
  balance random
  timeout server  30s
  timeout tunnel  70m

alebedev87:
/retitle OCPBUGS-14914: set timeouts on the backend level

Wrong PR was WIPed, sorry.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -425,6 +424,9 @@ frontend fe_sni
 ##########################################################################
 # backend for when sni does not exist, or ssl term needs to happen on the edge
 backend be_no_sni
+        {{- with $value := maxTimeoutFirstMatchedAndClipped .State "haproxy.router.openshift.io/timeout" $timeSpecPattern (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s"  }}
```
candita, line 427: Why do you call the special `maxTimeoutFirstMatchedAndClipped` here but continue to call `clipHAProxyTimeoutValue` in the other places?

pkg/router/template/template_helper.go
```diff
@@ -386,6 +387,27 @@ func parseIPList(list string) string {
...
+	var (
+		max         string
+		maxDuration time.Duration
+	)
+	for _, cfg := range aliases {
+		timeout := clipHAProxyTimeoutValue(firstMatch(pattern, append([]string{cfg.Annotations[annotation]}, values...)...))
+		if timeout != "" {
+			// No error handling because clipHAProxyTimeoutValue returns
+			// a valid timeout or an empty string. The latter is already handled.
+			timeoutDuration, _ := haproxytime.ParseDuration(timeout)
```
candita, line 401: haproxytime.ParseDuration is called from within clipHAProxyTimeoutValue.  Here you can use the `time` package's ParseDuration.
```suggestion
			timeoutDuration, _ := time.ParseDuration(timeout)
```

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -312,6 +308,9 @@ frontend public_ssl
 # traffic
 ##########################################################################
 backend be_sni
+        {{- with $value := maxTimeoutFirstMatchedAndClipped .State "haproxy.router.openshift.io/timeout" $timeSpecPattern (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s"  }}
+  timeout server  {{ $value }}
```
candita, line 312: Was this annotation ever being applied before?

images/router/haproxy/conf/haproxy-config.template
alebedev87: We want [the maximum of all the backend timeouts](https://github.com/openshift/router/pull/536#issuecomment-1855963547) to be set on the middle backends (`be_sni`, `be_no_sni`). Otherwise we get [the warning during the config reloading](https://github.com/openshift/router/pull/536#issuecomment-1834660723).

images/router/haproxy/conf/haproxy-config.template
alebedev87: I didn't understand which annotation you are referring to. `timeout server` was not set in `be_sni` before, it's needed to prevent [the warning message](https://github.com/openshift/router/pull/536#issuecomment-1834660723).

pkg/router/template/template_helper.go
alebedev87: Done.

alebedev87:
https://github.com/openshift/router/pull/536#discussion_r1468223836 addressed.

ShudiLi:
Tested it with default, 10s, 90s, 70m on 4.16.0-0.test-2024-02-01-013554-ci-ln-g418bqb-latest

1.  tunnel time out
# oc get clusterversion
NAME      VERSION                                                AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.16.0-0.test-2024-02-01-013554-ci-ln-g418bqb-latest   True        False         8h      Cluster version is 4.16.0-0.test-2024-02-01-013554-ci-ln-g418bqb-latest

2.
sh-4.4# ./run-cli.sh
WebSocket Client Connected
echo-protocol Connection Closed

real	0m15.130s
user	0m0.101s
sys	0m0.019s
sh-4.4# 
sh-4.4# ./run-cli.sh
WebSocket Client Connected
echo-protocol Connection Closed

real	0m10.132s
user	0m0.108s
sys	0m0.016s
sh-4.4# 
sh-4.4# 
sh-4.4# ./run-cli.sh
WebSocket Client Connected
echo-protocol Connection Closed

real	0m30.128s
user	0m0.102s
sys	0m0.017s
sh-4.4# 
sh-4.4# ./run-cli.sh
WebSocket Client Connected
echo-protocol Connection Closed

real	0m30.151s
user	0m0.110s
sys	0m0.023s
sh-4.4# 
sh-4.4# ./run-cli.sh
WebSocket Client Connected
echo-protocol Connection Closed

real	1m30.138s
user	0m0.109s
sys	0m0.019s
sh-4.4# 

sh-4.4# ./run-cli.sh
WebSocket Client Connected
echo-protocol Connection Closed

real	70m0.140s
user	0m0.109s
sys	0m0.019s
sh-4.4#

3. server time out, just put the test result for 70m here
    10.129.2.18.43904 > 10.128.2.13.webcache: Flags [P.], cksum 0x3f23 (correct), seq 1:420, ack 1, win 258, options [nop,nop,TS val 2217547651 ecr 3480210042], length 419: HTTP, length: 419
	GET / HTTP/1.1
	user-agent: curl/7.61.1
	accept: */*
	host: unsec-server3-default.apps.shudi-416g01.qe.gcp.devcluster.openshift.com
	x-forwarded-host: unsec-server3-default.apps.shudi-416g01.qe.gcp.devcluster.openshift.com
	x-forwarded-port: 80
	x-forwarded-proto: http
	forwarded: for=10.128.2.14;host=unsec-server3-default.apps.shudi-416g01.qe.gcp.devcluster.openshift.com;proto=http
	x-forwarded-for: 10.128.2.14

    10.129.2.18.43904 > 10.128.2.13.webcache: Flags [R.], cksum 0x9a82 (correct), seq 420, ack 1, win 258, options [nop,nop,TS val 2221747652 ecr 3480210043], length 0

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -581,10 +583,10 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
           {{- end }}
   tcp-request content reject if !whitelist
         {{- end }}
-        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
+        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout") (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s") }}
```
Miciah, line 586: All these additional `env` calls add some overhead.  As Andy says, "Trips across the Go/Template boundary can be expensive, particularly when you have many, many routes (10,000+)."  Consider adding `{{- $defaultServerTimeout := firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s" }}` at the top of the template and using `{{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout") $defaultServerTimeout) }}` here.  

---

As noted in https://github.com/openshift/router/pull/483#pullrequestreview-1484951102, `clipHAProxyTimeoutValue (firstMatch $timeSpecPattern foo)` could be simplified as `clipHAProxyTimeoutValue foo`.  However, that might be better left as a separate change.  

...And actually, now that I look at it again, I believe `clipHAProxyTimeoutValue (firstMatch $timeSpecPattern) foo` will silently ignore invalid values while `clipHAProxyTimeoutValue foo` will log them.  Logging errors for invalid values might be a positive change, but it's an additional behavioral change that you might not want to tie to OCPBUGS-14914.

...Although thinking about it still further, I wonder:  Is `maxTimeoutFirstMatchedAndClipped` going to cause `clipHAProxyTimeoutValue` to log an error for each route that has an invalid annotation value, where before this change those invalid values get filtered out by `firstMatch` before `clipHAProxyTimeoutValue` has a chance to report any errors?

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -785,7 +787,10 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
           {{- end }}
   tcp-request content reject if !whitelist
         {{- end }}
-        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel") (index $cfg.Annotations "haproxy.router.openshift.io/timeout")) }}
+        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout") (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s") }}
+  timeout server  {{ $value }}
+        {{- end }}
+        {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout-tunnel") (index $cfg.Annotations "haproxy.router.openshift.io/timeout") (env "ROUTER_DEFAULT_TUNNEL_TIMEOUT") "1h") }}
```
Miciah, line 793: Please replace `env "ROUTER_DEFAULT_SERVER_TIMEOUT"` and `env "ROUTER_DEFAULT_TUNNEL_TIMEOUT"` each with a variable that is compute just once, if possible.

pkg/router/template/template_helper.go
```diff
@@ -386,6 +387,27 @@ func parseIPList(list string) string {
...
+	var (
+		max         string
+		maxDuration time.Duration
+	)
+	for _, cfg := range aliases {
+		timeout := clipHAProxyTimeoutValue(firstMatch(pattern, append([]string{cfg.Annotations[annotation]}, values...)...))
+		if timeout != "" {
+			// No error handling because clipHAProxyTimeoutValue returns
+			// a valid timeout or an empty string. The latter is already handled.
+			timeoutDuration, _ := time.ParseDuration(timeout)
```
Miciah, line 401: This is a lot of redundant computation:

* `maxTimeoutFirstMatchedAndClipped` parses each route's timeout annotation value using `firstMatch`, then (if the value is valid) `clipHAProxyTimeoutValue`, then `time.ParseDuration`.
* `fe_sni` calls `maxTimeoutFirstMatchedAndClipped`, so you're parsing each valid value 3 times.
* `fe_no_sni` also calls `maxTimeoutFirstMatchedAndClipped`, so you're parsing each valid value 3 more times.
* The backend calls `clipHAProxyTimeoutValue (firstMatch $timeSpecPattern ...)`, so you're parsing each valid value 2 more times.

Granted, the regexp cache that Andy added in #268 avoids the cost of compiling the same regexp repeatedly, but you're still repeatedly doing a regexp match and then `time.ParseDuration` on the same value, right?

pkg/router/template/template_helper.go
```diff
@@ -409,6 +431,7 @@ var helperFunctions = template.FuncMap{
 	"validateHAProxyWhiteList":     validateHAProxyWhiteList,     //validates a haproxy whitelist (acl) content
 	"generateHAProxyWhiteListFile": generateHAProxyWhiteListFile, //generates a haproxy whitelist file for use in an acl
 
-	"clipHAProxyTimeoutValue": clipHAProxyTimeoutValue, //clips extrodinarily high timeout values to be below the maximum allowed timeout value
-	"parseIPList":             parseIPList,             //parses the list of IPs/CIDRs (IPv4/IPv6)
+	"clipHAProxyTimeoutValue":          clipHAProxyTimeoutValue,          //clips extraordinarily high timeout values to be below the maximum allowed timeout value
+	"parseIPList":                      parseIPList,                      //parses the list of IPs/CIDRs (IPv4/IPv6)
+	"maxTimeoutFirstMatchedAndClipped": maxTimeoutFirstMatchedAndClipped, //finds the maximum timeout managed by a given annotation among all the routes, matches it against a given pattern, and clips
```
Miciah, line 436: Keep in mind that this list of helper functions is effectively an API.  We cannot rename or remove a helper function without potentially breaking third-party templates, so we tend to prefer defining general and composable helpers.  That said, I think it is worth putting this functionality in a single specialized helper to avoid excessive performance impact.

pkg/router/template/template_helper_test.go
```diff
@@ -1067,3 +1067,105 @@ func TestParseIPList(t *testing.T) {
 		})
 	}
 }
+
+func TestMaxTimeoutFirstMatchedAndClipped(t *testing.T) {
+	testCases := []struct {
+		name       string
+		state      map[ServiceAliasConfigKey]ServiceAliasConfig
+		annotation string
+		pattern    string
```
Miciah, line 1076: Could you add test cases with different values for `annotation` and `value`?

pkg/router/template/template_helper_test.go
```diff
@@ -1067,3 +1067,105 @@ func TestParseIPList(t *testing.T) {
...
+					Annotations: map[string]string{
+						"haproxy.router.openshift.io/timeout": "999999999s",
+					},
+				},
+			},
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			values:     []string{"30s"},
+			expected:   "2147483647ms",
+		},
```
Miciah, line 1161: How about a test case with `values: []string{"999999999s"}`?

pkg/router/template/template_helper_test.go
```diff
@@ -1067,3 +1067,105 @@ func TestParseIPList(t *testing.T) {
...
+			state: map[ServiceAliasConfigKey]ServiceAliasConfig{
+				"test:route1": {},
+				"test:route2": {},
+				"test:route3": {},
+			},
+			annotation: "haproxy.router.openshift.io/timeout",
+			pattern:    `[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			values:     []string{"30s"},
+			expected:   "30s",
+		},
```
Miciah, line 1116: Should at least one of the routes specify a value?  It makes sense to me to have a test case for "some routes specify timeouts, but the default timeout is maximum" .

Miciah: By the way, what happens if you have 0 routes?

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -158,14 +158,10 @@ defaults
   timeout connect {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_CONNECT_TIMEOUT") "5s" }}
   timeout client {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_CLIENT_TIMEOUT") "30s" }}
   timeout client-fin {{ firstMatch $timeSpecPattern (env "ROUTER_CLIENT_FIN_TIMEOUT") "1s" }}
-  timeout server {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s" }}
   timeout server-fin {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_SERVER_FIN_TIMEOUT") "1s" }}
   timeout http-request {{ firstMatch $timeSpecPattern (env "ROUTER_SLOWLORIS_TIMEOUT") "10s" }}
   timeout http-keep-alive {{ firstMatch $timeSpecPattern (env "ROUTER_SLOWLORIS_HTTP_KEEPALIVE") "300s" }}
 
-  # Long timeout for WebSocket connections.
-  timeout tunnel {{ firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_TUNNEL_TIMEOUT") "1h" }}
```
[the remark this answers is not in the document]
alebedev87: > Why don't you need to specify timeout tunnel on be_sni or be_no_sni? Is HAProxy more lenient about not specifying timeout tunnel than it is about not specifying timeout server?

Yes. Only the missing server timeout gives a warning message. I'm not sure  the warning reports a real problem though taking into account that the bottom backends will still have the server timeout set.

> Does omitting timeout tunnel result in an infinite timeout, similarly to omitting timeout server, or does it fall back to using the server timeout as the tunnel timeout? Crucially, if you have a route that has a tunnel timeout that is greater than any route's server timeout, is the tunnel timeout respected, or does the server timeout in be_sni/be_no_sni cause the connection to be terminated before the route's tunnel timeout?

Regarding the link between server and tunnel timeouts. [I discovered that the tunnel timeout supersedes server timeout in the TCP proxies](https://github.com/haproxy/haproxy/issues/2280#issuecomment-1906246293). Note that the middle backends in the router configuration are TCP proxies. So setting tunnel timeout in `default` or the middle backends (`be_sni`/`be_no_sni`) results into interference in the server timeout behavior set on the bottom backend level. I didn't see the opposite though: the server timeout set on `default` or middle backends didn't influence the tunnel timeout set on the bottom backends.

alebedev87:
Reanimating the PR: start addressing Miciah's review.

pkg/router/template/template_helper_test.go
alebedev87: > Should at least one of the routes specify a value? It makes sense to me to have a test case for "some routes specify timeouts, but the default timeout is maximum" .

The goal of maxTimeoutFirstMatchedAndClipped is to get the maximum timeout among the ones set on the lowest layer backends, rather than the maximum of all values provided to the function. For instance, if a route has a timeout annotation set, it will take precedence over the default timeout, even if the default timeout is greater. I updated the godoc for the function to highlight this.

> By the way, what happens if you have 0 routes?

Unit test case added to cover this case (`Empty state`).

pkg/router/template/template_helper_test.go
alebedev87: Done.

pkg/router/template/template_helper.go
alebedev87: The goal of the function was to define which timeout value will be set on the lowest layer backends. That's why I used exactly the same combination of function which is used in the template.

After some time spent on this one I didn't find a lot of room for the improvement here. I implemented a benchmark test which:
- allowed me to find a faster way to do the same: the max annotation timeout is now found first and the values are computed only once afterwards.
- showed that I don't get any visible performance gain if I try to get the `time.Duration` from `clipHAProxyTimeoutValue` to avoid another call to `ParseDuration`.

Also, [as you advised](https://github.com/openshift/router/pull/536#discussion_r1488665727) I moved the calculation of the max server timeout out of the backends so that it's done only one.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -312,6 +308,9 @@ frontend public_ssl
 # traffic
 ##########################################################################
 backend be_sni
+        {{- with $value := maxTimeoutFirstMatchedAndClipped .State "haproxy.router.openshift.io/timeout" $timeSpecPattern (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s"  }}
```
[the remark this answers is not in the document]
alebedev87: Moved the calculation of the maxservertimeout out of the middle backends to the top of the template.

images/router/haproxy/conf/haproxy-config.template
alebedev87: Done.

images/router/haproxy/conf/haproxy-config.template
alebedev87: >  Consider adding {{- $defaultServerTimeout := firstMatch $timeSpecPattern (env "ROUTER_DEFAULT_SERVER_TIMEOUT") "30s" }} at the top of the template and using {{- with $value := clipHAProxyTimeoutValue (firstMatch $timeSpecPattern (index $cfg.Annotations "haproxy.router.openshift.io/timeout") $defaultServerTimeout) }} here.

Done.

alebedev87:
High level perf test of `writeConfig` events for a cluster with ~ 1000 routes.

Baseline (CI image == 4.17):
```
 $ grep -Po 'writeConfig.*' baseline-1000-2-pod*.logs | cut -d ' ' -f1,2
baseline-1000-2-pod1.logs:writeConfig" "duration"="794.385841ms"
baseline-1000-2-pod1.logs:writeConfig" "duration"="867.438091ms"
baseline-1000-2-pod1.logs:writeConfig" "duration"="990.070294ms"
baseline-1000-2-pod1.logs:writeConfig" "duration"="852.333994ms"
baseline-1000-2-pod1.logs:writeConfig" "duration"="815.822647ms"
baseline-1000-2-pod1.logs:writeConfig" "duration"="988.009627ms"
baseline-1000-2-pod1.logs:writeConfig" "duration"="853.049032ms"
```

Change (image with this change):
```
$ grep -Po 'writeConfig.*' c67ef49-* | cut -d ' ' -f1,2
c67ef49-pod1.logs:writeConfig" "duration"="1.081579122s"
c67ef49-pod1.logs:writeConfig" "duration"="914.782473ms"
c67ef49-pod1.logs:writeConfig" "duration"="1.002188869s"
c67ef49-pod1.logs:writeConfig" "duration"="1.046333068s"
c67ef49-pod1.logs:writeConfig" "duration"="958.092477ms"
c67ef49-pod2.logs:writeConfig" "duration"="947.33866ms"
c67ef49-pod2.logs:writeConfig" "duration"="1.003091258s"
c67ef49-pod2.logs:writeConfig" "duration"="917.348525ms"
c67ef49-pod2.logs:writeConfig" "duration"="1.013730305s"
c67ef49-pod2.logs:writeConfig" "duration"="995.759517ms"
```

pkg/router/template/template_helper.go
alebedev87: > I think it is worth putting this functionality in a single specialized helper to avoid excessive performance impact.

I'm not sure I understand how putting the function into a different helper supposed to remove the API contract. Taking into account that the function will still be used in the config template.

Miciah: My point was that there is a tradeoff here, and we should be aware of it.  

- Pro: Simpler, more efficient template.  
- Con: Risk of breaking unsupported customizations if we ever change or remove the helper.  

However, it is probably worth making that tradeoff here.  I'm not asking you to change anything; I'm just trying to be deliberate and explicit about the design choice.

Makefile
```diff
@@ -25,6 +25,11 @@ GO_LDFLAGS ?=-ldflags "-s -w $(call version-ldflags,$(PACKAGE)/pkg/version) $(GO
 GO=GO111MODULE=on GOFLAGS=-mod=vendor go
 GO_BUILD_RECIPE=CGO_ENABLED=1 $(GO) build -o $(BIN) $(GO_GCFLAGS) $(GO_LDFLAGS) $(MAIN_PACKAGE)
 
+BENCH_PKGS ?= $(shell \grep -lR '^func Benchmark' | xargs -I {} dirname {} | sed 's|^|./|' | paste -sd ' ')
+BENCH_TEST ?= .
```
candita, line 29: Would you mind adding a couple of comments about what these are for?  I see below that `BENCH_PKGS` is a value for flag `-benchmem`, so it could use a clarification.

candita:
If I understand correctly,  this represents a major change for our users - whatever is the max connect timeout in ANY route becomes the max server and tunnel timeout for all the routes on the backend.  I strongly feel like this needs to be something that is configurable, and is off by default in order for past users to see the same behavior when they upgrade to a version that has this fix.

alebedev87:
> If I understand correctly, this represents a major change for our users - whatever is the max connect timeout in ANY route becomes the max server and tunnel timeout for all the routes on the backend. 

@candita: 1) only the server timeout is set to the max value on the middle backend, 2) this is only one part of the change. The other part is the server and tunnel timeouts set on all the route backends. Note that [the default timeout values are added to the route level timeouts](https://github.com/openshift/router/pull/536/files#diff-21bd64a4ad8a7927e9d59e99e09628efc884fe6d6291e50eb33c5deb3d2a79f0R599-R603).

We should not have any change in the behavior. Because for the routes which don't set any value for the timeouts (in the annotations) the default value will be set on the route backend.

> I strongly feel like this needs to be something that is configurable, and is off by default in order for past users to see the same behavior when they upgrade to a version that has this fix.

"Configurable" this may put us in the same situation as right now - when a user can set a global (almost) timeout and override the timeouts set on the route level. "Off by default" would lead us to [this warning](https://github.com/openshift/router/pull/536#issuecomment-1834660723) during the config reloads.

Overall, I understand the concern. The change may appear to have a potential to hide something I didn't manage to test manually and with our test suites. However I don't see any other solution for this bug. Let me try to organize a discussion about this PR outside of GitHub so that we can come to a conclusion.

Makefile
alebedev87: Done.

alebedev87:
Squashed the commits.

alebedev87:
The slidedeck has been updated with sequence diagrams: [current](https://docs.google.com/presentation/d/1vU1CH20lzFWDs8Z_iNaWBhPwdKM2iyS-Qb7zlMD9XTc/edit#slide=id.g310730ae176_0_25), [future](https://docs.google.com/presentation/d/1vU1CH20lzFWDs8Z_iNaWBhPwdKM2iyS-Qb7zlMD9XTc/edit#slide=id.g310730ae176_0_39).

PR #543, opened 2023-11-30
hack: Update debug dockerfile to install from new rpm url

philipgough:
Looks like the original location of the RPM has moved. Change updates accordingly.


philipgough:
Looks like duplicate of #544

PR #552, opened 2024-01-09
OCPBUGS-26498: Reject routes with SHA1 certs

gcs278:
Add logic to the extended route validation to reject any SHA1-based certs as they are no longer supported in 4.16+.

Also, the router will now clear `UnservableInFutureVersions` conditions for routes with SHA1 certs. This is desirable because the router is now rejecting routes with SHA1 cert and `UnservableInFutureVersions` no longer applies.

Previously, #555 was merged into 4.16 which sets `UnservableInFutureVersions` conditions for routes with SHA1 certs. #555  PR was intended to merge in 4.16 and backport into 4.15. This PR (#552) needs to partially reverts changes in #555 because this PR (#552) now rejects SHA1 certificates. This PR (#552) is ONLY intended for 4.16+ (it's not going to be backported).

### DIFF
--- a/pkg/router/routeapihelpers/validation.go
+++ b/pkg/router/routeapihelpers/validation.go
@@ -351,6 +351,8 @@ func validateCertificatePEM(certPEM string, options *x509.VerifyOptions) ([]*x50
 
 	// Reject any unsupported cert algorithms as HaProxy will refuse to start with them.
 	switch certs[0].SignatureAlgorithm {
+	case x509.SHA1WithRSA, x509.ECDSAWithSHA1:
+		return certs, fmt.Errorf("router does not support certs using SHA1")
 	case x509.MD5WithRSA:
 		return certs, fmt.Errorf("router does not support certs using MD5")
 	default:
--- a/pkg/router/routeapihelpers/validation.go
+++ b/pkg/router/routeapihelpers/validation.go
@@ -382,36 +382,11 @@ func validateCertificatePEM(certPEM string, options *x509.VerifyOptions) ([]*x50
 // a route. This checks for issues that will cause failures in the next
 // OpenShift version.
 func UpgradeRouteValidation(route *routev1.Route) field.ErrorList {
-	tlsConfig := route.Spec.TLS
-	result := field.ErrorList{}
-
-	if tlsConfig == nil {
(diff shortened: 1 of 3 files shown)


pkg/router/routeapihelpers/validation.go
```diff
@@ -351,6 +351,8 @@ func validateCertificatePEM(certPEM string, options *x509.VerifyOptions) ([]*x50
 
 	// Reject any unsupported cert algorithms as HaProxy will refuse to start with them.
 	switch certs[0].SignatureAlgorithm {
+	case x509.SHA1WithRSA, x509.ECDSAWithSHA1:
```
frobware, line 354: Should this be "SHA1" or "SHA-1"? I see both in use and I'm not sure which is preferred, or correct? Perhaps this details is... just a detail.

pkg/router/template/plugin_test.go
```diff
@@ -1162,20 +1162,6 @@ func TestHandleRouteUpgradeValidation(t *testing.T) {
...
-					Host: "www.reencrypt.sha1.test",
-					TLS: &routev1.TLSConfig{
-						Termination: routev1.TLSTerminationReencrypt,
-						Certificate: testCertificateRsaSha1,
-						Key:         testCertificateRsaSha1Key,
-					},
-				},
-			},
-			unservableInFutureVersionsExpected: true,
-		},
```
Miciah, line 1178: Why are you removing the test case, as opposed to flipping the expectation?
```suggestion
		{
			name: "SHA1 certificate should be unservable in future versions",
			route: &routev1.Route{
				Spec: routev1.RouteSpec{
					Host: "www.reencrypt.sha1.test",
					TLS: &routev1.TLSConfig{
						Termination: routev1.TLSTerminationReencrypt,
						Certificate: testCertificateRsaSha1,
						Key:         testCertificateRsaSha1Key,
					},
				},
			},
			unservableInFutureVersionsExpected: false,
		},
```

gcs278: Oh yes, we should just flip the expectation instead of removing it. Done.

Miciah:
Thanks!
/approve
/lgtm

PR #555, opened 2024-01-11
OCPBUGS-26498: Upgrade Validation plugin for SHA1 certs

gcs278:
This PR adds support for setting a new status condition type, `UnservableInFutureVersions`, when a SHA1 cert is detected on a route. It accomplishes this by adding a new plugin, the upgrade validation plugin. This plugin is simple: when a route configuration that is unsupported in the future is detected, it adds a `UnservableInFutureVersions` status condition to the route. When the problematic configuration is no longer detected, it clears/removes the `UnservableInFutureVersions` condition.

In order to support updating a new route status condition, the `routeRejectionRecorder` and associated methods were refactored to be more generic. The `writerlease` logic used to update route status had to be refactored because the old code made assumptions that only one type of status condition would ever be updated. 

It also updates certificates for existing unit tests, as some certificates were SHA1 and incorrectly labeled as expired.

Dependent on https://github.com/openshift/api/pull/1722
Related Ingress Operator PR https://github.com/openshift/cluster-ingress-operator/pull/1014


pkg/router/controller/status.go
```diff
@@ -146,30 +179,87 @@ func performIngressConditionUpdate(action string, lease writerlease.Lease, track
...
+		log.V(4).Info("updating route status failed due to write conflict", "action", action, "namespace", route.Namespace, "name", route.Name)
+		return writerlease.Release, true
+	default:
+		utilruntime.HandleError(fmt.Errorf("Unable to write router status for %s/%s: %v", route.Namespace, route.Name, err))
+		return writerlease.Release, true
+	}
+}
+
+// leaseKey creates a unique key for the writer lease logic given a route UID and a condition type.
+func leaseKey(uid types.UID, conditionType routev1.RouteIngressConditionType) string {
```
Miciah, line 250: Define a new type instead of using string.  This will help us make sure no one is using `route.UID` instead of `leaseKey(route.UID, condition.Type)`.

pkg/router/routeapihelpers/validation.go
```diff
@@ -367,3 +367,32 @@ func validateCertificatePEM(certPEM string, options *x509.VerifyOptions) ([]*x50
...
+	if tlsConfig == nil {
+		return result
+	}
+
+	tlsCertFieldPath := field.NewPath("spec").Child("tls").Child("certificate")
+	if len(tlsConfig.Certificate) > 0 {
+		certs, err := cert.ParseCertsPEM([]byte(tlsConfig.Certificate))
+		if err != nil {
+			return result
+		}
```
Miciah, line 387: I think you can do this `return result` (where result is an empty list of errors) because other validation will report errors from `ParseCertsPEM`, right?  Please add a comment explaining this reasoning.

pkg/router/routeapihelpers/validation.go
```diff
@@ -367,3 +367,32 @@ func validateCertificatePEM(certPEM string, options *x509.VerifyOptions) ([]*x50
...
+
+		if len(certs) < 1 {
+			return result
+		}
+
+		if certs[0].SignatureAlgorithm == x509.SHA1WithRSA || certs[0].SignatureAlgorithm == x509.ECDSAWithSHA1 {
+			result = append(result, field.Invalid(tlsCertFieldPath, "redacted certificate data", "version 4.16 of OpenShift does not support SHA1"))
```
Miciah, line 394: Declare `tlsCertFieldPath` inside the `if` block, closer to use.

pkg/router/controller/status.go
```diff
@@ -112,17 +128,34 @@ func (a *StatusAdmitter) RecordRouteRejection(route *routev1.Route, reason, mess
...
+func (a *StatusAdmitter) RecordRouteNotDeprecated(route *routev1.Route) {
+	log.V(4).Info("RecordRouteNotDeprecated")
+	performIngressConditionRemoval("not-deprecated", a.lease, a.tracker, a.client, a.lister, route, a.routerName, routev1.RouteDeprecated)
+}
+
 // performIngressConditionUpdate updates the route to the appropriate status for the provided condition.
 func performIngressConditionUpdate(action string, lease writerlease.Lease, tracker ContentionTracker, oc client.RoutesGetter, lister routelisters.RouteLister, route *routev1.Route, routerName, hostName string, condition routev1.RouteIngressCondition) {
-	key := string(route.UID)
+	// Key the lease off of the route UID and the condition type, as different conditions will require separate updates.
+	key := leaseKey(route.UID, condition.Type)
```
Miciah, line 150: This is actually a work-item key, per our discussion.

pkg/router/controller/status.go
```diff
@@ -112,17 +128,34 @@ func (a *StatusAdmitter) RecordRouteRejection(route *routev1.Route, reason, mess
...
 	routeNamespace, routeName := route.Namespace, route.Name
 
 	lease.Try(key, func() (writerlease.WorkResult, bool) {
 		route, err := lister.Routes(routeNamespace).Get(routeName)
 		if err != nil {
 			return writerlease.None, false
 		}
-		if string(route.UID) != key {
+		if string(route.UID) != getUIDFromLeaseKey(key) {
```
Miciah, line 158: As we discussed, it would be simpler to add an `oldRouteUID := route.UID` variable declaration outside the `lease.Try(key, func() (writerlease.WorkResult, bool) {` block and use that variable, rather than extracting the original UID from the key.

pkg/router/controller/status.go
gcs278: I'm not certain I did this how you were thinking about it, but take a look. I added the new type in the writelease logic and changed the functions such as `Try(key WorkKey, fn WorkFunc)` to take `WorkKey`. Let me know if I misunderstood.

pkg/router/controller/status.go
gcs278: Done.

pkg/router/routeapihelpers/validation.go
gcs278: Done.

pkg/router/routeapihelpers/validation.go
gcs278: Done.

pkg/router/controller/status.go
gcs278: Done.

frobware (changes requested):
LGTM overall. But this is a big change and would benefit from additional reviewers.

gcs278 (commented):
Thanks @frobware for a detailed review.

gcs278:
/hold
Waiting for reviews, but I also have an action to manually test route status conflicting scenarios.

pkg/router/controller/status.go
```diff
@@ -196,14 +196,15 @@ func recordIngressCondition(route *routev1.Route, name, hostName string, conditi
 			return false, false, time.Time{}, existing, existing
 		}
 
-		// preserve a copy of the original ingress without conditions
+		// preserve a copy of the original ingress
```
gcs278, line 286: Preserve a shallow copy

pkg/cmd/infra/router/router.go
gcs278, line 1: This is specific to OpenShift 4.15 and is expected to be removed
from 4.16 once backported to avoid confusion.

Add WHY specific to 4.15

gcs278: Add UPGRADE_VALIDATION env variable, but default to true, so that we can
disable in the future.

Add complementary to ExtendedValidation Plugin ENV

pkg/router/routeapihelpers/validation.go
```diff
@@ -375,3 +375,35 @@ func validateCertificatePEM(certPEM string, options *x509.VerifyOptions) ([]*x50
 
 	return certs, nil
 }
+
+// UpgradeRouteValidation performs an upgrade validation for
+// a route. This checks for issues that will cause failures in the next
+// OpenShift version.
+func UpgradeRouteValidation(route *routev1.Route) field.ErrorList {
```
gcs278, line 382: https://github.com/uber-go/nilaway

pkg/router/controller/status.go
```diff
@@ -234,14 +321,52 @@ func recordIngressCondition(route *routev1.Route, name, hostName string, conditi
 	return true, true, now.Time, ingress, nil
 }
 
-// findMostRecentIngress returns the name of the ingress status with the most recent Admitted condition transition time,
+// removeIngressCondition removes a matching status condition if it is found.
+// It returns whether the route was changed, the time it removed the condition, and
+// a pointer to the latest and original ingress records.
+func removeIngressCondition(route *routev1.Route, name string, condType routev1.RouteIngressConditionType) (changed bool, at time.Time, latest, original *routev1.RouteIngress) {
```
gcs278, line 327: Can you have more than 1 status? How does that work?

pkg/router/controller/status.go
```diff
@@ -234,14 +321,52 @@ func recordIngressCondition(route *routev1.Route, name, hostName string, conditi
...
+	for i := range route.Status.Ingress {
+		existing := &route.Status.Ingress[i]
+		if existing.RouterName != name {
+			continue
+		}
+
+		// preserve a copy of the original ingress without conditions
+		original := *existing
+		original.Conditions = nil
```
gcs278, line 336: This should not be here!

We should consider deep copy here.

pkg/cmd/infra/router/router.go
gcs278: Done. Updated commit.

pkg/router/controller/status.go
gcs278: Done.

pkg/router/controller/status.go
gcs278: Yes, you can. I added a unit test to for that in the rare scenario we'd ever encounter that.

pkg/router/routeapihelpers/validation.go
gcs278: I tried it out, I think it works, but I don't see any issues for the two main packages I'd be concerned with
```
[gspence@gspence-thinkpadp1gen3 router]$ nilaway -include-pkgs="controller" ./pkg/router/
[gspence@gspence-thinkpadp1gen3 router]$ nilaway -include-pkgs="routeapihelpers" ./pkg/router/
[gspence@gspence-thinkpadp1gen3 router]$
```

gcs278:
@frobware Some updates after testing: https://github.com/openshift/router/compare/dc757d4784f9c1d9f30b736a9bd7c7752424bd54..6e322eca91d5742aa870a28c308ca02f45578d3b

- In `performIngressConditionRemoval`, when there is no change needed, we checked `if findMostRecentIngress(route) == routerName {` before we extend the lease, however, since this is a removal function, the most `findMostRecentIngress`  function doesn't make sense here because the most recent condition we are looking for, has been removed.
- `if !created && tracker.IsChangeContended(string(route.UID), now, original) {`
   - I was using the `workKey` here, that's wrong. The contention tracker [uses the Route UID](https://github.com/openshift/router/blob/6e322eca91d5742aa870a28c308ca02f45578d3b/pkg/router/controller/contention.go#L107), not the new `workKey`.
   - So we were detecting contention, but not skipping any updates.

gcs278:
I've done lots of manual testing for status converging, status contention, and writerlease leader election with these updates. I have E2E tests ready to merge after this PR is completed (https://github.com/openshift/origin/pull/28710 & https://github.com/openshift/router/pull/575). 
/hold cancel

Miciah (commented):
Excellent work!  I made some comments, but the one thing I want to clarify is whether `UpgradeRouteValidation` should be checking `tlsConfig.CACertificate` or `tlsConfig.DestinationCACertificate` in addition to `tlsConfig.Certificate`.

pkg/router/routeapihelpers/validation_test.go
```diff
@@ -982,50 +1160,82 @@ func TestExtendedValidateRoute(t *testing.T) {
...
+						Key:           testCAUnknownCertificateKey,
 						CACertificate: testCACertificate,
 					},
 				},
 			},
 			expectedErrors: 1,
 		},
 		{
-			name: "Edge termination expired cert unknown CA",
+			name: "Edge termination cert with unknown CA",
```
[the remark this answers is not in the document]
gcs278: Yes, sorry this is painful to review. Thanks for writing this out.

> - "Edge termination expired cert unknown CA" is deleted and replaced by these two tests:
>   - "Edge termination expired cert with unknown CA", which uses an actually expired certificate so it does what the name says.
>   - "Edge termination cert with unknown CA", which does what "Edge termination expired cert unknown CA" actually did (despite the latter's misleading name).

Yes, that seems right.

> "Edge termination expired cert key mismatch" is renamed to "Edge termination cert key mismatch" to reflect what it actually tests. We already have "Expired cert mismatched key", which actually covers what the name implies (though unlike "Edge termination expired cert key mismatch", "Expired cert mismatched key" doesn't specify CACertificate).

Yes, this one just had an incorrect name.

> - "Edge termination expired cert mismatched CA" is deleted and replaced by these two tests:
>    - "Edge termination with expired cert, ignoring CA mismatch", which uses an actually expired certificate so it does what the name says. (Using an actually expired certificate also means that expectedErrors changes from 0 to 1.)
>    - "Edge termination cert mismatched CA", which does what "Edge termination expired cert mismatched CA" actually did (despite the misleading name).

Correct, the `testExpiredCAUnknownCertificate` it was using was NOT expired. Super confusing.

> "Expired cert" is renamed to "expired cert edge" and fixed to use an actually expired certificate; what it actually tested is already tested by "Edge termination OK with certs".

Correct.

> The rest of the changes are changes to individual tests, the test runner logic, or net-new tests.

I think so.

pkg/router/controller/status.go
```diff
@@ -168,8 +168,8 @@ func performIngressConditionUpdate(action string, lease writerlease.Lease, track
 }
 
 // recordIngressCondition updates the matching ingress on the route (or adds a new one) with the specified
-// condition, returning whether the route was updated or created, the time assigned to the condition, and
-// a pointer to the current ingress record.
+// condition. It returns whether the route was updated or created, the time assigned to the condition, and
+// a pointer to the latest and original ingress records.
```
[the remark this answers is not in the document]
gcs278: Sure, did a modified version of this.

pkg/router/controller/status_test.go
```diff
@@ -849,6 +852,246 @@ func TestRouterContention(t *testing.T) {
...
+						Host:                    "foo.foo.local",
+						RouterCanonicalHostname: "router-foo.foo.local",
+						RouterName:              "foo",
+						Conditions:              []routev1.RouteIngressCondition{admittedTrueCondition},
+					},
+				}},
+			},
+			expectCreated: false,
+			expectChanged: true,
+		},
```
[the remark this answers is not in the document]
gcs278: Oh yea my brackets were all over the place. Done.

gcs278 (commented):
@Miciah Thanks for the detailed review. Changes are here: https://github.com/openshift/router/compare/d5ea141e7ff30459aeff301e46f09bd7715c839c..c22b003ff4b0140aaa0fecefdd699a1b74717415

pkg/router/controller/status.go
```diff
@@ -133,40 +166,92 @@ func performIngressConditionUpdate(action string, lease writerlease.Lease, track
...
+
+		return handleRouteStatusUpdate(workKey, action, oc, route, latest, tracker)
 	})
 }
 
+// handleRouteStatusUpdate manages the update of route status in conjunction with a writerlease and a tracker. It
+// attempts to update the route status and, depending on the outcome, clears the tracker if necessary. It returns the
+// writerlease's WorkResult and a boolean flag indicating whether the writerlease should retry.
+func handleRouteStatusUpdate(key writerlease.WorkKey, action string, oc client.RoutesGetter, route *routev1.Route, latest *routev1.RouteIngress, tracker ContentionTracker) (workResult writerlease.WorkResult, retry bool) {
+	switch _, err := oc.Routes(route.Namespace).UpdateStatus(context.TODO(), route, metav1.UpdateOptions{}); {
```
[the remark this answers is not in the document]
Miciah: We should have a backlog item for replacing all instances of `context.TODO()`.

pkg/router/routeapihelpers/validation_test.go
```diff
@@ -982,50 +1160,82 @@ func TestExtendedValidateRoute(t *testing.T) {
...
+			name: "Edge termination cert key mismatch",
 			route: &routev1.Route{
 				Spec: routev1.RouteSpec{
 					Host: "www.example.com",
 					TLS: &routev1.TLSConfig{
 						Termination:   routev1.TLSTerminationEdge,
-						Certificate:   testExpiredCAUnknownCertificate,
+						Certificate:   testCAUnknownCertificate,
 						Key:           testPrivateKey,
 						CACertificate: testCACertificate,
```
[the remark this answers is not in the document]
Miciah: > > The test case expects only one error—perhaps for the same reason that "Edge termination with expired cert, ignoring CA mismatch" passes, in which case it might be helpful to add a comment saying as much.
> 
> Sorry, this is where I get confused. I thought you were taking about previously named `"Edge termination cert key mismatch"`, which returns 2 expectedErrors.

Sorry, I must have been mixing up test cases more misreading the diff.

pkg/router/controller/status_test.go
```diff
@@ -1089,6 +1091,9 @@ func Test_recordIngressCondition(t *testing.T) {
 			if tc.expectChanged != changed {
 				t.Errorf("expected changed=%t, but got changed=%t", tc.expectChanged, changed)
 			}
+			if diff := cmp.Diff(baselineIngressState, original, cmpOpts...); len(diff) > 0 {
+				t.Errorf("expected original to match baseline ingress from route.Status (-want +got):\n%s", diff)
+			}
```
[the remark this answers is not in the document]
Miciah: > * First, it had no impact other than logging. `original` was only used for logging

Hm, I see, I was mistakenly thinking that `original` was stored in the contention tracker.  (Even though `original` is passed to`(*SimpleContentionTracker) IsChangeContended`, the method doesn't actually use the value at all.)  You're right, there should be no memory impact, so it only matters for logging.

> * Besides being confusing to modify a variable called `original`,  removing the conditions removes half the benefit of logging [here](https://github.com/openshift/router/blob/d928435a15b19261166cfa57e504df58a61a4414/pkg/router/controller/status.go#L145).

Fair enough.  It's worth keeping the conditions if it makes logging more useful.  

> * It only clears conditions on the `original` return value when the ingress has changed...that feels confusing and inconsistent.

Also fair.

pkg/router/controller/status.go
```diff
@@ -133,40 +168,95 @@ func performIngressConditionUpdate(action string, lease writerlease.Lease, track
...
-			log.V(4).Info("updating route status failed due to write conflict", "action", action, "namespace", route.Namespace, "name", route.Name)
-			return writerlease.Release, true
-		default:
-			utilruntime.HandleError(fmt.Errorf("Unable to write router status for %s/%s: %v", route.Namespace, route.Name, err))
-			return writerlease.Release, true
 		}
+
+		return handleRouteStatusUpdate(context.TODO(), action, oc, route, latest, tracker)
 	})
 }
```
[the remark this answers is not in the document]
Miciah: > Yes I like that. I had the idea to do a callback, but got stuck following through with it, but this was I wanted.
> 
> > Maybe better left as a follow-up as it would be nontrivial to check for correctness.
> 
> Could you elaborate? It appear "correct" for our two use-cases (`performIngressConditionUpdate`, `performIngressConditionRemoval`), but are you saying we would need to explore "correctness" for other (future) use cases?
> 
> I ask because I prefer this logic, and don't exactly see how any more "non-trivial" it is than what I currently have.

I mean that it would be nontrivial for someone reviewing such a change to verify that the behavior of what you have and the behavior of what I suggested is identical.  If you're confident that my late-night refactoring is correct (or that your tests would catch any mistakes), feel free to use it.  * grin *.

Miciah:
Thanks!
/approve
/lgtm
Do you want to implement https://github.com/openshift/router/pull/555#discussion_r1571692555, or should we go with what you have?
/hold
for the above question, and in case Andy wants to have another look.

gcs278:
Thanks @Miciah! For the sake of time, I'm going to save https://github.com/openshift/router/pull/555#discussion_r1571692555 as a follow up. I do plan a follow up change in a hopeful fix for https://issues.redhat.com/browse/OCPBUGS-1689 as this PR has laid the groundwork from removing conditions.

I'll wait for to hear from @frobware to remove the hold.

gcs278:
Unrelated pod IP failure:
/test e2e-upgrade

pkg/router/controller/status.go
```diff
@@ -196,14 +196,15 @@ func recordIngressCondition(route *routev1.Route, name, hostName string, conditi
 			return false, false, time.Time{}, existing, existing
 		}
 
-		// preserve a copy of the original ingress without conditions
-		original := *existing
-		original.Conditions = nil
+		// preserve a deep copy of the original ingress
+		original := existing.DeepCopy()
```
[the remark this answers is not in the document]
frobware: > This seems like a pessimization.

You have shared ownership of the underlying array tied to the Conditions slice. If `existing` goes on to modify (i.e., reallocate) Conditions then the shallow copy is OK. But if no reallocation takes place then mutations to `existing.Conditions[x].Foo` will reflect in `original` too.

pkg/router/controller/status_test.go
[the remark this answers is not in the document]
frobware: > Hm, I see, I was mistakenly thinking that `original` was stored in the contention tracker. (Even though `original` is passed to`(*SimpleContentionTracker) IsChangeContended`, the method doesn't actually use the value at all.) You're right, there should be no memory impact, so it only matters for logging.

If `IsChangeContended` doesn't use value then I'd be inclined to drop it from the function signature. It would clear up any concern I have now (or in the future) w.r.t. the shallow copy.  

I didn't really chase this through as I didn't run the tests but the diff seems small (for a future PR). 

```diff
diff --git a/pkg/router/controller/contention.go b/pkg/router/controller/contention.go
index d9e1e234..0e7df160 100644
--- a/pkg/router/controller/contention.go
+++ b/pkg/router/controller/contention.go
@@ -24,7 +24,7 @@ type ContentionTracker interface {
 	// expected state of the object at this time and may be used by the tracker to
 	// determine if the most recent update was a contention. This method does not
 	// update the state of the tracker.
-	IsChangeContended(id contentionKey, now time.Time, current *routev1.RouteIngress) bool
+	IsChangeContended(id contentionKey, now time.Time) bool
 	// Clear informs the tracker that the provided ingress state was confirmed to
 	// match the current state of this process. If a subsequent call to IsChangeContended
 	// is made within the expiration window, the object will be considered as contended.
@@ -208,7 +208,7 @@ func (t *SimpleContentionTracker) Changed(id contentionKey, current *routev1.Rou
 	}
 }
 
-func (t *SimpleContentionTracker) IsChangeContended(id contentionKey, now time.Time, current *routev1.RouteIngress) bool {
+func (t *SimpleContentionTracker) IsChangeContended(id contentionKey, now time.Time) bool {
 	t.lock.Lock()
 	defer t.lock.Unlock()
 
diff --git a/pkg/router/controller/status.go b/pkg/router/controller/status.go
index 102a5730..f06f76d8 100644
--- a/pkg/router/controller/status.go
+++ b/pkg/router/controller/status.go
@@ -177,7 +177,7 @@ func performIngressConditionUpdate(action string, lease writerlease.Lease, track
 		// value, skip updating altogether and rely on the next resync to resolve conflicts. This prevents routers
 		// with different configurations from endlessly updating the route status.
 		// TRICKY: The tracker keys off of the route UID, not the workKey.
-		if !created && tracker.IsChangeContended(contentionKey(route.UID), now, original) {
+		if !created && tracker.IsChangeContended(contentionKey(route.UID), now) {
 			log.V(4).Info("skipped update due to another process altering the route with a different ingress status value", "action", action, "workKey", workKey, "original", original)
 			return writerlease.Release, false
 		}
@@ -204,7 +204,7 @@ func performIngressConditionRemoval(action string, lease writerlease.Lease, trac
 		}
 
 		route = route.DeepCopy()
-		changed, now, latest, original := removeIngressCondition(route, routerName, condType)
+		changed, now, latest := removeIngressCondition(route, routerName, condType)
 		if !changed {
 			log.V(4).Info("no changes to route needed", "action", action, "namespace", route.Namespace, "name", route.Name)
 			// Extending the lease ensures a delay in future work ONLY for followers. Unlike in
@@ -219,8 +219,8 @@ func performIngressConditionRemoval(action string, lease writerlease.Lease, trac
 		// value, skip updating altogether and rely on the next resync to resolve conflicts. This prevents routers
 		// with different configurations from endlessly updating the route status.
 		// TRICKY: The tracker keys off of the route UID, not the workKey.
-		if tracker.IsChangeContended(contentionKey(route.UID), now, original) {
-			log.V(4).Info("skipped update due to another process altering the route with a different ingress status value", "action", action, "workKey", workKey, "original", original)
+		if tracker.IsChangeContended(contentionKey(route.UID), now) {
+			log.V(4).Info("skipped update due to another process altering the route with a different ingress status value", "action", action, "workKey", workKey, "original")
 			return writerlease.Release, false
 		}
 
@@ -328,16 +328,13 @@ func recordIngressCondition(route *routev1.Route, name, hostName string, conditi
 // removeIngressCondition removes a matching status condition if it is found.
 // It returns whether the route was changed, the time it removed the condition, and
 // a pointer to the latest and original ingress records.
-func removeIngressCondition(route *routev1.Route, name string, condType routev1.RouteIngressConditionType) (changed bool, at time.Time, latest, original *routev1.RouteIngress) {
+func removeIngressCondition(route *routev1.Route, name string, condType routev1.RouteIngressConditionType) (changed bool, at time.Time, latest *routev1.RouteIngress) {
 	for i := range route.Status.Ingress {
 		existing := &route.Status.Ingress[i]
 		if existing.RouterName != name {
 			continue
 		}
 
-		// preserve a deep copy of the original ingress
-		original := existing.DeepCopy()
-
 		at := time.Time{}
 
 		// Remove the condition if we can find it.
@@ -357,10 +354,10 @@ func removeIngressCondition(route *routev1.Route, name string, condType routev1.
 			existing.Conditions = updatedConditions
 		}
 
-		return changed, at, existing, original
+		return changed, at, existing
 	}
 
-	return false, time.Time{}, nil, nil
+	return false, time.Time{}, nil
 }
 
 // findMostRecentIngress returns the name of the ingress status with the most recent condition transition time,
diff --git a/pkg/router/controller/status_test.go b/pkg/router/controller/status_test.go
index 0d8ce8ec..73ccde5b 100644
--- a/pkg/router/controller/status_test.go
+++ b/pkg/router/controller/status_test.go
@@ -1749,7 +1749,7 @@ func Test_removeIngressCondition(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			baselineIngressState := findIngressForRoute(tc.route, tc.routerName).DeepCopy()
-			changed, _, latest, original := removeIngressCondition(tc.route, tc.routerName, tc.conditionType)
+			changed, _, latest := removeIngressCondition(tc.route, tc.routerName, tc.conditionType)
 
 			// Compare expected route, but ignore LastTransitionTime since that is generated
 			cmpOpts := []cmp.Option{
```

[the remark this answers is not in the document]
Miciah: > I didn't really chase this through as I didn't run the tests but the diff seems small (for a future PR).

That's the thing—the fake tracker used in tests actually does need the argument.  If dropping the argument makes the tests more contorted or reduces our ability to test the tracker, then the change might not be worth it.

pkg/router/controller/status.go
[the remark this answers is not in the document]
frobware: Does this help explain my concern? 

```go
package main_test

import (
	"testing"
	"time"
)

var unixEpochTime = time.Unix(0, 0)

const epochMsg string = "Epoch"

type Condition struct {
	Type               string
	Status             string
	LastTransitionTime *time.Time
	Message            string
}

type Ingress struct {
	Host       string
	Conditions []Condition
}

type mutateConditionOp int

const (
	MutateExistingCondition mutateConditionOp = iota
	AddNewCondition
)

func findIngress(ingress *Ingress) *Ingress {
	// Return a shallow copy.
	original := *ingress
	return &original
}

func doStuffWithIngress(ingress *Ingress, op mutateConditionOp) {
	switch op {
	case MutateExistingCondition:
		for i := range ingress.Conditions {
			newTime := time.Now()
			ingress.Conditions[i].LastTransitionTime = &newTime
			ingress.Conditions[i].Message = "Mutated condition"
		}
	case AddNewCondition:
		// newTime := time.Now()
		// ingress.Conditions[0].LastTransitionTime = &newTime
		// ingress.Conditions[0].Message = "Mutated condition"
		//    or ...
		// ingress.Conditions = nil
		for i := 0; i < 9; i++ {
			newTime := time.Now()
			newCondition := Condition{
				Type: "Info", Status: "True", LastTransitionTime: &newTime, Message: "Additional condition",
			}
			ingress.Conditions = append(ingress.Conditions, newCondition)
		}
	}
}

// Test to demonstrate the surprise mutation due to shallow copying.
func TestUnexpectedMutation(t *testing.T) {
	ingress := Ingress{
		Host: "example.com",
		Conditions: []Condition{
			{Type: "Ready", Status: "True", LastTransitionTime: &unixEpochTime, Message: epochMsg},
		},
	}

	original := findIngress(&ingress)

	// Another part of the system unknowing to all and sundry
	// mutates an condition.
	doStuffWithIngress(&ingress, MutateExistingCondition)

	if !original.Conditions[0].LastTransitionTime.Equal(unixEpochTime) {
		t.Errorf("Expected original.Conditions[0].LastTransitionTime to be %q, got %q, so something somehere mutated the shallow copy", unixEpochTime, original.Conditions[0].LastTransitionTime)
	}

	if original.Conditions[0].Message != epochMsg {
		t.Errorf("Expected original.Conditionss[0].Message to be %q, got %q, so something somewhere mutated the shallow copy", epochMsg, original.Conditions[0].Message)
	}
}

// Test to demonstrate no mutation due to slice realloc.
func TestNoMutationBecauseOfSliceReallocation(t *testing.T) {
	ingress := Ingress{
		Host: "example.com",
		Conditions: []Condition{
			{Type: "Ready", Status: "True", LastTransitionTime: &unixEpochTime, Message: epochMsg},
		},
	}

	original := findIngress(&ingress)

	// Another part of the system unknowing to all and sundry
	// mutates the number of conditions.
	doStuffWithIngress(&ingress, AddNewCondition)

	// Verify that the "original" object was NOT mutated because
	// the underlying array backing ingress.Conditions got
	// reallocated.
	if e, l := 1, len(original.Conditions); e != l {
		t.Errorf("Expected `len(original.Conditions) %d, got %d", e, l)
	}

	if e, l := 1, cap(original.Conditions); e != l {
		t.Errorf("Expected `cap(original.Conditions) %d, got %d", e, l)
	}

	if !original.Conditions[0].LastTransitionTime.Equal(unixEpochTime) {
		t.Errorf("Expected original.Conditions[0].LastTransitionTime to remain unchanged, got %q, so something mutated the shallow copy", *original.Conditions[0].LastTransitionTime)
	}

	if original.Conditions[0].Message != epochMsg {
		t.Errorf("Expected original.Conditions[0].Message to remain %q, got %q, so something mutated the shallow copy", original.Conditions[0].Message, epochMsg)
	}

	if e, l := 10, len(ingress.Conditions); l != e {
		t.Errorf("Expected len(ingress.Conditions) = %v, got %v", e, l)
	}
}
```

PR #565, opened 2024-02-28
CFE-1020: feature:route external certificate reference

chiragkyal:
## Description

- Introduces new `RouteSecretManager` plugin for the management and validation of routes with `externalCertificate`.
- Integration with `secretManager` to register, unregister, and read referenced secret.
- Handling of the routes when referenced secret is updated or deleted.
- `ValidateTLSExternalCertificate` function to test preconditions for using `externalCertificate`.
- Everything is behind `AllowExternalCertificates` flag which is true when cluster-ingress-operator sets it, if `RouteExternalCertificate` feature-gate is enabled.

Implements: https://github.com/openshift/enhancements/pull/1307
Related to https://issues.redhat.com//browse/CFE-1020
Depends on: [CFE-866: Add secret monitor for Routes](https://github.com/openshift/library-go/pull/1626)
Feature-gate wiring: https://github.com/openshift/cluster-ingress-operator/pull/1017


chiragkyal:
/hold 
WIP

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,217 @@
...
+	return p.plugin.Commit()
+}
+
+// HandleRoute manages the registration, unregistration, and validation of routes with external certificates.
+// For Added events, it validates the route's external certificate configuration and registers it with the secret manager.
+// For Modified events, it first unregisters the route if it's already registered and then revalidates and registers it again.
+// For Deleted events, it unregisters the route if it's registered.
+// Additionally, it delegates the handling of the event to the next plugin in the chain after performing the necessary actions.
+func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *routev1.Route) error {
+	klog.Infof("Executing RouteSecretManager plugin with eventType %s...", eventType)
```
Miciah, line 69: Other plugins use the `log` variable; is there some reason to use `klog` here?
```suggestion
	log.V(10).Infof("HandleRoute: RouteSecretManager, eventType %s", eventType)
```
Log-level 10 would be consistent with this recent change: https://github.com/openshift/router/pull/588/commits/66eb2237bb1c6366b035ecd085d6456efa5f8663.

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,217 @@
...
+			if err := p.validateAndRegister(route); err != nil {
+				return err
+			}
+		}
+
+	case watch.Modified:
+		// unregister associated secret monitor, if registered
+		if p.secretManager.IsRouteRegistered(route.Namespace, route.Name) {
+			if err := p.secretManager.UnregisterRoute(route.Namespace, route.Name); err != nil {
+				klog.Error("failed to unregister route", err)
```
Miciah, line 84:
```suggestion
				log.Error(err, "failed to unregister route")
```

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,217 @@
...
+// UpdateFunc: Invoked when an existing secret is updated. It performs validation of the route's external certificate configuration.
+// If the validation fails, it records the route rejection, and triggers the deletion of the route by calling the HandleRoute method with a watch.Deleted event.
+// If the validation succeeds, it updates the route's TLS certificate and key with the new secret data and calls the next plugin's HandleRoute method with a watch.Modified event, and then the next plugin's Commit() method.
+// DeleteFunc: Invoked when the secret is deleted. It unregisters the associated route, records the route rejection, and triggers the deletion of the route by calling the HandleRoute method with a watch.Deleted event.
+func (p *RouteSecretManager) generateSecretHandler(route *routev1.Route) cache.ResourceEventHandlerFuncs {
+	// secret handler
+	secreth := cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			secret := obj.(*kapi.Secret)
+			klog.Infof("secret %s added for %s/%s", secret.Name, route.Namespace, route.Name)
```
Miciah, line 154:
```suggestion
			log.V(4).Info("secret added for route", "namespace", route.Namespace, "secret", secret.Name, "route", route.Name)
```

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,217 @@
...
+	if tls != nil && tls.ExternalCertificate != nil && len(tls.ExternalCertificate.Name) > 0 {
+		return true
+	}
+	return false
+}
+
+// must be called after hasExternalCertificate
+func getReferencedSecret(route *routev1.Route) string {
+	secretName := route.Spec.TLS.ExternalCertificate.Name
+	klog.Info("Referenced secretName: ", secretName)
```
Miciah, line 215: It seems odd that `getReferencedSecret` logs the secret name.  If you keep this log statement, it should have a high log level (I'd say at least 6, maybe 10).

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,217 @@
...
+	// read referenced secret
+	secret, err := p.secretManager.GetSecret(context.TODO(), route.Namespace, route.Name)
+	if err != nil {
+		klog.Error("failed to read referenced secret", err)
+		return err
+	}
+	// update tls.Certificate and tls.Key
+	// NOTE: this will be in-memory change and won't update actual route resource
+	route.Spec.TLS.Certificate = string(secret.Data["tls.crt"])
+	route.Spec.TLS.Key = string(secret.Data["tls.key"])
```
Miciah, line 138: Should this also clear `route.Spec.TLS.CACertificate`?  If not, please add a comment explaining why.

pkg/router/routeapihelpers/validation.go
```diff
@@ -390,3 +405,58 @@ func UpgradeRouteValidation(route *routev1.Route) field.ErrorList {
 	// are getting rejected now).
 	return nil
 }
+
+// validateTLSExternalCertificate tests different pre-conditions required for
+// using externalCertificate.
+func ValidateTLSExternalCertificate(route *routev1.Route, fldPath *field.Path, sarc authorizationclient.SubjectAccessReviewInterface, secretsGetter corev1client.SecretsGetter) field.ErrorList {
```
Miciah, line 411:
```suggestion
// ValidateTLSExternalCertificate tests different pre-conditions required for
// using externalCertificate.
func ValidateTLSExternalCertificate(route *routev1.Route, fldPath *field.Path, sarc authorizationclient.SubjectAccessReviewInterface, secretsGetter corev1client.SecretsGetter) field.ErrorList {
```

pkg/router/routeapihelpers/validation.go
```diff
@@ -2,17 +2,32 @@ package routeapihelpers
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	authorizationclient "k8s.io/client-go/kubernetes/typed/authorization/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+const (
+	// routerServiceAccount is used to validate RBAC permissions for externalCertificate
+	routerServiceAccount = "system:serviceaccount:openshift-ingress:router"
```
Miciah, line 30: Can you avoid hard-coding the serviceaccount name?  Maybe using an environment variable as described here: <https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables>.  That might require a change in cluster-ingress-operator, though, so feel free to leave this as a followup and just add a "to-do" comment.

Miciah:
This seems to be going in a good direction.  I like the approach and the way the core logic is isolated to the RouteSecretManager plugin.  I'd like to see more consistent logging and some unit tests.

pkg/router/routeapihelpers/validation.go
chiragkyal: Sure, we can use an environment variable to avoid hard coding. I've added a TODO as a follow-up.

pkg/router/routeapihelpers/validation.go
chiragkyal: Thanks, updated

pkg/router/controller/route_secret_manager.go
chiragkyal: since externalCertificate will not contain CACertificate, `tls.CACertificate` won't be updated. I've added a comment explaining this.

pkg/router/controller/route_secret_manager.go
chiragkyal: Thanks, I've removed `klog` from the plugin and now using `log` with required log-levels.

pkg/router/controller/route_secret_manager.go
chiragkyal: Thanks, updated to `log`

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,217 @@
...
+// validateAndRegister validates the route's externalCertificate configuration and registers it with the secret manager.
+// It also updates the in-memory TLS certificate and key after reading from secret informer's cache.
+func (p *RouteSecretManager) validateAndRegister(route *routev1.Route) error {
+	fldPath := field.NewPath("spec").Child("tls").Child("externalCertificate")
+	// validate
+	if err := routeapihelpers.ValidateTLSExternalCertificate(route, fldPath, p.sarClient, p.secretsGetter).ToAggregate(); err != nil {
+		klog.Error(err, "skipping route due to invalid externalCertificate configuration", " route ", route.Name)
+		p.recorder.RecordRouteRejection(route, "ExternalCertificateValidationFailed", err.Error())
+		p.plugin.HandleRoute(watch.Deleted, route)
+		return fmt.Errorf("invalid route configuration for externalCertificate")
```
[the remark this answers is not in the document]
chiragkyal: Sure, updated to `return err` so that it includes the actual validation error message.

pkg/router/controller/route_secret_manager.go
chiragkyal: Thanks, updated to `log` with correct log-level

pkg/router/controller/route_secret_manager.go
chiragkyal: I'll keep this for now at the highest log-level (10), incase of any debugging need.

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
+			eventType: watch.Added,
+			expectedRoute: &routev1.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "route-test",
+					Namespace: "sandbox",
+				},
+				Spec: routev1.RouteSpec{},
+			},
+			expectedEventType:  watch.Added,
+			expectedRejections: map[string]string{},
```
alebedev87, line 138: Can we modify the test case checks below so that the default value for `expectedRejections` would be fine? This would help us not to specify an empty map each time.

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
+						ExternalCertificate: &routev1.LocalObjectReference{
+							Name: "tls-secret",
+						},
+					},
+				},
+			},
+			expectedEventType: watch.Deleted,
+			expectedRejections: map[string]string{
+				"sandbox-route-test": "ExternalCertificateValidationFailed",
+			},
```
alebedev87, line 177: You don't need to set all the expected values because you test case will fail before, at the error check.

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
```
alebedev87, line 216: Same here, these fields won't be checked.

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
+							Name: "tls-secret",
+						},
+						Certificate: "my-crt",
+						Key:         "my-key",
+					},
+				},
+			},
+			expectedEventType:  watch.Added,
+			expectedRejections: map[string]string{},
+			expectedError:      false,
```
alebedev87, line 298: This line can be removed, the default for boolean is `false`. Also, can you please move all the positive cases at the top or at the bottom? This would help the reader.

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
+		},
+		{
+			name: "route updated: old route without externalCertificate, new route without externalCertificate",
+			route: &routev1.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "route-test",
+					Namespace: "sandbox",
+				},
+				Spec: routev1.RouteSpec{},
+			},
```
alebedev87, line 335: This is the new route I suppose. Where is the definition of the old one? How can I understand the scenario tests what it states in the name?

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
+				"sandbox-route-test": "ExternalCertificateSecretDeleted",
+			},
+		},
+	}
+
+	for _, s := range scenarios {
+		t.Run(s.name, func(t *testing.T) {
+			oldSecret := fakeSecret("sandbox", "tls-secret", corev1.SecretTypeTLS, map[string][]byte{})
+			kubeClient := testclient.NewSimpleClientset(oldSecret)
+			informer := fakeSecretInformer(context.TODO(), kubeClient, "sandbox", "tls-secret")
```
alebedev87, line 1096: You don't need a dedicated argument for a context which you don't really care about. Just set it to `TODO()` everywhere you need in `fakeSecretInformer()`.

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
+}
+
+func (t *testSecretGetter) Secrets(_ string) corev1client.SecretInterface {
+	return testclient.NewSimpleClientset(t.secret).CoreV1().Secrets(t.namespace)
+}
+
+// fakeSecretInformer will list/watch only one secret inside a namespace
+func fakeSecretInformer(ctx context.Context, fakeKubeClient *testclient.Clientset, namespace, name string) cache.SharedInformer {
+	fieldSelector := fields.OneTermEqualSelector("metadata.name", name).String()
+	klog.Info(fieldSelector)
```
alebedev87, line 52: Why this log in a fake function used only in tests?

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
+func (p *fakePluginNew) HandleNode(t watch.EventType, node *corev1.Node) error {
+	return fmt.Errorf("not expected")
+}
+
+func (p *fakePluginNew) HandleEndpoints(watch.EventType, *corev1.Endpoints) error {
+	return fmt.Errorf("not expected")
+}
+func (p *fakePluginNew) HandleNamespaces(namespaces sets.String) error {
+	return fmt.Errorf("not expected")
+}
```
alebedev87, line 102: Does `fakePluginNew` implements some interface? Why do you need all these methods defined?

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
+			recorder := routeStatusRecorder{rejections: make(map[string]string)}
+			rsm := NewRouteSecretManager(p, recorder, &s.secretManager, &testSecretGetter{namespace: s.route.Namespace, secret: oldSecret}, &testSARCreator{allow: s.allow})
+
+			if _, err := informer.AddEventHandler(rsm.generateSecretHandler(s.route)); err != nil {
+				t.Fatal("failed to add handler", err)
+			}
+			// wait for informer to start
+			if !cache.WaitForCacheSync(context.TODO().Done(), informer.HasSynced) {
+				t.Fatal("cache not synced yet")
+			}
```
alebedev87, line 1111: Why do you wait here and not above where you run the informer?

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
+	for _, s := range scenarios {
+		t.Run(s.name, func(t *testing.T) {
+			oldSecret := fakeSecret("sandbox", "tls-secret", corev1.SecretTypeTLS, map[string][]byte{})
+			kubeClient := testclient.NewSimpleClientset(oldSecret)
+			informer := fakeSecretInformer(context.TODO(), kubeClient, "sandbox", "tls-secret")
+			go informer.Run(context.TODO().Done())
+
+			p := &fakePluginNew{
+				stopCh: make(chan struct{}),
+			}
```
alebedev87, line 1101: `New` looks too vague, maybe something like:
```suggestion
			p := &fakePluginDone{
				doneCh: make(chan struct{}),
			}
```

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
+			recorder := routeStatusRecorder{rejections: make(map[string]string)}
+			rsm := NewRouteSecretManager(p, recorder, &s.secretManager, &testSecretGetter{namespace: s.route.Namespace, secret: oldSecret}, &testSARCreator{allow: s.allow})
+
+			if _, err := informer.AddEventHandler(rsm.generateSecretHandler(s.route)); err != nil {
+				t.Fatal("failed to add handler", err)
```
alebedev87, line 1106:
```suggestion
				t.Fatalf("failed to add handler: %v", err)
```

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
+			}
+			// wait for informer to start
+			if !cache.WaitForCacheSync(context.TODO().Done(), informer.HasSynced) {
+				t.Fatal("cache not synced yet")
+			}
+
+			if s.deleteSecret {
+				// delete the secret
+				if err := kubeClient.CoreV1().Secrets(s.route.Namespace).Delete(context.TODO(), s.secretManager.Secret.Name, metav1.DeleteOptions{}); err != nil {
+					t.Fatal("failed to delete secret", err)
```
alebedev87, line 1116:
```suggestion
					t.Fatalf("failed to delete secret: %v", err)
```

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1139 @@
...
+			expectedEventType: watch.Deleted,
+			expectedRejections: map[string]string{
+				"sandbox-route-test": "ExternalCertificateSecretDeleted",
+			},
+		},
+	}
+
+	for _, s := range scenarios {
+		t.Run(s.name, func(t *testing.T) {
+			oldSecret := fakeSecret("sandbox", "tls-secret", corev1.SecretTypeTLS, map[string][]byte{})
```
alebedev87, line 1094: Why this is not a part of the test scenario input so that the reader can understand immediately the update cases?

alebedev87 (commented):
Mostly nitpicks and remarks about the unit tests.

pkg/cmd/infra/router/router.go
```diff
@@ -107,6 +111,7 @@ func (o *RouterSelection) Bind(flag *pflag.FlagSet) {
 	flag.MarkDeprecated("enable-ingress", "Ingress resources are now synchronized to routes automatically.")
 	flag.StringVar(&o.ListenAddr, "listen-addr", env("ROUTER_LISTEN_ADDR", ""), "The name of an interface to listen on to expose metrics and health checking. If not specified, will not listen. Overrides stats port.")
 	flag.BoolVar(&o.WatchEndpoints, "watch-endpoints", isTrue(env("ROUTER_WATCH_ENDPOINTS", "")), "Watch Endpoints instead of the EndpointSlice resource.")
+	flag.BoolVar(&o.AllowExternalCertificates, "allow-external-certificates", isTrue(env("ROUTER_ENABLE_EXTERNAL_CERTIFICATE", "")), "Enable RouteExternalCertificate feature-gate, disabled by default.")
```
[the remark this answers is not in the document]
chiragkyal: Agreed, that router doesn't know anything about this feature gate.

I've updated it to `Enable RouteSecretManager plugin and validation of external certificates` because the main purpose of this flag is to enable `RouteSecretManager` plugin along with validating external certificates.

pkg/cmd/infra/router/router.go
```diff
@@ -78,6 +78,10 @@ type RouterSelection struct {
 	// WatchEndpoints when true will watch Endpoints instead of
 	// EndpointSlices.
 	WatchEndpoints bool
+
+	// AllowExternalCertificates is set when cluster-ingress-operator propagates
+	// ROUTER_ENABLE_EXTERNAL_CERTIFICATE env var, if RouteExternalCertificate feature-gate is enabled.
```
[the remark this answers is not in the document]
chiragkyal: Updated to 
```go
// AllowExternalCertificates when true enables RouteSecretManager plugin and the external certificate validation.
// The cluster-ingress-operator sets it if RouteExternalCertificate feature-gate is enabled.
```

because it will also enable `RouteSecretManager` plugin.

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,219 @@
...
+	// validate
+	if err := routeapihelpers.ValidateTLSExternalCertificate(route, fldPath, p.sarClient, p.secretsGetter).ToAggregate(); err != nil {
+		log.Error(err, "skipping route due to invalid externalCertificate configuration", "namespace", route.Namespace, "route", route.Name)
+		p.recorder.RecordRouteRejection(route, "ExternalCertificateValidationFailed", err.Error())
+		p.plugin.HandleRoute(watch.Deleted, route)
+		return err
+	}
+
+	// register route with secretManager
+	secreth := p.generateSecretHandler(route)
```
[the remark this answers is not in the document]
chiragkyal: Updated

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,219 @@
...
+
+// generateSecretHandler creates ResourceEventHandlerFuncs to handle Add, Update, and Delete events on secrets.
+// AddFunc: Invoked when a new secret is added. It logs the addition of the secret.
+// UpdateFunc: Invoked when an existing secret is updated. It performs validation of the route's external certificate configuration.
+// If the validation fails, it records the route rejection, and triggers the deletion of the route by calling the HandleRoute method with a watch.Deleted event.
+// If the validation succeeds, it updates the route's TLS certificate and key with the new secret data and calls the next plugin's HandleRoute method with a watch.Modified event, and then the next plugin's Commit() method.
+// DeleteFunc: Invoked when the secret is deleted. It unregisters the associated route, records the route rejection, and triggers the deletion of the route by calling the HandleRoute method with a watch.Deleted event.
+func (p *RouteSecretManager) generateSecretHandler(route *routev1.Route) cache.ResourceEventHandlerFuncs {
+	// secret handler
+	secreth := cache.ResourceEventHandlerFuncs{
```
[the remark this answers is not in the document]
chiragkyal: Updated

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,219 @@
...
+		AddFunc: func(obj interface{}) {
+			secret := obj.(*kapi.Secret)
+			log.V(4).Info("secret added for route", "namespace", route.Namespace, "secret", secret.Name, "route", route.Name)
+			// Do nothing for add event
```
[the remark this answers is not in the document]
chiragkyal: Added a comment with the reason behind it. Let me know if it's better now.

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,219 @@
...
+
+			// call the next plugin with watch.Modified
+			p.plugin.HandleRoute(watch.Modified, route)
+			// commit the changes
+			p.plugin.Commit()
+		},
+		DeleteFunc: func(obj interface{}) {
+			secret := obj.(*kapi.Secret)
+			msg := fmt.Sprintf("secret %s deleted for route %s/%s", secret.Name, route.Namespace, route.Name)
+			log.V(4).Info(msg)
```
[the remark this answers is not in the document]
chiragkyal: I chose this to avoid duplication of the message. We need the same message while logging and also while calling `RecordRouteRejection`. 
I'm fine removing the `msg` variable, and directly using in both the places.

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,219 @@
...
+
+// validateAndRegister validates the route's externalCertificate configuration and registers it with the secret manager.
+// It also updates the in-memory TLS certificate and key after reading from secret informer's cache.
+func (p *RouteSecretManager) validateAndRegister(route *routev1.Route) error {
+	fldPath := field.NewPath("spec").Child("tls").Child("externalCertificate")
+	// validate
+	if err := routeapihelpers.ValidateTLSExternalCertificate(route, fldPath, p.sarClient, p.secretsGetter).ToAggregate(); err != nil {
+		log.Error(err, "skipping route due to invalid externalCertificate configuration", "namespace", route.Namespace, "route", route.Name)
+		p.recorder.RecordRouteRejection(route, "ExternalCertificateValidationFailed", err.Error())
+		p.plugin.HandleRoute(watch.Deleted, route)
```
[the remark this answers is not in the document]
chiragkyal: We don't want the router to serve this route if validation failed, and to indicate the same, we're passing down `Deleted`.
This is similar to what we are currently doing with `ExtendedValidator` 
https://github.com/openshift/router/blob/56ab14f257e123317c00ce7ba5b4ecae7b2e0102/pkg/router/controller/extended_validator.go#L51-L57

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,219 @@
...
+func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *routev1.Route) error {
+	log.V(10).Info("HandleRoute: RouteSecretManager", "eventType", eventType)
+
+	switch eventType {
+	case watch.Added:
+		// register with secret monitor
+		if hasExternalCertificate(route) {
+			if err := p.validateAndRegister(route); err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
chiragkyal: I don’t think we always want to return after calling `validateAndRegister()` , we will only return if there is an error from `validateAndRegister` . If no error, then we want to call the next plugin. 

If we always return, then the next plugin won’t be called after registering with the `secretManager`.

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,219 @@
...
+			if err := p.secretManager.UnregisterRoute(route.Namespace, route.Name); err != nil {
+				log.Error(err, "failed to unregister route")
+				return err
+			}
+		}
+		// register with secret monitor
+		if hasExternalCertificate(route) {
+			if err := p.validateAndRegister(route); err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
chiragkyal: Same as https://github.com/openshift/router/pull/565#discussion_r1593656264

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,219 @@
...
+	// register route with secretManager
+	secreth := p.generateSecretHandler(route)
+	if err := p.secretManager.RegisterRoute(context.TODO(), route.Namespace, route.Name, getReferencedSecret(route), secreth); err != nil {
+		log.Error(err, "failed to register route")
+		return err
+	}
+	// read referenced secret
+	secret, err := p.secretManager.GetSecret(context.TODO(), route.Namespace, route.Name)
+	if err != nil {
+		log.Error(err, "failed to read referenced secret")
```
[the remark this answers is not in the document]
chiragkyal: Updated

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,219 @@
...
+		return err
+	}
+	// update tls.Certificate and tls.Key
+	// since externalCertificate will not contain CACertificate, tls.CACertificate won't be updated.
+	// NOTE: this will be in-memory change and won't update actual route resource.
```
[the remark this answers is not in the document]
chiragkyal: I've added the reason. Hope it's better now

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,219 @@
...
+				p.recorder.RecordRouteRejection(route, "ExternalCertificateValidationFailed", err.Error())
+				p.plugin.HandleRoute(watch.Deleted, route)
+				return
+			}
+
+			// read referenced secret (updated data)
+			secret, err := p.secretManager.GetSecret(context.TODO(), route.Namespace, route.Name)
+			if err != nil {
+				log.Error(err, "failed to read referenced secret")
+				p.recorder.RecordRouteRejection(route, "ExternalCertificateReadFailed", err.Error())
```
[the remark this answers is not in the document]
chiragkyal: Updated

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	return secreth
+}
+
+func hasExternalCertificate(route *routev1.Route) bool {
+	tls := route.Spec.TLS
+	if tls != nil && tls.ExternalCertificate != nil && len(tls.ExternalCertificate.Name) > 0 {
+		return true
+	}
+	return false
```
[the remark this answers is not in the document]
chiragkyal: Thanks, updated

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// must be called after hasExternalCertificate
+func getReferencedSecret(route *routev1.Route) string {
+	secretName := route.Spec.TLS.ExternalCertificate.Name
+	log.V(10).Info("referenced externalCertificate", "secret", secretName)
+	return secretName
+}
```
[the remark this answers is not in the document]
chiragkyal: Yeah, makes sense to me. Removed `getReferencedSecret` and directly using `route.Spec.TLS.ExternalCertificate.Name`

pkg/router/controller/route_secret_manager_test.go
chiragkyal: Thanks for the suggestion. Updated the test to accommodate default value for `expectedRejections`

pkg/router/controller/route_secret_manager_test.go
chiragkyal: Test case will not fail, because we are expecting error from the test case by setting `expectedError: true`

So,

```go
			gotErr := rsm.HandleRoute(s.eventType, s.route)
			if (gotErr != nil) != s.expectedError {
				t.Fatalf("expected error to be %t, but got %t", s.expectedError, gotErr != nil)
			}
```
^^ the above `Fatalf` won’t be executed (even when  `expectedError: true`) and test will continue checking other expected values like `expectedRejections`. So, we need to define `expectedRejections` field as well.

pkg/router/controller/route_secret_manager_test.go
chiragkyal: Same as https://github.com/openshift/router/pull/565#discussion_r1593668081

I think this field will be checked unless the test fails earlier.

pkg/router/controller/route_secret_manager_test.go
chiragkyal: >This line can be removed, the default for boolean is `false`. 

Done, thanks

>Also, can you please move all the positive cases at the top or at the bottom? This would help the reader.

Updated the logical grouping of the tests like the following:

- scenarios when route is added
- scenarios when route is updated (old route without externalCertificate, new route with externalCertificate)
- scenarios when route is updated (old route with externalCertificate, new route with externalCertificate)
- scenarios when route is updated (old route with externalCertificate, new route without externalCertificate)
- scenario when route is updated (old route without externalCertificate, new route without externalCertificate)
- scenarios when route is deleted

All the above logical groups now have all `negative` cases first, followed by all `positive` cases

pkg/router/controller/route_secret_manager_test.go
chiragkyal: Yes, this is the new route. 

The reason behind not having the old route spec is that we never pass the old route in the plugin chain while calling `HandleRoute()`. It’s only the new route that propagates in the plugin chain. So, the definition for the old route will be of no use in the test case scenarios also. 

https://github.com/openshift/router/blob/56ab14f257e123317c00ce7ba5b4ecae7b2e0102/pkg/router/interfaces.go#L13-L20

The only way to know whether the route was registered earlier or not, is to rely on `SecretManager`’s `IsRouteRegistered()` function. And so, we have `IsRegistered: true / false`  in the `fake.SecretManager` definition, to indicate that in the test cases.

So, if the old route was having `externalCertificate` , we will define 

```go
			secretManager: fake.SecretManager{
				IsRegistered: true,
			},
```

Similarly, when route is deleted.

pkg/router/controller/route_secret_manager_test.go
chiragkyal: Correct, thanks for the suggestion. Updated

pkg/router/controller/route_secret_manager_test.go
chiragkyal: Removed, thanks

pkg/router/controller/route_secret_manager_test.go
chiragkyal: Yes, it implements the Plugin interface.

https://github.com/openshift/router/blob/56ab14f257e123317c00ce7ba5b4ecae7b2e0102/pkg/router/interfaces.go#L13-L20

pkg/router/controller/route_secret_manager_test.go
chiragkyal: Okay, updated to `fakePluginDone`

pkg/router/controller/route_secret_manager_test.go
chiragkyal: Done

pkg/router/controller/route_secret_manager_test.go
chiragkyal: Updated

pkg/router/controller/route_secret_manager_test.go
chiragkyal: Because `oldSecret` will be same for all the test cases and doesn't fit well as an input scenario. I think we only care about the updated secret content and whether the `tls.Certificate` and `tls.Key` fields of the `expectedRoute` are getting updated correctly or not.

pkg/router/controller/route_secret_manager.go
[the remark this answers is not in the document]
Miciah: I agree that it's better to keep the `msg` variable if the message is used for both logging and status.  This is similar to this pattern that I referenced in https://github.com/openshift/router/pull/565#discussion_r1585305204: https://github.com/openshift/router/blob/b6f7c63c52e6b20fde2549a3c3e69a052ab453cc/pkg/router/controller/unique_host.go#L106-L119

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,290 @@
...
+		if p.secretManager.IsRouteRegistered(route.Namespace, route.Name) {
+			if err := p.secretManager.UnregisterRoute(route.Namespace, route.Name); err != nil {
+				klog.Error("failed to unregister route", err)
+				return err
+			}
+		}
+		// create new watch
+		if err := p.validateAndRegister(route); err != nil {
+			return err
+		}
```
Miciah: It does seem odd that the plugin *always* unregisters and reregisters the route, even if the TLS configuration did not change.  In particular, I'm concerned about the overhead of validating and reregistering the route for updates that don't touch the TLS configuration.  If that is necessary to do, can you add a comment explaining why?  If it isn't necessary, please either add a check to skip the unnecessary deregister-reregister or add a to-do comment to improve the logic before GA.

pkg/router/controller/route_secret_manager.go
```diff
@@ -0,0 +1,290 @@
...
+			Name:      tls.ExternalCertificate.Name,
+		}); err != nil {
+		errs = append(errs, field.Forbidden(fldPath, "router serviceaccount does not have permission to list this secret"))
+	}
+
+	// The secret should be in the same namespace as that of the route.
+	secret, err := secretsGetter.Secrets(route.Namespace).Get(context.TODO(), tls.ExternalCertificate.Name, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return append(errs, field.NotFound(fldPath, err))
```
Miciah:
```suggestion
			return append(errs, field.NotFound(fldPath, err.Error()))
```
Using `err.Error()` results in error messages like this:
```text
spec.tls.externalCertificate: Not found: "secrets \"tls-secret\" not found"
```
instead of this:
```text
spec.tls.externalCertificate: Not found: errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:"", RemainingItemCount:(*int64)(nil)}, Status:"Failure", Message:"secrets \"tls-secret\" not found", Reason:"NotFound", Details:(*v1.StatusDetails)(0xc0001a8c60), Code:404}}
```
(Thanks, @lihongan, for pointing out the overly verbose error message!)

pkg/router/controller/route_secret_manager.go
chiragkyal: I got the validation and reregistration overhead concern here. The plugin *always* needs to unregister and reregister the route even if the TLS configuration did not change, for two reasons

1.  The `HandleRoute()` method does not carry the old route spec; it's only the new route that propagates in the plugin chain. So, there isn't a definite way to compare the old TLS configuration and the new TLS configuration, and decide whether there is any change happened or not. Hence, we assume that it's *always* updated, so reregistration is required.
2. Another reason for always doing reregistration is to have an updated `secretHandler`. If we don't always create a new handler (with every route update event), there may be changes of serving stale route spec in the next plugin chain, when the referenced `secret` is updated or deleted. Sending outdated route in the next plugin, may impact their expected functionality.

https://github.com/openshift/router/blob/c9ce8871adb319e79839566fd483f7d763c17610/pkg/router/controller/route_secret_manager.go#L163-L207

chiragkyal: If we want to improve the logic for GA, then I think we could rely on `secretManager` to give the previously referenced secret name. It could have a new method called `GetSecretName()` to return the previously registered secret name. And we will only unregister and re-register if it's updated.

This could look something like

```go
	case watch.Modified:

		newHasExt := hasExternalCertificate(route)
		oldHadExt := p.secretManager.IsRouteRegistered(route.Namespace, route.Name)

		if newHasExt { // new route has externalCertificate
			if oldHadExt { // old route was also having externalCertificate
				oldSecret := p.secretManager.GetSecretName(route.Namespace, route.Name)
				if oldSecret != route.Spec.TLS.ExternalCertificate.Name { // externalCertificate is updated
					// unregister and re-register
					p.secretManager.UnregisterRoute(route.Namespace, route.Name)
					p.validateAndRegister(route)
				} else {
					// do nothing if externalCertificate is not updated
				}
			} else { // old route was not having externalCertificate
				p.validateAndRegister(route)
			}
		} else { // new route doesn't have externalCertificate
			if oldHadExt { // old route was having externalCertificate
				p.secretManager.UnregisterRoute(route.Namespace, route.Name)
			} else {
				// do nothing : no update to externalCertificate
			}
		}
```

This could solve the issue of *always* doing unregister and re-register. 

Along with that, we also need to ensure that the `secretHandler` always processes the updated route object whenever there is a secret update or delete event.

I think this can be archived by getting the updated route object from the client. So the `UpdateFunc` and `DeleteFunc` of `secretHandler` will first query the apiserver for the latest route, and then will continue the next steps. 

Let me know if this approach would improve the overhead issue. We can update it for GA.
If there are any other suggestions, we can discuss them too.

pkg/router/controller/route_secret_manager.go
chiragkyal: Thanks for the suggestion, updated to `err.Error()`

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1148 @@
...
+			Name:      name,
+			Namespace: namespace,
+		},
+		Data: data,
+		Type: secretType,
+	}
+}
+
+type fakePluginDone struct {
+	t      watch.EventType
```
rfredette, line 79: Can this be given a more descriptive name, such as `eventType`? Reading through this, I kept confusing it with a `t *testing.T` variable at first glance.

pkg/router/controller/route_secret_manager_test.go
alebedev87: > test will continue checking other expected values like expectedRejections. So, we need to define expectedRejections field as well.

Why checking for expected values if the tested function is expected to fail?

pkg/router/controller/route_secret_manager_test.go
alebedev87: Maybe a simple type check would be of help here:
```go
var _ Plugin = &fakePluginDone{}
```

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -0,0 +1,1148 @@
...
+
+	for _, s := range scenarios {
+		t.Run(s.name, func(t *testing.T) {
+			p := &fakePlugin{}
+			recorder := routeStatusRecorder{rejections: make(map[string]string)}
+
+			// assign default value to expectedRejections
+			if s.expectedRejections == nil {
+				s.expectedRejections = map[string]string{}
+			}
```
alebedev87, line 864: If you keep `rejections` and `expectedRejections` unset `reflect.DeepEqual()` should give `true`:
```suggestion
			recorder := routeStatusRecorder{}
```

chiragkyal: If we keep `rejections` unset, it will cause panic 
```
panic: assignment to entry in nil map
```
So, need to create the map.

pkg/router/controller/route_secret_manager_test.go
chiragkyal: because we will have to make sure that the status of the route is getting updated correctly, when this plugin rejects any route.

https://github.com/openshift/router/blob/d23f540103830840b033ea33e8967289212ad52f/pkg/router/controller/route_secret_manager.go#L115-L120

^^ status is getting updated before returing with error

pkg/router/controller/route_secret_manager.go
chiragkyal: I've added a comment in the code for now, with a to-do to improve the logic while going to GA.

Placeholder to track : https://issues.redhat.com/browse/CFE-1057

alebedev87:
/lgtm
/approve
/hold

Holding for @rfredette to have the final look.

chiragkyal:
/label acknowledge-critical-fixes-only

The payload jobs look okay to me. Additionally, this is a TechPreview feature behind feature-gate. I think we are good to add  `acknowledge-critical-fixes-only` label

PR #571, opened 2024-04-11
[WIP] OCPBUGS-29690: Count all service units, mitigating memory issues

frobware:
This commit changes the way HAProxy configurations are generated for routes with multiple service units.

Previously, the logic would exclude service units with a weight of zero. This leads to the router template encoding the "random" load-balancing algorithm, which significantly increases memory consumption when multiplied across many routes, and particularly when the associated weight is 256.

By including inactive service units in the count of total service units associated with a route, this change ensures that routes with alternate backends are considered active. When a backend has multiple service units the logic in the template router will encode "roundrobin" as the load-balancing algorithm instead of "random", mitigating the issue of increased memory consumption.

The adjustment is made by altering the calculation of ActiveServiceUnits to count all service units rather than just those that have a weight greater than zero.

This is a critical fix aimed at improving memory efficiency.

Fixes: https://issues.redhat.com/browse/OCPBUGS-29690

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -553,10 +553,10 @@ func (r *templateRouter) writeConfig() error {
 
 		// calculate the server weight for the endpoints in each service
 		// called here to make sure we have the actual number of endpoints.
-		cfg.ServiceUnitNames = r.calculateServiceWeights(cfg.ServiceUnits)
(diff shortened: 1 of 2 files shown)


frobware:
/hold
Don't merge this. This change will first target 4.16. I first want this change in 4.12 so that I can build a test image.

PR #572, opened 2024-04-12
[WIP] OCPBUGS-32044: add option idle-close-on-response

frobware:
WIP - exploratory change to see if we pass CI with the
idle-close-on-response option set.

Xref: https://github.com/haproxy/haproxy/issues/1735

PR #573, opened 2024-04-15
OCPBUGS-32044: Introduce 'idle-close-on-response' option for frontends

alebedev87:
This PR addresses a known issue ([OCPBUGS-32044](https://issues.redhat.com//browse/OCPBUGS-32044)) encountered with certain HTTP clients, notably the Apache HTTP client (prior to version 5), where closed idle connections are erroneously reused. To mitigate this, we introduce the `idle-close-on-response` option, which is now applied to all frontends operating in HTTP mode. Note that the `public_ssl` frontend remains in TCP mode and is unaffected by this change.

Spin off of https://github.com/openshift/router/pull/572.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -213,6 +213,12 @@ frontend public
   bind :{{ env "ROUTER_SERVICE_HTTP_PORT" "80" }}{{ if isTrue (env "ROUTER_USE_PROXY_PROTOCOL") }} accept-proxy{{ end }}
     {{- end }}
   mode http
+
+  # Workaround for a known issue encountered with certain HTTP clients,
+  # particularly the Apache HTTP client (prior to version 5),
+  # where closed idle connections are erroneously reused.
+  # Bug reference: https://issues.redhat.com/browse/OCPBUGS-32044.
+  option idle-close-on-response
   tcp-request inspect-delay {{ firstMatch $timeSpecPattern (env "ROUTER_INSPECT_DELAY") "5s" }}
   tcp-request content accept if HTTP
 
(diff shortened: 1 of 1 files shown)


alebedev87:
/payload-job  periodic-ci-openshift-release-master-nightly-4.16-e2e-aws-ovn-serial

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -213,6 +213,9 @@ frontend public
   bind :{{ env "ROUTER_SERVICE_HTTP_PORT" "80" }}{{ if isTrue (env "ROUTER_USE_PROXY_PROTOCOL") }} accept-proxy{{ end }}
     {{- end }}
   mode http
+
+  # https://issues.redhat.com/browse/OCPBUGS-32044
+  option idle-close-on-response
```
Miciah, line 218: If you provide a bare URL or bug number with no further information, then the reader has to look up the bug report.  If the reader has to do a lookup, then the reader could just as easily use git-blame to look up the commit message, which should explain the change and have a reference to the bug report.  Could you add more details to the comment, or just remove the comment entirely rely on the commit message to explain the change?
```suggestion
  # Work around an issue with Apache HttpClient 4.5's connection pooling manager,
  # which may attempt to re-use closed idle connections if this option is not enabled.
  option idle-close-on-response
```

alebedev87: I didn't give this diff enough of care yet. I just dropped the option to be able to test it. If this change will be the way to go I think I'll add some text as you proposed.

Btw the issue is not only with the Apache HTTP Client, initially the option was implemented for AWS because the ALBs kept sending requests to the closed (before idle ) connections (https://github.com/haproxy/haproxy/issues/1506).

alebedev87: The comment is updated. PTAL.

alebedev87:
Redundant comments about the option are removed.

Miciah:
Thanks!
/approve
/lgtm

PR #576, opened 2024-04-16
OCPBUGS-29690: Count active services before setting weight to 1

gcs278:
The previous logic, which reduced the weight to 1 when there was only one service, didn't filter out inactive services. If there's only one active service, there's no need for a weight greater than 1 because traffic is directed only to active services.

Additionally, the template logic correctly accounted for active services when setting the algorithm, resulting in "random" being set in situations where our logic didn't reduce the active service's weight to 1.

While this isn't inherently problematic, using the "random" algorithm with higher weights significantly increases memory usage on HaProxy startup. This can lead to excessive memory usage in scenarios where a user has many inactive services or routes backends using weight 0.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -1360,28 +1360,32 @@ func (r *templateRouter) getActiveEndpoints(serviceUnits map[ServiceUnitKey]int3
 // Each service gets (weight/sum_of_weights) fraction of the requests.
 // For each service, the requests are distributed among the endpoints.
 // Each endpoint gets weight/numberOfEndpoints portion of the requests.
-// The largest weight per endpoint is scaled to 256 to permit better
-// precision results.  The remainder are scaled using the same scale factor.
+// If there is more than one active service, the largest weight per endpoint
+// is scaled to 256 to permit better precision results.  The remainder are
+// scaled using the same scale factor. If there is only one active service,
+// then non-zero weights are configured with a weight of 1.
 // Inaccuracies occur when converting float32 to int32 and when the scaled
 // weight per endpoint is less than 1.0, the minimum.
 // The above assumes roundRobin scheduling.
 // The port parameter, if set, will only count endpoints matching that port.
 func (r *templateRouter) calculateServiceWeights(serviceUnits map[ServiceUnitKey]int32, port string) map[ServiceUnitKey]int32 {
 	serviceUnitNames := make(map[ServiceUnitKey]int32)
-
-	// If there is only 1 service unit, then always set the weight 1
-	// for all the endpoints, except when the service weight is 0.
-	// Scaling the weight to 256 is redundant and causes haproxy to allocate more memory on startup.
-	if len(serviceUnits) == 1 {
-		for key, weight := range serviceUnits {
-			if r.numberOfEndpoints(key, port) > 0 {
-				if weight == 0 {
-					serviceUnitNames[key] = 0
-				} else {
-					serviceUnitNames[key] = 1
-				}
+	// If there is only 1 active service unit, then always reduce the weight to 1
+	// for all the endpoints, except when the service weight is 0, or it contains no endpoints.
+	// Scaling the weight to 256 in this case is redundant and causes haproxy to allocate more
+	// memory on startup.
+	activeServiceUnits := 0
+	for key, weight := range serviceUnits {
+		if r.numberOfEndpoints(key, port) > 0 {
+			if weight > 0 {
+				activeServiceUnits++
+				serviceUnitNames[key] = 1
+			} else if weight == 0 {
+				serviceUnitNames[key] = 0
 			}
 		}
+	}
+	if activeServiceUnits == 1 {
 		return serviceUnitNames
 	}
 
--- a/pkg/router/template/router_test.go
+++ b/pkg/router/template/router_test.go
@@ -942,6 +942,7 @@ func TestFilterNamespaces(t *testing.T) {
 func TestCalculateServiceWeights(t *testing.T) {
 	suKey1 := ServiceUnitKey("ns/svc1")
 	suKey2 := ServiceUnitKey("ns/svc2")
+	suKey3 := ServiceUnitKey("ns/svc3")
 	ep1 := Endpoint{
 		ID:       "ep1",
 		IP:       "ip",
@@ -1215,7 +1216,7 @@ func TestCalculateServiceWeights(t *testing.T) {
 				suKey2: 0,
 			},
 			expectedWeights: map[ServiceUnitKey]int32{
-				suKey1: 256,
+				suKey1: 1,
 				suKey2: 0,
 			},
 		},
@@ -1233,6 +1234,57 @@ func TestCalculateServiceWeights(t *testing.T) {
 			expectedWeights: map[ServiceUnitKey]int32{
 				suKey1: 0,
 				suKey2: 0,
+			},
+		},
+		{
+			name:      "two services with weight 0",
+			routePort: "8080",
+			serviceUnits: map[ServiceUnitKey][]Endpoint{
+				suKey1: {ep1, ep2},
+				suKey2: {ep2},
+				suKey3: {ep3},
+			},
+			serviceWeights: map[ServiceUnitKey]int32{
+				suKey1: 0,
+				suKey2: 100,
+				suKey3: 0,
+			},
+			expectedWeights: map[ServiceUnitKey]int32{
+				suKey1: 0,
+				suKey2: 1,
+				suKey3: 0,
+			},
+		},
+		{
+			name:      "one service with no endpoints",
+			routePort: "8080",
+			serviceUnits: map[ServiceUnitKey][]Endpoint{
+				suKey1: {},
+				suKey2: {ep2},
+			},
+			serviceWeights: map[ServiceUnitKey]int32{
+				suKey1: 100,
+				suKey2: 100,
+			},
+			expectedWeights: map[ServiceUnitKey]int32{
+				suKey2: 1,
+			},
+		},
+		{
+			name:      "two services with no endpoints",
+			routePort: "8080",
+			serviceUnits: map[ServiceUnitKey][]Endpoint{
+				suKey1: {},
+				suKey2: {ep2},
+				suKey3: {},
+			},
+			serviceWeights: map[ServiceUnitKey]int32{
+				suKey1: 0,
+				suKey2: 100,
+				suKey3: 75,
+			},
+			expectedWeights: map[ServiceUnitKey]int32{
+				suKey2: 1,
 			},
 		},
 	}


pkg/router/template/router.go
```diff
@@ -1369,10 +1369,16 @@ func (r *templateRouter) getActiveEndpoints(serviceUnits map[ServiceUnitKey]int3
 func (r *templateRouter) calculateServiceWeights(serviceUnits map[ServiceUnitKey]int32, port string) map[ServiceUnitKey]int32 {
 	serviceUnitNames := make(map[ServiceUnitKey]int32)
 
-	// If there is only 1 service unit, then always set the weight 1
+	// If there is only 1 active service unit, then always set the weight 1
 	// for all the endpoints, except when the service weight is 0.
 	// Scaling the weight to 256 is redundant and causes haproxy to allocate more memory on startup.
```
candita, line 1374: If we don't actually do this scaling, can you update the pre-signature comment that says "The largest weight per endpoint is scaled to 256 to permit better precision results." ?

pkg/router/template/router.go
```diff
@@ -1369,10 +1369,16 @@ func (r *templateRouter) getActiveEndpoints(serviceUnits map[ServiceUnitKey]int3
...
-	if len(serviceUnits) == 1 {
+	activeServiceUnits := 0
+	for key, weight := range serviceUnits {
+		if weight > 0 && r.numberOfEndpoints(key, port) > 0 {
+			activeServiceUnits++
+		}
+	}
+	if activeServiceUnits == 1 {
```
candita, line 1381: Does this work the same, without counting active services ahead of time?

> for key, weight := range serviceUnits {
> 		if weight > 0 && r.numberOfEndpoints(key, port) > 0 {
> 			serviceUnitNames[key] = 1
> 		} else {
> 			serviceUnitNames[key] = 0
> 		}
> 	}

pkg/router/template/router.go
gcs278: Sure, I added details to that comment regarding having 1 active endpoint.

pkg/router/template/router.go
gcs278: It doesn't - you need to count the total number of active endpoints first to determine if you should force weight to be 1.

It's not clear if you meant to keep `if len(serviceUnits) == 1 {` in your suggested logic. I'll assume you meant this because otherwise your suggestion would set all active services to weight of 1, which would defeat the purpose of weight scaling entirely:
```
if len(serviceUnits) == 1 {
 		if weight > 0 && r.numberOfEndpoints(key, port) > 0 {
 			serviceUnitNames[key] = 1
 		} else {
 			serviceUnitNames[key] = 0
 		}
 	}
       return serviceUnitNames
}
```

This logic doesn't work for the scenarios where we have > 1 service, but only 1 active service. An example:
- ServiceA has weight 0
- ServiceB has weight 256

which needs to become:
- ServiceA stays weight 0
- ServiceB becomes weight 1

Total number of services = 2. Total number of active services = 1. The suggested logic doesn't trigger because total number of services = 2.

Do that make sense?

candita: Is there any way to do this without iterating over the service units twice?  Maybe keep a record of the key/s of active service/s?  If there is only one key and it passes the condition, then set it.

gcs278: Hmm. Yes - I made an improvement to not iterate twice, by creating a data structure to store the "reduced" weight at the same iteration that we count in.

Then, if our counter == 1, just return the data structure with the reduced values in it. 

Should be slightly more efficient. Let me know what you think.

frobware: > Should be slightly more efficient

I wonder how often the number of service units is above 10, 100, 1000? Is there a real-world optimisation here?

gcs278: We actually only allow a total of 4 services (1 `spec.to` and 3 `spec.alternativeBackends`).

However, my previous logic was calling `numberOfEndpoints` twice. It looks like the max number of endpoints if [1000](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#:~:text=By%20default%2C%20the%20control%20plane,to%20a%20maximum%20of%201000.).

So the last improvement went from 2 numberOfEndpoints x 4 services x 1000 endpoints to 4 services x 1000 endpoints. A minor improvement in the real-world.

Miciah: The extra allocation + GC is probably going to cost significantly more than looping twice.

Miciah:
If you have a route that specifies `spec.port` and `spec.alternateBackends` with one service, and the alternate backend service has a port that doesn't match `spec.port.to`, then do we erroneously use `roundrobin` instead of `random`?  Do we also need to modify `createServiceAliasConfig` to filter out service units without matching ports when calculating `activeServiceUnits`?

pkg/router/template/router.go
gcs278: Right. I didn't think about the time it takes to allocate the list.

I don't want to over complicate...but I feel obligated to provide data vs. just trying to guess at what is more efficient. So I ran an experiment. I change the unit test `TestCalculateServiceWeights` to run this function 10 million times for each unit test case:
```
	var endpointWeights map[ServiceUnitKey]int32
	for i := 1; i <= 10000000; i++ {
		endpointWeights = router.calculateServiceWeights(tc.serviceWeights, tc.routePort)
	}
```

**Findings with my current revision (2 makes 1 iteration) compared to my previous revision (1 make 2 iterations) :**
- In the cases that there is 1 active service, current revision is ~25% faster (due to not having to count endpoints again)
- In the cases that there > 1 active service, current revision ~10% slower (due to the extra allocation)

Disclaimer that these are plug-and-chug numbers. But I think the outcome is whether we favor scenarios where there is 1 active service or > 1 active service.

If I took a wild guess, I'd say our customers are using a 50 / 50 split between 1 active service and > 1 active service routes. I can't say that either one is more common than the other. Let me know if you disagree.

I'm going to leave it as it is currently to favor 1 active service. But feel free to disagree.

gcs278: Oh but I did move the `serviceUnitNames := make(map[ServiceUnitKey]int32)` to below the active service unit logic because that wasn't needed yet, and in the case of 1 active service unit, it's an unnecessary allocation.

https://github.com/openshift/router/compare/28515b833a28db1e14853ff0044190339b02212b..3875a5660ae1d96ecce0e7325f36f35d8d0b8f1b

pkg/router/template/router.go
candita: So, my concern was just that these two comments don't really agree with each other.
// If there is more than one active service, the largest weight per endpoint is scaled to 256 to permit better precision results.
// Scaling the weight to 256 is redundant and causes haproxy to allocate more memory on startup.

candita: Just changing this might do the trick?
// Scaling the weight to 256 **in this case** is redundant and causes haproxy to allocate more memory on startup.

pkg/router/template/router.go
candita: You can just keep the one map serviceUnitNames and use it again later.  in the subsequent code it doesn't look at the values that are already in serviceUnitNames, it just resets the value for each key in L1428.

gcs278: True, that does work. As long as that doesn't make the code too fragile, I suppose that safe to do. Done.

frobware: With all the micro optimisation effort ... whose dreams come true if we analyse at this level? How many milliseconds did we save (who cares)? Will anybody notice?

frobware:
Do we need to resolve https://github.com/openshift/router/pull/576#issuecomment-2065372101 in this PR too?

gcs278:
> Do we need to resolve [#576 (comment)](https://github.com/openshift/router/pull/576#issuecomment-2065372101) in this PR too?

I don't think so, seems like a separate bug as I don't think they have dependency between each other. In some situations with ports on service that don't match route's `spec.port.to`, you'll get `roundrobin` in the HaProxy config file with only 1 server line, when we actually want `random`. With #576, we scale the weights of 1 "active" endpoints to `weight 1`. If you only have 1 server line weight > 1 doesn't matter, regardless what the balance algorithm is (except that random uses a lot more memory...).

In other words, I don't see any harm in changing a single server backend from `weight 256` to `weight 1` with using `roundrobin`, it's a No-Op.

ShudiLi:
The  weights were changed to 1
```
1.
% oc get clusterversion
version   4.16.0-0.ci.test-2024-04-23-081430-ci-ln-fgiq082-latest   True        False         28m     Cluster version is 4.16.0-0.ci.test-2024-04-23-081430-ci-ln-fgiq082-latest

2.
% oc get route
NAME               HOST/PORT                                                                             PATH   SERVICES           PORT    TERMINATION   WILDCARD
reencrypt1         reencrypt1-default.apps.ci-ln-fgiq082-72292.origin-ci-int-gce.dev.rhcloud.com                service-secure     https   reencrypt     None
service-unsecure   service-unsecure-default.apps.ci-ln-fgiq082-72292.origin-ci-int-gce.dev.rhcloud.com          service-unsecure   http                  None

3.
% oc -n openshift-ingress rsh router-default-7dd68dd75f-jjtlb 
sh-5.1$ grep -Rn weight haproxy.config 
127:  server fe_sni unix@/var/lib/haproxy/run/haproxy-sni.sock weight 1 send-proxy
169:  server fe_no_sni unix@/var/lib/haproxy/run/haproxy-no-sni.sock weight 1 send-proxy
227:  server pod:web-server-rc-nnbf8:service-secure:https:10.131.0.14:8443 10.131.0.14:8443 cookie 3be72d2f6b0917638bbddc5a4d310a6a weight 1 ssl verifyhost service-secure.default.svc verify required ca-file /var/run/configmaps/service-ca/service-ca.crt
245:  server pod:web-server-rc-nnbf8:service-unsecure:http:10.131.0.14:8080 10.131.0.14:8080 cookie bf59e111822d7c46847171b9fe117bd4 weight 1
253:  server pod:oauth-openshift-6d759cf-ngtlk:oauth-openshift:https:10.128.0.60:6443 10.128.0.60:6443 weight 1 check inter 5000ms
254:  server pod:oauth-openshift-6d759cf-tx6fw:oauth-openshift:https:10.129.0.62:6443 10.129.0.62:6443 weight 1 check inter 5000ms
255:  server pod:oauth-openshift-6d759cf-tsxhm:oauth-openshift:https:10.130.0.76:6443 10.130.0.76:6443 weight 1 check inter 5000ms
274:  server pod:console-6967998f67-p56d7:console:https:10.128.0.64:8443 10.128.0.64:8443 cookie 8679ace67e2fa40ee6c30f71b8d6788f weight 1 ssl verifyhost console.openshift-console.svc verify required ca-file /var/run/configmaps/service-ca/service-ca.crt check inter 5000ms
275:  server pod:console-6967998f67-mr7c9:console:https:10.129.0.68:8443 10.129.0.68:8443 cookie 370fef5aeed56e1697df5ffc07859b86 weight 1 ssl verifyhost console.openshift-console.svc verify required ca-file /var/run/configmaps/service-ca/service-ca.crt check inter 5000ms
293:  server pod:downloads-cdb6f8949-6mgjk:downloads:http:10.128.0.42:8080 10.128.0.42:8080 cookie 9b7c18f36c60c4656fb144de59d5b2a3 weight 1 check inter 5000ms
294:  server pod:downloads-cdb6f8949-lhgck:downloads:http:10.130.0.62:8080 10.130.0.62:8080 cookie ac2f8ba61d894025df89059782bc1968 weight 1 check inter 5000ms
312:  server pod:ingress-canary-spdv5:ingress-canary:8080-tcp:10.128.2.11:8080 10.128.2.11:8080 cookie a275d1f255efa0180a7ea1f40e79c48c weight 1 check inter 5000ms
313:  server pod:ingress-canary-8wg9n:ingress-canary:8080-tcp:10.129.2.7:8080 10.129.2.7:8080 cookie 1dd0c54b43aa597e9c4a84dc786204ab weight 1 check inter 5000ms
314:  server pod:ingress-canary-qjbdz:ingress-canary:8080-tcp:10.131.0.8:8080 10.131.0.8:8080 cookie 401cc917bfb3e9a7b212f1406777f231 weight 1 check inter 5000ms
332:  server pod:alertmanager-main-0:alertmanager-main:web:10.128.2.22:9095 10.128.2.22:9095 cookie 359fec7c5c15177e0aec87fe172a0b9c weight 1 ssl verifyhost alertmanager-main.openshift-monitoring.svc verify required ca-file /var/run/configmaps/service-ca/service-ca.crt check inter 5000ms
333:  server pod:alertmanager-main-1:alertmanager-main:web:10.131.0.13:9095 10.131.0.13:9095 cookie 5dcb2e178c26b447649da3439ed7484b weight 1 ssl verifyhost alertmanager-main.openshift-monitoring.svc verify required ca-file /var/run/configmaps/service-ca/service-ca.crt check inter 5000ms
351:  server pod:prometheus-k8s-1:prometheus-k8s:web:10.128.2.21:9091 10.128.2.21:9091 cookie 9bcaae492092f3606e7ae8848a9fe3d7 weight 1 ssl verifyhost prometheus-k8s.openshift-monitoring.svc verify required ca-file /var/run/configmaps/service-ca/service-ca.crt check inter 5000ms
352:  server pod:prometheus-k8s-0:prometheus-k8s:web:10.129.2.12:9091 10.129.2.12:9091 cookie 9c5b8f6d0b9d1ff464d16c9a6f2f0a83 weight 1 ssl verifyhost prometheus-k8s.openshift-monitoring.svc verify required ca-file /var/run/configmaps/service-ca/service-ca.crt check inter 5000ms
370:  server pod:prometheus-k8s-1:prometheus-k8s:web:10.128.2.21:9091 10.128.2.21:9091 cookie 9bcaae492092f3606e7ae8848a9fe3d7 weight 1 ssl verifyhost prometheus-k8s.openshift-monitoring.svc verify required ca-file /var/run/configmaps/service-ca/service-ca.crt check inter 5000ms
371:  server pod:prometheus-k8s-0:prometheus-k8s:web:10.129.2.12:9091 10.129.2.12:9091 cookie 9c5b8f6d0b9d1ff464d16c9a6f2f0a83 weight 1 ssl verifyhost prometheus-k8s.openshift-monitoring.svc verify required ca-file /var/run/configmaps/service-ca/service-ca.crt check inter 5000ms
389:  server pod:thanos-querier-5c89b7859-b9dkj:thanos-querier:web:10.128.2.17:9091 10.128.2.17:9091 cookie 95776b0f70ae90779ad25efd8cc21ad5 weight 1 ssl verifyhost thanos-querier.openshift-monitoring.svc verify required ca-file /var/run/configmaps/service-ca/service-ca.crt check inter 5000ms
390:  server pod:thanos-querier-5c89b7859-dwsp5:thanos-querier:web:10.131.0.11:9091 10.131.0.11:9091 cookie 76f5757f419b4a722d6065673971b937 weight 1 ssl verifyhost thanos-querier.openshift-monitoring.svc verify required ca-file /var/run/configmaps/service-ca/service-ca.crt check inter 5000ms
```

ShudiLi:
/label qe-approved
thanks

frobware:
> The weight is 256 which is larger than the configured
> 
> ```
> backend be_http:e2e-test-router-env-n6zjv:service-unsecure01
>   mode http
>   option redispatch
>   option forwardfor
>   balance roundrobin
> 
>   timeout check 5000ms
>   http-request add-header X-Forwarded-Host %[req.hdr(host)]
>   http-request add-header X-Forwarded-Port %[dst_port]
>   http-request add-header X-Forwarded-Proto http if !{ ssl_fc }
>   http-request add-header X-Forwarded-Proto https if { ssl_fc }
>   http-request add-header X-Forwarded-Proto-Version h2 if { ssl_fc_alpn -i h2 }
>   http-request add-header Forwarded for=%[src];host=%[req.hdr(host)];proto=%[req.hdr(X-Forwarded-Proto)]
>   cookie e393e394aa38b3591557660f1e226508 insert indirect nocache httponly
>   server pod:web-server-rc01-dll2p:service-unsecure01:http:10.131.0.27:8080 10.131.0.27:8080 cookie 0ea43df433f7c28f54614f5e49bbee26 weight 256 check inter 5000ms
>   server pod:web-server-rc02-pc96q:service-unsecure02:http:10.131.0.28:8080 10.131.0.28:8080 cookie 1edbf08e519de047bb6e49c7901b78aa weight 256 check inter 5000ms
>   server pod:web-server-rc03-xxjcg:service-unsecure03:http:10.131.0.29:8080 10.131.0.29:8080 cookie 2b78972471dc2752949b8fa0be506378 weight 256 check inter 5000ms
> ```

For roundobin this is OK - no huge memory growth comes from a combination of roundrobin x 256.

frobware:
> /approve /hold I want to get some confirmation from the customer that is running the [4.12 proposed change](https://github.com/openshift/router/pull/581)

We had confirmation that the 4.12 PR works for the customer. Removing the /hold.

/hold cancel

PR #578, opened 2024-04-18
[WIP] [release-4.12] OCPBUGS-29690: Count active services before setting weight to 1

frobware:
**DO NOT MERGE**

Temporary PR in release-4.12 so I can build an image via cluster-bot.

This is taken from: https://github.com/openshift/router/pull/576.

The previous logic, which reduced the weight to 1 when there was only one service, didn't filter out inactive services. If there's only one active service, there's no need for a weight greater than 1 because traffic is directed only to active services.

Additionally, the template logic correctly accounted for active services when setting the algorithm, resulting in "random" being set in situations where our logic didn't reduce the active service's weight to 1.

While this isn't inherently problematic, using the "random" algorithm with higher weights significantly increases memory usage on HaProxy startup. This can lead to excessive memory usage in scenarios where a user has many inactive services or routes backends using weight 0.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -553,10 +553,10 @@ func (r *templateRouter) writeConfig() error {
 
 		// calculate the server weight for the endpoints in each service
 		// called here to make sure we have the actual number of endpoints.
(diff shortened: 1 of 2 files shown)


frobware:
Closing in favour of #581.

PR #581, opened 2024-04-18
[release-4.12] OCPBUGS-33517: Count active services before setting weight to 1

gcs278:
**This is a manual cherry-pick of https://github.com/openshift/router/pull/576**

**Merge Conflict Resolution:**

- A merge conflict occurred because https://github.com/openshift/router/pull/439 is not backported to 4.12
  - We added the port parameter in  `numberOfEndpoints(key, port)` so it counts endpoints only for the port that the router is using.
  - Without backporting https://github.com/openshift/router/pull/439, it will double count the endpoints if you have 2 ports.
  - Our backport logic, we now additionally leverage `r.numberOfEndpoints(key)` to see if it's an active endpoint, i.e. if endpoints are > 0. `port` only matters if there are > 0 endpoints. Filtering by `port` might cause > 0 endpoint situations to change, but when there are 0 endpoints, number of ports has no impact.
  - **Summary**: the port counting fix doesn't impact the use of `r.numberOfEndpoints(key)` in our backport as we are only concerned with whether it is zero or non-zero. Additionally, filtering by port only influences situations where endpoints are greater than 0. ([Slack Discussion](https://redhat-internal.slack.com/archives/CCH60A77E/p1713458411878909?thread_ts=1709895019.274599&cid=CCH60A77E))

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -1296,28 +1296,32 @@ func (r *templateRouter) getActiveEndpoints(serviceUnits map[ServiceUnitKey]int3
 // Each service gets (weight/sum_of_weights) fraction of the requests.
 // For each service, the requests are distributed among the endpoints.
 // Each endpoint gets weight/numberOfEndpoints portion of the requests.
-// The largest weight per endpoint is scaled to 256 to permit better
-// percision results.  The remainder are scaled using the same scale factor.
+// If there is more than one active service, the largest weight per endpoint
+// is scaled to 256 to permit better precision results.  The remainder are
+// scaled using the same scale factor. If there is only one active service,
+// then non-zero weights are configured with a weight of 1.
 // Inaccuracies occur when converting float32 to int32 and when the scaled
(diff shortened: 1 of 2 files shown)


frobware:
/retitle "[WIP] Count active services before setting weight to 1". 

We will retitle again once we have a bug number for 4.12.

frobware:
/retitle "[release-4.12] OCPBUGS-33517: Count active services before setting weight to 1"

frobware:
/lgtm
/approve

This is the commit that the customer is currently running in their pre-prod cluster; we built them an image based on this PR.

/hold

Adding a hold; @gcs278 please could you correct the PR description as we want to go ahead and merge this now.

frobware:
> This is the commit that the customer is currently running in their pre-prod cluster; we built them an image based on this PR.

Slack discussion: https://redhat-internal.slack.com/archives/CCH60A77E/p1713364810748119?thread_ts=1709895019.274599&cid=CCH60A77E.

PR #587, opened 2024-04-25
OCPBUGS-26498: Optimize Upgrade Validation plugin by skipping unnecessary changes

gcs278:
Both performIngressConditionUpdate and performIngressConditionRemoval functions add tasks to the writerlease queue even if no work needed to be done. This commit optimizes the Upgrade Validation plugin by ensuring that tasks for updating UnservableInFutureVersions are only added to the queue when changes are required.

This reduction in unnecessary work significantly lowers the frequency of lease extensions. Previously, excessive lease extensions could delay route status updates under certain conditions, such as during temporary contention periods
where a router pod gets demoted to a follower. After the contention is resolved, the pod’s subsequent retry will be delayed more than necessary.

Additionally, this update provides a clearer indication of when updates are actually required, but have been completed by another router pod. The prior logic did not clearly distinguish between unnecessary updates and updates that were completed by another pod.

The selective updates are only added to the interface used by Upgrade Validation plugin to avoid perturbing existing Admitted condition logic. This means nominal clusters without SHA1 routes should not be impacted by the Upgrade
Validation plugin, minimizing the risk associated with its introduction.

Fixes [TestRouteAdmissionPolicy Flake](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-ingress-operator/1049/pull-ci-openshift-cluster-ingress-operator-master-e2e-gcp-operator/1783251298133479424) by reducing the number of unnecessary lease extensions.

### DIFF
--- a/pkg/router/controller/status.go
+++ b/pkg/router/controller/status.go
@@ -27,6 +27,12 @@ const (
 	unservableInFutureVersionsClearAction = string(routev1.RouteUnservableInFutureVersions) + "-Clear"
 )
 
+// TODO: Refactor the router to perform a single status update per route reconciliation.
+// Currently, executing multiple status updates simultaneously can complicate the writer lease and
+// contention tracker logic, increasing the likelihood of write conflicts and contentions. While the
+// existing logic is eventually consistent and acceptable for infrequent scenarios, it may lead to
+// prolonged status updates and complicate the debugging process.
+
 // RouteStatusRecorder is an object capable of recording why a route status condition changed.
 type RouteStatusRecorder interface {
 	RecordRouteRejection(route *routev1.Route, reason, message string)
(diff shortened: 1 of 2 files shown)


gcs278:
`TestRouteAdmissionPolicy` is still flaking, and this may be a reasonable fix.
/reopen

pkg/router/controller/status.go
```diff
@@ -133,16 +139,39 @@ func (a *StatusAdmitter) RecordRouteRejection(route *routev1.Route, reason, mess
...
 		Reason:  reason,
 		Message: message,
-	})
+	}
+	// First, verify if updates are required by checking if the route ingress status matches expected values.
+	// This helps avoid unnecessary tasks in the writerlease queue and prevents unneeded lease extensions.
+	// We only do this in for the new UnservableInFutureVersions condition to avoid perturbing existing logic for the
+	// Admitted condition. This should be reevaluated when refactoring to use a single route status update per route
+	// reconciliation (see TODO at top of file).
+	changed, _, _, _, _ := recordIngressCondition(route.DeepCopy(), a.routerName, a.routerCanonicalHostname, expectedCondition)
```
Miciah, line 153: Please add a comment to note that `route.DeepCopy()` is necessary because `recordIngressCondition` actually modifies `route`.

Miciah: Consider refactoring to use `findCondition` instead of `recordIngressCondition` to save some overhead from deepcopies etc.

gcs278: Done.

pkg/router/controller/status.go
```diff
@@ -328,6 +357,36 @@ func recordIngressCondition(route *routev1.Route, name, hostName string, conditi
...
+		if existing.RouterName != name {
+			continue
+		}
+
+		changed := existing.Host != route.Spec.Host ||
+			existing.WildcardPolicy != route.Spec.WildcardPolicy ||
+			existing.RouterCanonicalHostname != hostName
+		if changed {
+			return true
+		}
```
Miciah, line 374: Reading this and the unit test makes me realize that both the StatusAdmitter plugin and the UpgradeValidation plugin update the status ingress entry in case one of these fields (`Host`, `WildcardPolicy`, or `RouterCanonicalHostname`) needs to be updated.  

Suppose the route has a host name that conflicts with another route's, and then the route is updated to have a non-conflicting host name.  UpgradeValidation runs first in the plugin chain, so it updates the status ingress entry first, and the StatusAdmitter plugin updates the status condition second.  Do I have that right?  

Would it be safer for `RecordRouteUnservableInFutureVersions` to ignore changes to the status ingress entry other than the UnservableInFutureVersions status condition?

gcs278: > Suppose the route has a host name that conflicts with another route's, and then the route is updated to have a non-conflicting host name. UpgradeValidation runs first in the plugin chain, so it updates the status ingress entry first, and the StatusAdmitter plugin updates the status condition second. Do I have that right?

Correct. Both UpgradeValidation and StatusAdmitter try to update the new host name as part of the ingress fields. 

> Would it be safer for RecordRouteUnservableInFutureVersions to ignore changes to the status ingress entry other than the UnservableInFutureVersions status condition?

Hmmm. Yea it might be *slightly* safer. If the route is also going from Rejected to Admitted, you'll have this sequence:
1. Route is updated to fix hostname
2. UpgradeValidation updates status to fix hostname
3. StatusAdmitter updates status from Admitted=False to Admitted=True

If we limit UpgradeValidation to just changes in UnservableInFutureVersions, we should only get one status update:
1. Route is updated to fix hostname
2. UpgradeValidation does nothing since it only cares about UnservableInFutureVersions
3. StatusAdmitter updates status to fix hostname and sets Admitted=False to Admitted=True

So yes, I agree. I can update it so we only care about conditions changes. That way, if it's just a hostname change, the UpgradeValidation plugin will let StatusAdmitter take care of that.

Miciah:
Thanks!
/approve
/lgtm

gcs278:
I've ran nearly a hundred times with my e2e test in https://github.com/openshift/origin/pull/28710 and no issues. Reviewed the router logs and lease extends are drastically reduced.
/hold cancel

PR #589, opened 2024-04-30
OCPBUGS-29373: generateHAProxyCertConfigMap: No H2 with dup certs

Miciah:
Add a check in `generateHAProxyCertConfigMap` for whether the same certificate is specified on more than one route, or whether a route specifies the default certificate as a custom certificate.  If either of these conditions is detected, then configure HAProxy not to allow HTTP/2 client connections to this route.

Using the same certificate for multiple routes signals to Web browsers that they can do HTTP/2 connection coalescing.  This means that a browser could open a connection for one route and then re-use this connection to connect to a second route, which can cause the client's request to go to the wrong backend server.  We try to prevent this client behavior by advertising HTTP/2 in TLS ALPN only for routes that have unique certificates.

In general, we assume that if a route specifies a custom certificate, then it is a unique certificate.  However, this assumption is not necessarily valid; indeed, it is possible to specify the same custom certificate on multiple routes, or specify the default certificate as an explicit certificate on a route.

After this commit, OpenShift router detects the cases where multiple routes have the same certificate, or where a route specifies the default certificate; in these cases, we prohibit negotiating HTTP/2 using ALPN for those routes.  This should prevent connection coalescing in this situation.

### DIFF
--- a/pkg/router/router_test.go
+++ b/pkg/router/router_test.go
@@ -102,6 +102,7 @@ func TestMain(m *testing.M) {
 	h.workdir = workdir
 	h.dirs = map[string]string{
 		"whitelist": filepath.Join(workdir, "router", "whitelists"),
+		"certs":     filepath.Join(workdir, "router", "certs"),
 	}
 
 	createRouterDirs()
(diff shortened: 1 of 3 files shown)


ShudiLi:
/label qe-approved
thanks

pkg/router/router_test.go
```diff
@@ -685,6 +835,31 @@ func (m mustMatchConfig) Match(parser haproxyconfparser.Parser) error {
...
+	fileName := filepath.Join(h.workdir, "conf", mapFileName)
+
+	content, err := ioutil.ReadFile(fileName)
+	if err != nil {
+		return err
+	}
+	contains := false
+	for _, line := range strings.Split(string(content), "\n") {
+		if line == entry {
+			contains = true
```
frobware, line 848: Can we break once `contains` is true?

frobware (commented):
LGTM, just one question about `break`ing in a loop.

pkg/router/router_test.go
Miciah: It's just test code, but I can add a `break`.

Miciah:
https://github.com/openshift/origin/pull/28757 should fix the e2e-agnostic failure.

Miciah:
/hold
Waiting for https://github.com/openshift/origin/pull/28757.

frobware:
> /hold Waiting for [openshift/origin#28757](https://github.com/openshift/origin/pull/28757).

28757 has merged.
/hold cancel

pkg/router/router_test.go
Miciah: Fixed in https://github.com/openshift/router/compare/f86a9ba57bfa08239f396d8bf1fc597c332c1c62..0bc6f5ecf4c52169ca2a802cede164ac0a46f138.

Miciah:
https://github.com/openshift/router/compare/f86a9ba57bfa08239f396d8bf1fc597c332c1c62..0bc6f5ecf4c52169ca2a802cede164ac0a46f138 adds an aesthetic `break` statement in a test and fixes gofmt (I hope).

PR #590, opened 2024-05-01
[WIP] OCPBUGS-26498: Remove lease.Extend from performIngressConditionRemoval

gcs278:
Since performIngressConditionUpdate already extends the lease, we should not extend the lease in performIngressConditionRemoval. The lease gets over extended because the UpgradeValidation plugin and the Status plugin both attempt to update route status in two separate lease work items and both are extending the lease when there is no work to be done.

In a scenario where the Status plugin needs to admit a route, but the UpgradeValidation has no change to UnservableInFutureVersions condition, the UpgradeValidation lease work item will unnecessarily extend the lease, leading to an extra delay in the Status plugin admitted the route.

Also add a TODO item to document current limitations of doing multiple status updates in the same route reconcile loop for future developers.

### DIFF
--- a/pkg/router/controller/status.go
+++ b/pkg/router/controller/status.go
@@ -27,6 +27,12 @@ const (
 	unservableInFutureVersionsClearAction = string(routev1.RouteUnservableInFutureVersions) + "-Clear"
 )
 
+// TODO: Refactor the router to perform a single status update per route reconciliation.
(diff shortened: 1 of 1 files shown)


gcs278:
Closing in favor of https://github.com/openshift/router/pull/587
/close

PR #595, opened 2024-05-13
OCPBUGS-29894: Check if CRLs are downloaded when determining ready status

rfredette:
Require all CRLs to be downloaded before the router can report that it's ready. This prevents forwarding requests to a router until it's ready to handle mTLS.

This fixes OCPBUGS-29894

### DIFF
--- a/pkg/router/crl/crl.go
+++ b/pkg/router/crl/crl.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	logf "github.com/openshift/router/log"
@@ -66,6 +67,11 @@ var (
 	CRLFilename = filepath.Join(mtlsLatestSymlink, crlBasename)
 	// CABundleFilename is the fully qualified path to the currently in use CA bundle.
 	CABundleFilename = filepath.Join(mtlsLatestSymlink, caBundleBasename)
+	// crlsUpdated is true when all CRLs have been successfully updated, and false when there are missing CRLs.
+	// You must take crlsMutex before using crlsUpdated.
+	crlsUpdated = false
+	// crlsMutex protects crlsUpdated.
+	crlsMutex = sync.Mutex{}
 )
 
 // authorityKeyIdentifier is a certificate's authority key identifier.
@@ -143,6 +149,9 @@ func ManageCRLs(caBundleFilename string, caUpdateChannel <-chan struct{}, update
 			log.Error(err, "failed to parse CA bundle", "CA bundle filename", caBundleFilename)
 			nextUpdate = time.Now().Add(errorBackoffTime)
 		}
+		if !shouldHaveCRLs {
+			SetCRLsUpdated(true)
+		}
 		for {
 			updated := false
 			if nextUpdate.IsZero() {
@@ -175,8 +184,9 @@ func ManageCRLs(caBundleFilename string, caUpdateChannel <-chan struct{}, update
 				nextUpdate = time.Now().Add(errorBackoffTime)
 				continue
 			}
-			// After successfully updating the CRL file, reset caUpdated
+			// After successfully updating the CRL file, reset caUpdated and mark CRLs as updated
 			caUpdated = false
+			SetCRLsUpdated(true)
 			if updated {
 				updateCallback(shouldHaveCRLs)
 			}
@@ -505,4 +515,18 @@ func makeStagingDirectory() (string, error) {
 		return "", err
 	}
 	return stagingDirName, nil
+}
+
+// GetCRLsUpdated returns true if all necessary CRLs have been downloaded.
+func GetCRLsUpdated() bool {
+	crlsMutex.Lock()
+	defer crlsMutex.Unlock()
+	return crlsUpdated
+}
+
+// SetCRLsUpdated sets crlsUpdated to the specified value, indicating whether all necessary CRLs have been downloaded.
+func SetCRLsUpdated(value bool) {
+	crlsMutex.Lock()
+	defer crlsMutex.Unlock()
+	crlsUpdated = value
 }
--- a/pkg/router/crl/crl.go
+++ b/pkg/router/crl/crl.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	logf "github.com/openshift/router/log"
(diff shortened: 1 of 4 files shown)


pkg/router/crl/crl.go
```diff
@@ -66,6 +67,9 @@ var (
 	CRLFilename = filepath.Join(mtlsLatestSymlink, crlBasename)
 	// CABundleFilename is the fully qualified path to the currently in use CA bundle.
 	CABundleFilename = filepath.Join(mtlsLatestSymlink, caBundleBasename)
+	// crlsUpdated is true when all CRLs have been successfully updated, and false when there are missing CRLs.
+	crlsUpdated = false
+	crlsMutex   = sync.Mutex{}
```
Miciah, line 72: It's better to be very explicit about mutexes.  
```suggestion
	// crlsUpdated is true when all CRLs have been successfully updated, and false when there are missing CRLs.
	// You must take crlsMutex before using crlsUpdated.  
	crlsUpdated = false
	// crlsMutex protects crlsUpdated.  
	crlsMutex = sync.Mutex{}
```

Miciah (commented):
This makes a router pod start failing readiness checks if it has outdated CRLs, right?  

To fix OCPBUGS-29894, it should be sufficient to fail readiness only for the initial synch, so that startup probes (which use the readiness endpoint) fail until the initial synch is done.  

Once the router pod has done the initial synch, we want readiness checks to pass even if refresh fails, for two reasons:

- The expectation is to restore the behavior prior to https://github.com/openshift/cluster-ingress-operator/pull/939 and https://github.com/openshift/router/pull/472, and that behavior was to prevent a router pod from serving traffic until it had CRLs, not to prevent a router pod from serving traffic if it had outdated CRLs.  
- It is generally less bad to continue using outdated CRLs, rather than to stop serving traffic entirely when refresh fails.  

This does make me realize that we need a Prometheus metric and an alert when refresh fails for a prolonged period.  Failure to refresh has two nasty implications:

- Router pods are using outdated CRLs.
- The next rolling update of the router deployment (for an upgrade, configuration change, or whatever reason) could get stuck as presumably the new pods would fail on initial synch.

rfredette:
e2e-upgrade failed during bootstrap.

/test e2e-upgrade

pkg/router/crl/crl.go
```diff
@@ -506,3 +516,15 @@ func makeStagingDirectory() (string, error) {
...
 	return stagingDirName, nil
 }
+
+func GetCRLsUpdated() bool {
+	crlsMutex.Lock()
+	defer crlsMutex.Unlock()
+	return crlsUpdated
+}
+
+func SetCRLsUpdated(value bool) {
```
alebedev87, line 526: Would it make sense to remove the possibility to set updated to `false`? Taking into account the fact that we want the CRLs to be fully present only at startup.
```suggestion
func SetCRLsUpdated() {
```

alebedev87 (commented):
LGTM, just a nit question.

pkg/router/crl/crl.go
rfredette: I think that's reasonable. I've updated this to include that change :+1:

Miciah:
@coderabbitai, please review!

pkg/router/template/router.go
```diff
@@ -502,6 +480,8 @@ func (r *templateRouter) watchMutualTLSCert() error {
 			log.V(0).Error(err, "failed to establish watch on mTLS certificate directory")
 			return nil
 		}
+	} else {
+		crl.SetCRLsUpdated(true)
 	}
```
Miciah: It might improve readability if you flipped the condition and swapped the `if` and `else` blocks to have the shorter block first:
```go
	if len(caPath) == 0 {
		// There are no CRLs to update, so report that everything is up to date.
		crl.SetCRLsUpdated(true)
	} else {
		r.haveClientCA = true
		if err := crl.InitMTLSDirectory(caPath); err != nil {
			return err
		}
		haveCRLs, err := crl.CABundleHasCRLs(caPath)
		if err != nil {
			log.V(0).Error(err, "failed to parse CA Bundle", "path", caPath)
			return err
		}
		r.haveCRLs = haveCRLs
		caUpdateChannel := make(chan struct{})
		crlReloadFn := func(haveCRLs bool) {
			r.haveCRLs = haveCRLs
			log.V(0).Info("reloading to get updated client CA CRL", "name", crl.CRLFilename, "have CRLs", haveCRLs)
			r.rateLimitedCommitFunction.RegisterChange()
		}
		crl.ManageCRLs(caPath, caUpdateChannel, crlReloadFn)
		caReloadFn := func() {
			// Send signal to CRL management goroutine that client CA has been changed
			caUpdateChannel <- struct{}{}
		}
		if err := r.watchVolumeMountDir(filepath.Dir(caPath), caReloadFn); err != nil {
			log.V(0).Error(err, "failed to establish watch on mTLS certificate directory")
			return nil
		}
	}
```

pkg/router/crl/crl.go
```diff
@@ -143,6 +149,9 @@ func ManageCRLs(caBundleFilename string, caUpdateChannel <-chan struct{}, update
 			log.Error(err, "failed to parse CA bundle", "CA bundle filename", caBundleFilename)
 			nextUpdate = time.Now().Add(errorBackoffTime)
 		}
+		if !shouldHaveCRLs {
+			SetCRLsUpdated(true)
+		}
```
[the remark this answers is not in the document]
Miciah: @rfredette, this seems like a legitimate concern.  Is there a reason it is better to fail open?  If so, please add a code comment; otherwise, we should take CodeRabbit's suggestion.

pkg/router/crl/crl.go
Miciah: Did Andrey's proposed change get dropped (or never implemented)?

pkg/router/crl/crl.go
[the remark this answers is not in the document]
rfredette: ack, this is a real concern. Fixed!

pkg/router/template/router.go
rfredette: After flipping the conditions, I realized you can also unwrap the larger part of the conditional (now the `else` block). Should be more readable in the latest update.

rfredette:
@coderabbitai please review!

pkg/router/template/router.go
```diff
@@ -454,32 +454,36 @@ func (r *templateRouter) writeDefaultCert() error {
...
+		r.rateLimitedCommitFunction.RegisterChange()
+	}
+	crl.ManageCRLs(caPath, caUpdateChannel, crlReloadFn)
+	caReloadFn := func() {
+		// Send signal to CRL management goroutine that client CA has been changed
+		caUpdateChannel <- struct{}{}
+	}
+	if err := r.watchVolumeMountDir(filepath.Dir(caPath), caReloadFn); err != nil {
+		log.V(0).Error(err, "failed to establish watch on mTLS certificate directory")
+		return nil
```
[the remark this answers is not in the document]
Miciah: @coderabbitai, is this change directly related to the original defect?  If it is a fix for a separate defect, it should go in a separate commit or PR that explains what it is fixing.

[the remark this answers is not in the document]
Miciah: Thanks!  @rfredette, would you mind putting this change in a separate commit to make it clear which code change addresses which issue?

PR #607, opened 2024-05-29
[WIP] OCPBUGS-27741: Properly handle route path while rewriting

gcs278:
Process the route's spec.path while using the
`haproxy.router.openshift.io/rewrite-target` annotation to prevent HAProxy configuration issues.

Previously, values in the route's path would be interpreted as regex meta characters. Now, this fix forces literal interpretation of the characters in spec.path when matching for rewriting. This change introduces some unavoidable incompatibilities if users were previously using regex meta characters in a spec.path.

Use cases have been documented here: https://docs.google.com/document/d/1wA6ZdRYaiN70TY5UbTOn1wl5qAn_NPb2ty_G85XhnYM/edit#heading=h.pca9p2i57r2u

### DIFF
--- a/pkg/router/template/util/rewritetarget/rewritetarget_test.go
+++ b/pkg/router/template/util/rewritetarget/rewritetarget_test.go
@@ -5,7 +5,7 @@ import (
 	"testing"
 )
 
-func Test_SanitizeInput(t *testing.T) {
+func Test_SanitizeRewriteTargetInput(t *testing.T) {
 	testCases := []struct {
 		name   string
 		input  string
@@ -85,7 +85,55 @@ func Test_SanitizeInput(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			got := rewritetarget.SanitizeInput(tc.input)
+			got := rewritetarget.SanitizeRewriteTargetInput(tc.input)
+			if got != tc.output {
+				t.Errorf("Failure: expected %s, got %s", tc.output, got)
+			}
+		})
+	}
+}
+
+func Test_SanitizeRewritePathInput(t *testing.T) {
+	testCases := []struct {
+		name   string
+		input  string
+		output string
+	}{
+		{
+			name:   "normal path",
+			input:  `/foo/bar`,
+			output: `/foo/bar`,
+		},
+		{
+			name:   "single quotes should be escaped",
+			input:  `'foo'foo\'`,
+			output: `'\''foo'\''foo\\'\''`,
+		},
+		{
+			name:   "number sign should NOT be escaped",
+			input:  `/foo/bar#`,
+			output: `/foo/bar#`,
+		},
+		{
+			name:   "backslashes should be escaped",
+			input:  `\foo\\foo\\\foo`,
+			output: `\\foo\\\\foo\\\\\\foo`,
+		},
+		{
+			name:   "special characters should be escaped",
+			input:  `/foo\.+*?()|[]{}^$`,
+			output: `/foo\\\.\+\*\?\(\)\|\[\]\{\}\^\$`,
+		},
+		{
+			name:   "new line characters should be escaped",
+			input:  "/foo\r\n",
+			output: `/foo\r\n`,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := rewritetarget.SanitizeRewritePathInput(tc.input)
 			if got != tc.output {
 				t.Errorf("Failure: expected %s, got %s", tc.output, got)
 			}
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -683,9 +683,9 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
         {{- with $pathRewriteTarget := firstMatch $pathRewriteTargetPattern (index $cfg.Annotations "haproxy.router.openshift.io/rewrite-target") }}
   # Path rewrite target
           {{- if eq $pathRewriteTarget "/" }}
-  http-request replace-path ^{{ $cfg.Path }}/?(.*)$ '{{ processRewriteTarget $pathRewriteTarget }}'
+  http-request replace-path '^{{ processRewritePath $cfg.Path }}/?(.*)$' '{{ processRewriteTarget $pathRewriteTarget }}'
           {{- else }}
-  http-request replace-path ^{{ $cfg.Path }}(.*)$ '{{ processRewriteTarget $pathRewriteTarget }}'
+  http-request replace-path '^{{ processRewritePath $cfg.Path }}(.*)$' '{{ processRewriteTarget $pathRewriteTarget }}'
           {{- end }}
         {{- end }}{{/* rewrite target */}}
   
--- a/pkg/router/template/template_helper.go
+++ b/pkg/router/template/template_helper.go
@@ -415,5 +415,6 @@ var helperFunctions = template.FuncMap{
 	"clipHAProxyTimeoutValue": clipHAProxyTimeoutValue, //clips extrodinarily high timeout values to be below the maximum allowed timeout value
 	"parseIPList":             parseIPList,             //parses the list of IPs/CIDRs (IPv4/IPv6)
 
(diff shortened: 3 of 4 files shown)


ShudiLi:
/label qe-approved

tested it with 4.16.0-0.ci.test-2024-05-31-012256-ci-ln-982w7yk-latest
```
1.
% oc get clusterversion
version   4.16.0-0.ci.test-2024-05-31-012256-ci-ln-982w7yk-latest   True        False         55m     Cluster version is 4.16.0-0.ci.test-2024-05-31-012256-ci-ln-982w7yk-latest

2.
% oc -n openshift-ingress get pods
router-default-7fd8bdbb55-ffvj9   1/1     Running   0          59m
router-default-7fd8bdbb55-g8cl9   1/1     Running   0          59m

3.
% oc -n openshift-ingress rsh router-default-7fd8bdbb55-ffvj9       
sh-5.1$ 
sh-5.1$ grep "http-request replace-path" haproxy         
haproxy-config.template  haproxy.config           
sh-5.1$ grep "http-request replace-path" haproxy-config.template 
  http-request replace-path '^\Q{{ processRewritePath $cfg.Path }}/\E?(.*)$' '{{ processRewriteTarget $pathRewriteTarget }}'
  http-

4.
% oc create route edge edge1 --service=unsec-server3               
route.route.openshift.io/edge1 created

5.
% oc patch route edge1 --type='merge' -p '{"spec":{"path":"/path1/#"}}' 
route.route.openshift.io/edge1 patched

6.
% oc annotate route edge1 "haproxy.router.openshift.io/rewrite-target"='/path/app/#'
route.route.openshift.io/edge1 annotate

7.
% oc get route edge1 -oyaml | grep -E "haproxy.router.openshift.io|path:" 
    haproxy.router.openshift.io/rewrite-target: /path/app/#
  path: /path1/#

8.
% oc -n openshift-ingress logs router-default-7fd8bdbb55-ffvj9 --tail=20
I0531 01:47:59.063258       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:48:06.009948       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:48:29.568934       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:48:34.564661       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:48:44.395788       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:48:49.372038       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:48:55.357546       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:49:00.353958       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:49:43.442301       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:49:48.410537       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:50:03.750952       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:50:08.705220       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:50:16.114628       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:50:44.912551       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:50:54.981343       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:51:06.963978       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 01:51:11.949069       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 02:23:15.068772       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 02:31:19.633735       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"
I0531 02:37:44.769399       1 router.go:669] "msg"="router reloaded" "logger"="template" "output"=" - Checking http://localhost:80 ...\n - Health check ok : 0 retry attempt(s).\n"

```

pkg/router/template/util/rewritetarget/rewritetarget_test.go
```diff
@@ -85,7 +85,35 @@ func Test_SanitizeInput(t *testing.T) {
...
+func Test_SanitizeRewritePathInput(t *testing.T) {
+	testCases := []struct {
+		name   string
+		input  string
+		output string
+	}{
+		{
+			name:   "single quotes should be always be escaped",
+			input:  `'foo'foo\'`,
+			output: `'\''foo'\''foo\'\''`,
```
rfredette, line 105: `\'` ends up generating an unpaired quote. Doesn't that still break the template?

gcs278: So `\'` generates `\'\''` because our rule is simple & universal: `'` gets translated to `'\''`, and we don't do anything with the `\`, so you mostly disregard it.

To explain the `'\''` sequence:
1. The first `'` terminates the outer quote provided by the template. 
2. The `\'` specifies a literal escaped quote (it must be outside of the protection of `''`)
3. The last `'` starts the outer quote again.

So the answer is no: it doesn't break the template, the `\` in front of the quote doesn't matter. The result of `\'` is that it gets interpreted as literal `\'`, which is desirable in this specific situation.

I'll also mention that we did handle single quotes in a different way in the fix for https://issues.redhat.com/browse/OCPBUGS-22739, but the key different here is that, in OCPBUGS-22739, `\'` was a valid rewrite annotation sequence (it matched `'`), where for this fix, though it doesn't cause a syntax error, it causes an unreachable situation, because the HAProxy map files MUST match the same pattern as the replace-path regex. The [GenerateRouteRegexp](https://github.com/openshift/router/blob/fb70ac4d547954e7f96c12b4bc329917f1695d35/pkg/router/template/util/util.go#L53) function is the reason why the HAPRoxy map path regexs are interpreted literally. I updated https://docs.google.com/document/d/1wA6ZdRYaiN70TY5UbTOn1wl5qAn_NPb2ty_G85XhnYM/edit with more info on this situation where the map files don't match the `<match-regex>` argument for rewrite target.

I'm on a bit of a tangent in this comment. I also realized `GenerateRouteRegexp` used `regexp.QuoteMeta(path)` instead of `\Q...\E`. They appear to do the same exact thing, but I think it's better to be consistent, especially if there is a subtle different between the two. We need to have the path interpreted in the same exact way for rewrites to work.

So I'll update the code to use `QuoteMeta`, and I also found that you can add a newline character in spec.path and break this config. So I fixed that too.

gcs278:
Per our discussion, I am planning to investigate what it would take to reject a route based on certain characters that are currently invalid.

/hold

gcs278:
After much discussion, this PR along isn't a satisfactory solution as it introduces changes in behavior for the rewrite-target annotation. More discussion is needed, but this PR can/should still be used as a reference.

/close

PR #612, opened 2024-07-03
OCPBUGS-36630: Add hack Makefile target to build CI version of router image

alebedev87:
This PR enables the build of the CI version router image on local. The same [base and router dockerfiles are used](https://github.com/openshift/release/blob/d27dd22029dc85e2d58327cc8ad88d6eaf33f44e/ci-operator/config/openshift/router/openshift-router-master.yaml#L16-L29) by the new target. Use case: I want the same image that the CI would build but with my fix.

### DIFF
--- a/hack/Makefile.debug
+++ b/hack/Makefile.debug
@@ -4,12 +4,19 @@ export GOOS=linux
 
 REGISTRY ?= quay.io
 IMAGE ?= openshift/openshift-router
+IMAGE_BASE ?= openshift/openshift-router-base
 TAG ?= latest
 IMAGEBUILDER ?= podman
 
 new-openshift-router-image:
 	GO111MODULE=on CGO_ENABLED=0 GOFLAGS=-mod=vendor go build -o openshift-router -gcflags=all="-N -l" ./cmd/openshift-router
 	$(IMAGEBUILDER) build -t $(IMAGE):$(TAG) -f hack/Dockerfile.debug .
+
+new-ci-router-image:
+	$(IMAGEBUILDER) build -t $(IMAGE_BASE):$(TAG) -f images/router/base/Dockerfile.ocp .
+	cp -f images/router/haproxy/Dockerfile.ocp images/router/haproxy/Dockerfile.ocp.tmp
+	sed -i "s|FROM .*|FROM $(IMAGE_BASE):$(TAG)|" images/router/haproxy/Dockerfile.ocp.tmp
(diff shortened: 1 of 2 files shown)


hack/Makefile.debug
```diff
@@ -3,14 +3,21 @@
...
 IMAGEBUILDER ?= podman
 
 new-openshift-router-image:
 	GO111MODULE=on CGO_ENABLED=0 GOFLAGS=-mod=vendor go build -o openshift-router -gcflags=all="-N -l" ./cmd/openshift-router
 	$(IMAGEBUILDER) build -t $(IMAGE):$(TAG) -f hack/Dockerfile.debug .
 
+new-ci-router-image:
+	$(IMAGEBUILDER) build -t $(IMAGE_BASE):$(TAG) -f images/router/base/Dockerfile.rhel .
+	cp -f images/router/haproxy/Dockerfile.rhel8 images/router/haproxy/Dockerfile.rhel8.debug
+	sed -i "s|FROM .*|FROM $(IMAGE_BASE):$(TAG)|" images/router/haproxy/Dockerfile.rhel8.debug
```
gcs278, line 18: Can we use dockerfile build args https://docs.docker.com/build/guide/build-args/? You can apply default arguments.

alebedev87: The rebase conflict I'm having on this PR is due to [the ART automation](https://github.com/openshift/router/commit/4b3e2655a7590e21dfdaab7c86c3ae5365842a4c) which updates `images/router/haproxy/Dockerfile.rhel8`'s `FROM`. The `ARG` may be preventing this automation from working smoothly.

grzpiotrowski:
Since this touches the hacks, I wanted to throw in a little change to propose alongside. Please ignore if this shouldn't be included here.

In [HACKING.md: lines 28-29](https://github.com/openshift/router/blob/master/HACKING.md?plain=1#L28) describe the`Dockerfile.debug` file and state that:

```
(...) The outputted image uses `centos:8` as it's base
since installing packages on an OpenShift RHEL base image requires RHEL entitlements.
```
but `ubi9` is actually used as the base image there instead. This mention of `centos:8` could be deleted then?

ShudiLi:
/label qe-approved
thanks

ShudiLi:
/label qe-approved
thanks

alebedev87:
Reverted the default image repository to `openshift` to reduce the diff with https://github.com/openshift/router/pull/609.

alebedev87:
It doesn't make sense to rebase and retest this PR, it's stale. I will keep using this commit in my fork.

PR #614, opened 2024-07-15
OCPBUGS-33958: secret re-creation scenario for externalCertificate with active informer

chiragkyal:
This PR:

- Handles secret re-creation scenarios for externalCertificate where a previously deleted secret is added again.
- Introduces a `deletedSecrets` map in `RouteSecretManager` to track routes for which the associated secret was deleted. This helps differentiate between new secret creation and recreation of a previously deleted secret.

- Refactors secret handling logic in `generateSecretHandler` to address these changes.
  - `AddFunc` handles the secret recreations. Upon detecting a recreated secret, it revalidates the route, updates its TLS configuration, and triggers a `Modified` event for the next plugins.
  - `DeleteFunc` does not  call `UnregisterRoute()`, instead tracks deleted secrets.
- Improves the logic for handling `Modified` events for routes with externalCertificate to avoid blindly unregistering and re-registrations with the secret manager. 

  - This change is also required to avoid always calling `UnregisterRoute()`, which leads to the stopping of the informer. 
  - Depends on https://github.com/openshift/library-go/pull/1740

- Makes stopping of the informer self-contained to the `RouteSecretManager` plugin.
    - `RemoveRoute`  (`templateRouter`) should not call `UnregisterRoute()` to avoid removal of informer when `RouteSecretManager` calls the next plugin chain with `watch.Deleted` event. 
    - Informer will be stopped only during the following cases
       - Route is deleted
       - Route is updated and `externalCertificate` field is removed
       - Route is updated and `externalCertificate` field is changed to a new secret name. [Old secret name is cached inside `SecretManager`'s `registeredHandlers` map. 
           - https://github.com/openshift/library-go/pull/1740

- Adds unit tests to cover the new logic and scenarios.

    **New Helper Functions:**

- `validate()` to encapsulate the externalCertificate validation logic.

- `populateRouteTLSFromSecret()` to update a route's in-memory TLS configuration from the referenced secret.

- `unregister()` to remove the route's registration with the secret manager and clean up references in `deletedSecrets`.

Part of : https://issues.redhat.com//browse/OCPBUGS-33958


alebedev87 (commented):
Some questions and nit picks.

pkg/router/controller/route_secret_manager.go
```diff
@@ -118,54 +180,67 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+			// This helps to differentiate between a new secret creation and a re-creation of a previously deleted secret.
+			key := generateKey(route)
+			if _, deleted := p.deletedSecrets.LoadAndDelete(key); deleted {
+				log.V(4).Info("secret recreated for route", "namespace", route.Namespace, "secret", secret.Name, "route", route.Name)
+
+				// re-validate
+				// since secret re-creation will not trigger the apiserver route admission,
+				// we need to rely on the router controller for this validation.
+				if err := p.validate(route); err != nil {
+					return
```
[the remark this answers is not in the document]
chiragkyal: If we fail here, then there are two choices
- Delete that secret which will re-populate `deletedSecrets` ; and then re-create it
- Update that secret

Both choices will only work after fixing the issue which had caused the validation to fail. 

> Of we fail here, will the fact that the deletedSecrets has been cleaned already cause any problem?

So, it should not cause any problem.

pkg/router/controller/route_secret_manager.go
```diff
@@ -118,54 +180,67 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+				}
+				// read the re-created secret and update TLS certificate and key
+				if err := p.populateRouteTLSFromSecret(route); err != nil {
+					return
+				}
+				// call the next plugin with watch.Modified
+				p.plugin.HandleRoute(watch.Modified, route)
+				// commit the changes
+				p.plugin.Commit()
```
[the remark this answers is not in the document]
chiragkyal: The  `RouterController` won't get an event to trigger the plugin chain in case of secret (re-)creation since the route spec won't update. 

So, it can not commit the changes in this case because that line won't be called. 
https://github.com/openshift/router/blob/72114ea78be0a032d5cdddcad694e0a352a00e93/pkg/router/controller/router_controller.go#L216

We're already doing a similar commit in case of the `UpdateFunc` logic just below it.

pkg/router/controller/route_secret_manager.go
```diff
@@ -76,37 +95,80 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+			// Old route had externalCertificate, new route does not
+			log.V(4).Info("Old route had externalCertificate, new route does not", "namespace", route.Namespace, "route", route.Name)
+			// unregister with secret monitor
+			if err := p.unregister(route); err != nil {
+				return err
+			}
+
+		case !newHasExt && !oldHadExt:
+			// Neither new nor old route have externalCertificate
+			// Do nothing
```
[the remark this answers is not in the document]
chiragkyal: Removed, thanks

pkg/router/controller/route_secret_manager.go
```diff
@@ -76,37 +95,80 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+			} else {
+				// ExternalCertificate is not updated
+				// Revalidate and update the in-memory TLS certificate and key
+				// It is the responsibility of this plugin to ensure everything is synced properly.
+
+				// One Scenario: The user deletes the secret, causing the route's status to be updated to reject.
+				// This triggers the entire plugin chain again. Without re-validating the external certificate
+				// and re-syncing the secret here, the route could become active again and start serving
+				// the default certificate, even though its spec has an external certificate.
+				// Therefore, it is essential to re-sync the secret to ensure the plugin chain correctly handles the route.
```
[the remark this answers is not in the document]
chiragkyal: When the user recreates the secret, `HandleRoute()` won't be called since route spec will be unchanged. However, I've updated the comment to call out explicitly that even when `ExternalCertificate` remains unchanged, we'll fall into this case.

pkg/router/controller/route_secret_manager.go
```diff
@@ -76,37 +95,80 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
-				log.Error(err, "failed to unregister route")
-				return err
+		// Determine if the route's external certificate configuration has changed
+		newHasExt := hasExternalCertificate(route)
+		oldSecret, oldHadExt := p.secretManager.LookupRouteSecret(route.Namespace, route.Name)
+
+		switch {
+		case newHasExt && oldHadExt:
+			// Both new and old routes have externalCertificate
+			log.V(6).Info("Both new and old routes have externalCertificate", "namespace", route.Namespace, "route", route.Name)
```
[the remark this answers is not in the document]
chiragkyal: Sure, updated the verbosity level to 4, to keep it consistent with other cases.

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -718,6 +760,210 @@ func TestRouteSecretManager(t *testing.T) {
...
+					},
+				},
+			},
+			secretManager: fake.SecretManager{
+				Secret: fakeSecret("other-sandbox", "different-tls-secret", corev1.SecretTypeTLS, map[string][]byte{
+					"tls.crt": []byte("my-crt"),
+					"tls.key": []byte("my-key"),
+				}),
+				IsPresent:  true,
+				SecretName: "tls-secret", // Used by LookupRouteSecret() to get the old secretName
```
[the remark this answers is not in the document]
chiragkyal: Since the `TestRouteSecretManager` test has grown, I think keeping some comments for certain test cases always helps in easier understanding and future debugging. WDYT?

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -880,21 +1130,25 @@ func TestRouteSecretManager(t *testing.T) {
 			if !reflect.DeepEqual(s.expectedRejections, recorder.rejections) {
 				t.Fatalf("expected rejections %v, but got %v", s.expectedRejections, recorder.rejections)
 			}
+			if _, exists := rsm.deletedSecrets.Load(generateKey(s.route)); exists {
+				t.Fatalf("deletedSecrets must be cleaned")
```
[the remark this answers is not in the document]
chiragkyal: Updated thanks.

pkg/router/controller/route_secret_manager.go
```diff
@@ -76,37 +95,76 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
-	// For Modified events always unregister and reregister the route even if the TLS configuration did not change.
-	// Since the `HandleRoute()` method does not carry the old route spec,
-	// and there's no definite way to compare old and new TLS configurations,
-	// assume that the TLS configuration is always updated, necessitating re-registration.
-	// Additionally, always creating a new `secretHandler` ensures that there are no stale route specs
+	// TODO always creating a new `secretHandler` ensures that there are no stale route specs
 	// in the next plugin chain, especially when the referenced secret is updated or deleted.
 	// This prevents sending outdated routes to subsequent plugins, preserving expected functionality.
-	// TODO: Refer https://github.com/openshift/router/pull/565#discussion_r1596441128 for possible ways to improve the logic.
+	// TODO : we might need to add RouteLister()
```
chiragkyal, line 101: I think we need to use `RouteLister()` inside all `ResourceEventHandlerFuncs` of `generateSecretHandler()` to avoid sending stale route object  in the next plugin chain.

pkg/router/controller/route_secret_manager_test.go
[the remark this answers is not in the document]
alebedev87: I'm fine with it.

pkg/router/controller/route_secret_manager.go
alebedev87: As discussed during the sync call, the problem statement is valid - the route state is captured at the time of the handler registration which this may lead to the update of the route to the older state. The solution which gets the route state from the controller-runtime cache may be problematic as it breaks out of the single source of truth for routes - route controller. Alternative solutions:
-  when the secret manager detects a secret recreation the corresponding route should be re-synced by the route controller forcing it to go through the plugin chain (preferred).
- when a route is registered into the secret manager its whole state can be passed as an additional parameter forming a dedicated cache of routes with external certificates. When the secret manager detects a secret recreation it can get the latest route state from this cache.

chiragkyal: > when the secret manager detects a secret recreation the corresponding route should be re-synced by the route controller forcing it to go through the plugin chain (preferred).

If we can send some event (maybe forcefully) to the Shared RouteInformer from the secret handlerFuncs without updating the actual route object, then that would be great. 

> when a route is registered into the secret manager its whole state can be passed as an additional parameter forming a dedicated cache of routes with external certificates. When the secret manager detects a secret recreation it can get the latest route state from this cache.

A route is registered only once with the secret manager. We're removing the logic of always doing re-registration of route for every plugin chain call in this PR to have the informer active. So, if we want to keep a dedicated cache of routes, it has to be part of `RouteSecretManager` rather than `SecretManager` (exists in library-go)

pkg/router/controller/route_secret_manager.go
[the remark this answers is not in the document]
chiragkyal: That's correct.

chiragkyal:
@alebedev87 Apart from https://github.com/openshift/router/pull/614#discussion_r1791310100 discussion, is this PR lgtm from your side?

pkg/router/controller/route_secret_manager.go
```diff
@@ -120,13 +126,23 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+				//    fetch the new TLS data. Note: These events won't trigger the apiserver
+				//    route admission, hence we need to rely on the router controller for this validation.
+				//
+				// 2. Consider a case where a user deletes the secret, causing the route's
+				//    status to be updated to "reject". This status update triggers the
+				//    entire plugin chain again.  Without re-validating the external certificate
+				//    and re-syncing the secret here, the route could incorrectly transition
+				//    back to an "active" state and start serving the default certificate
+				//    even though its spec still references an external certificate.
+				//
```
Miciah, line 145: Why would the router use the default certificate for the route?  Why wouldn't it continue to use the route's old external certificate?  

This point is central to my concern about the secret manager's setting `Admitted=False`.  I'm fine with setting `Admitted=False` in these two cases:

- The secret was deleted.
- A new secret was created after the old one had been deleted (which means that the route should already have `Admitted=False`).

In other words, the secret manager can set `Admitted=False` in any sequence of events that includes deleting the secret that the route references.  However, I still don't understand the necessity of setting `Admitted=False` in the case where the secret was _updated_.  We do need to force the controller to revalidate the route, but is it really necessary to mark the route as rejected in the meantime?  

Does setting `Admitted=False` when the secret is updated imply that rotating certificates necessarily causes downtime when using an external certificate?  ...Hm, maybe not, as you only call `p.plugin.HandleRoute(watch.Deleted, route)` in the delete case, and not in the update case; does that mean we report that the route is rejected but continue serving it unless revalidation re-rejects it?

In comparison, if the route has an inline certificate, the route can be updated with a new certificate, and the router continues using the old certificate until it observes the update; once the router does observe the update, it switches to the new certificate without rejecting the route and without any interruption to service.  I don't understand why the experience should or must be different when using an external certificate.

pkg/router/controller/route_secret_manager.go
```diff
@@ -120,13 +126,23 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
-				// the default certificate, even though its spec has an external certificate.
+				// Re-validate and update the in-memory TLS certificate and key (even if ExternalCertificate remains unchanged)
+				// It is the responsibility of this plugin to ensure everything is synced properly, because there might
+				// have been updates to the secret data or other events that require re-evaluation.
+				//
+				// 1. Secret update and re-create: Even if the externalCertificate name remains
+				//    the same, the router needs to be aware of the events when the content of
+				//    the secret is updated or the secret is recreated, to re-validate and
+				//    fetch the new TLS data. Note: These events won't trigger the apiserver
+				//    route admission, hence we need to rely on the router controller for this validation.
```
Miciah, line 137: You mean we rely on the controller to _trigger_ re-validation, right?  Or are you saying that this scenario requires the controller to perform some validation steps that would otherwise be handled by the admission plugin?    
```suggestion
				//    route admission, hence we need to rely on the router controller to trigger this validation.
```

pkg/router/controller/route_secret_manager.go
```diff
@@ -196,78 +212,102 @@ func (p *RouteSecretManager) validateAndRegister(route *routev1.Route) error {
...
+// Triggering a re-evaluation ensures that both:
+//   - Changes made by `RouteModifierFn` registered with the `RouterControllerFactory`
+//     are propagated correctly to all plugins.
+//   - All plugins get a chance to react to the changes and make necessary
+//     in-memory modifications to the route object, ensuring consistent behavior.
+//
+// - AddFunc:
+//   - Invoked when a new secret is added.
+//   - Handles secret recreation: If a secret is recreated (created after being
+//     deleted), it fetches the latest route object associated with it using the
```
Miciah, line 230: What happens if multiple routes reference the same secret?

pkg/router/controller/route_secret_manager.go
```diff
@@ -196,78 +212,102 @@ func (p *RouteSecretManager) validateAndRegister(route *routev1.Route) error {
...
-					return
-				}
-				// call the next plugin with watch.Modified
-				p.plugin.HandleRoute(watch.Modified, route)
-				// commit the changes
-				p.plugin.Commit()
+
+				// The route should *remain* rejected until it's re-evaluated
+				// by all the plugins (including this plugin). Once passes, the route will become active again.
+				p.recorder.RecordRouteRejection(route, "ExternalCertificateSecretReCreated", msg)
```
Miciah, line 271:
```suggestion
				p.recorder.RecordRouteRejection(route, "ExternalCertificateSecretRecreated", msg)
```

pkg/router/controller/route_secret_manager.go
```diff
@@ -196,78 +212,102 @@ func (p *RouteSecretManager) validateAndRegister(route *routev1.Route) error {
...
 			}
 		},
 
 		UpdateFunc: func(old interface{}, new interface{}) {
 			secretOld := old.(*kapi.Secret)
 			secretNew := new.(*kapi.Secret)
-			log.V(4).Info("secret updated for route", "namespace", route.Namespace, "secret", secretNew.Name, "old-version", secretOld.ResourceVersion, "new-version", secretNew.ResourceVersion, "route", route.Name)
+			key := generateKey(namespace, routeName)
+			msg := fmt.Sprintf("secret %q updated for route %q : old-version %q new-version %q", secretNew.Name, key, secretOld.ResourceVersion, secretNew.ResourceVersion)
```
Miciah, line 279:
```suggestion
			msg := fmt.Sprintf("secret %q updated for route %q (old resourceVersion=%v, new resourceVersion=%v)", secretNew.Name, key, secretOld.ResourceVersion, secretNew.ResourceVersion)

**Review by Miciah — COMMENTED** [review]
I took at look at the latest changes so I could leave some quick feedback and questions about recent changes.  I still need to go back and re-review the older changes.

File: pkg/router/controller/route_secret_manager.go (hunk 9 above)
Comment (chiragkyal) [reply pkg/router/controller/route_secret_manager.go:145]: >  I'm fine with setting Admitted=False in these two cases:
>
>The secret was deleted.
A new secret was created after the old one had been deleted (which means that the route should already have Admitted=False).

I do also think that we should keep `Admitted=False` for the above two cases. 

However, for secret update, I've now understood the inconsistency that might arise with inline certificate if we make `Admitted=False` when using an external certificate. Earlier, my thought process was to keep the route rejected until it's re-evaluated by all the plugins again. And since we were not doing `p.plugin.HandleRoute(watch.Deleted, route)` during secret update event; there should not be any downtime in terms of traffic.

Having said that, I've now added a new method called `RecordRouteUpdate` for `RouteStatusRecorder`, which will keep the `Admitted=True` and provide a reason for the route status update. In that latest changes I'm using `RecordRouteUpdate` during secret update events so that the route remains `Active` during it's re-evaluation process like the inline certificate unless revalidation re-rejects it.

File: pkg/router/controller/route_secret_manager.go (hunk 10 above)
Comment (chiragkyal) [reply pkg/router/controller/route_secret_manager.go:137]: Both reasons are valid I think.

We rely on the secret manager to trigger re-validation by updating the route status. Since the secret manager is managed by the controller, it’s reasonable to say the controller handles triggering re-validation.

Another reason for performing validation here is that updating the secret does not trigger the route admission plugin. Therefore, the controller must ensure validation is handled appropriately.

Previously, validation was performed [directly when the secret was updated](https://github.com/openshift/router/blob/78322373359023782b66c3e58e3178b9a4932bde/pkg/router/controller/route_secret_manager.go#L176-L192). Now, the process is split into two steps:

- Update the route status so all plugins can process the route.
- Re-validate everything within the plugin.

File: pkg/router/controller/route_secret_manager.go (hunk 11 above)
Comment (chiragkyal) [reply pkg/router/controller/route_secret_manager.go:230]: It will work similarly for multiple routes referencing the same secret. Each route has its own informer, so `AddFunc` will be called for each of the routes.

File: pkg/router/controller/route_secret_manager.go (hunk 12 above)
Comment (chiragkyal) [reply pkg/router/controller/route_secret_manager.go]: Updated, thanks

File: pkg/router/controller/route_secret_manager.go (hunk 13 above)
Comment (chiragkyal) [reply pkg/router/controller/route_secret_manager.go]: Updated, thanks

File: pkg/router/controller/route_secret_manager.go (hunk 9 above)
Comment (Miciah) [reply pkg/router/controller/route_secret_manager.go:145]: Thanks!  Is there any scenario in which `Admitted` could be `False` when the secret handler calls `RecordRouteUpdate`, and it ends up erroneously changing `Admitted=False` to `Admitted=True`?

File: pkg/router/controller/route_secret_manager.go [hunk 14]
```diff
@@ -118,106 +192,199 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+	if err := p.validate(route); err != nil {
 		return err
 	}
-
 	// register route with secretManager
-	handler := p.generateSecretHandler(route)
+	handler := p.generateSecretHandler(route.Namespace, route.Name)
 	if err := p.secretManager.RegisterRoute(context.TODO(), route.Namespace, route.Name, route.Spec.TLS.ExternalCertificate.Name, handler); err != nil {
 		log.Error(err, "failed to register route")
 		return err
```
alebedev87, line 203:
```suggestion
		return fmt.Errorf("failed to register router: %w", err)
```

pkg/router/controller/route_secret_manager.go
```diff
@@ -118,106 +192,199 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+			// Otherwise, no-op (new secret creation scenario and no race condition with that flow)
+			// This helps to differentiate between a new secret creation and a re-creation of a previously deleted secret.
+			key := generateKey(namespace, routeName)
+			if _, deleted := p.deletedSecrets.LoadAndDelete(key); deleted {
+				msg := fmt.Sprintf("secret %q recreated for route %q", secret.Name, key)
+				log.V(4).Info(msg)
+
+				// Ensure fetching the updated route
+				route, err := p.getUpdatedRoute(namespace, routeName)
+				if err != nil {
```
alebedev87, line 265: Can we log the error here (where we decide what to do with it) instead inside of the function?

pkg/router/controller/route_secret_manager.go
```diff
@@ -118,106 +192,199 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+			// Ensure fetching the updated route
+			route, err := p.getUpdatedRoute(namespace, routeName)
+			if err != nil {
+				return
 			}
 
+			// Reject this route
 			p.recorder.RecordRouteRejection(route, "ExternalCertificateSecretDeleted", msg)
+			// Stop serving this route
 			p.plugin.HandleRoute(watch.Deleted, route)
```
alebedev87, line 311: Does it mean that the route whose external certificate was removed will be removed from the HAProxy config?

pkg/router/controller/route_secret_manager.go
```diff
@@ -118,106 +192,199 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
-			fldPath := field.NewPath("spec").Child("tls").Child("externalCertificate")
-			if err := routeapihelpers.ValidateTLSExternalCertificate(route, fldPath, p.sarClient, p.secretsGetter).ToAggregate(); err != nil {
-				log.Error(err, "skipping route due to invalid externalCertificate configuration", "namespace", route.Namespace, "route", route.Name)
-				p.recorder.RecordRouteRejection(route, "ExternalCertificateValidationFailed", err.Error())
-				p.plugin.HandleRoute(watch.Deleted, route)
-				return
-			}
+			key := generateKey(namespace, routeName)
+			msg := fmt.Sprintf("secret %q updated for route %q (old resourceVersion=%v, new resourceVersion=%v)", secretNew.Name, key, secretOld.ResourceVersion, secretNew.ResourceVersion)
+			log.V(4).Info(msg)
```
alebedev87, line 280: Same comment about the structured logging.

pkg/router/controller/route_secret_manager.go
```diff
@@ -118,106 +192,199 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+}
+
+// getUpdatedRoute fetches the latest version of the route using the RouteLister.
+func (p *RouteSecretManager) getUpdatedRoute(namespace, routeName string) (*routev1.Route, error) {
+	updatedRoute, err := p.routelister.Routes(namespace).Get(routeName)
+	if err != nil {
+		log.Error(err, "failed to get route", "namespace", namespace, "route", routeName)
+		return nil, err
+	}
+	return updatedRoute, nil
```
alebedev87, line 389: If you follow the advice to log the error in the places of their usage, this function can be simplified to:
```suggestion
	return p.routelister.Routes(namespace).Get(routeName)
```
As a matter of fact this function would not be needed at all `p.routelister.Routes(namespace).Get(routeName)` can be used directly where it's needed.

pkg/router/controller/route_secret_manager.go
```diff
@@ -118,106 +192,199 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+//
+// NOTE: TLS data validation and sanitization are handled by the next plugin `ExtendedValidator`,
+// by reading the "tls.crt" and "tls.key" added by populateRouteTLSFromSecret.
+func (p *RouteSecretManager) validate(route *routev1.Route) error {
+	fldPath := field.NewPath("spec").Child("tls").Child("externalCertificate")
+	if err := routeapihelpers.ValidateTLSExternalCertificate(route, fldPath, p.sarClient, p.secretsGetter).ToAggregate(); err != nil {
+		log.Error(err, "skipping route due to invalid externalCertificate configuration", "namespace", route.Namespace, "route", route.Name)
+		p.recorder.RecordRouteRejection(route, "ExternalCertificateValidationFailed", err.Error())
+		p.plugin.HandleRoute(watch.Deleted, route)
+		return err
```
alebedev87, line 328: Can this be moved up the callstack in `HandleRoute` function where `validate` is called? This way we will have the plugin chain propagation (`p.plugin.HandleRoute`)  and logging in  the same function and not hidden in the callstack.

```suggestion
		return fmt.Errorf("invalid external certificate configuration: %w", err)
```

pkg/router/controller/route_secret_manager.go
```diff
@@ -118,106 +192,199 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+// the deletion of the route by calling the HandleRoute method with a watch.Deleted event.
+// Note: This function performs an in-place update of the route. The caller should be aware that the route's TLS configuration will be modified directly.
+func (p *RouteSecretManager) populateRouteTLSFromSecret(route *routev1.Route) error {
+	// read referenced secret
+	secret, err := p.secretManager.GetSecret(context.TODO(), route.Namespace, route.Name)
+	if err != nil {
+		log.Error(err, "failed to get referenced secret")
+		p.recorder.RecordRouteRejection(route, "ExternalCertificateGetFailed", err.Error())
+		p.plugin.HandleRoute(watch.Deleted, route)
+		return err
```
alebedev87, line 344: Same suggestion:
```suggestion
		return fmt.Errorf("failed to get referenced secret: %w", err)
```

alebedev87 (commented):
Moslty logging and nit picks.

pkg/router/controller/route_secret_manager.go
chiragkyal: As per the discussion during the sync call, I've added `isRouteAdmittedTrue` function which checks the current route status and decides how to update the status. 

```go
        // Update the route status to notify plugins, including this plugin, for re-evaluation.
	// - If the route is admitted (Admitted=True), record an update event.
	// - If the route is not admitted, record a rejection event.
	if isRouteAdmittedTrue(route.DeepCopy()) {
		p.recorder.RecordRouteUpdate(route, "ExternalCertificateSecretUpdated", msg)
	} else {
		p.recorder.RecordRouteRejection(route, "ExternalCertificateSecretUpdated", msg)
	}
```

pkg/router/controller/route_secret_manager.go
```diff
@@ -76,37 +105,82 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+		newHasExt := hasExternalCertificate(route)
+		oldSecret, oldHadExt := p.secretManager.LookupRouteSecret(route.Namespace, route.Name)
+
+		switch {
+		case newHasExt && oldHadExt:
+			// Both new and old routes have externalCertificate
+			log.V(4).Info("Both new and old routes have externalCertificate", "namespace", route.Namespace, "route", route.Name)
+			if oldSecret != route.Spec.TLS.ExternalCertificate.Name {
+				// ExternalCertificate is updated
+				log.V(4).Info("ExternalCertificate is updated", "namespace", route.Namespace, "route", route.Name)
```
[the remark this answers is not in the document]
chiragkyal: We're also registering the updated external certificate. So I've made it
```go
log.V(4).Info("Validating and registering updated external certificate", "namespace", route.Namespace, "oldSecret", oldSecret, "newSecret", route.Spec.TLS.ExternalCertificate.Name, "route", route.Name)
```

pkg/router/controller/route_secret_manager.go
```diff
@@ -76,37 +105,82 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
-				log.Error(err, "failed to unregister route")
-				return err
+		// Determine if the route's external certificate configuration has changed
+		newHasExt := hasExternalCertificate(route)
+		oldSecret, oldHadExt := p.secretManager.LookupRouteSecret(route.Namespace, route.Name)
+
+		switch {
+		case newHasExt && oldHadExt:
+			// Both new and old routes have externalCertificate
+			log.V(4).Info("Both new and old routes have externalCertificate", "namespace", route.Namespace, "route", route.Name)
```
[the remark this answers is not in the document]
chiragkyal: Removed this log line.
Added logs for `watch.Added` and `watch.Deleted` events

```go
log.V(4).Info("Validating and registering external certificate", "namespace", route.Namespace, "secret", route.Spec.TLS.ExternalCertificate.Name, "route", route.Name)

log.V(4).Info("Un-registering external certificate", "namespace", route.Namespace, "secret", secretName, "route", route.Name)
```

pkg/router/controller/route_secret_manager.go
```diff
@@ -76,37 +105,82 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+				// 2. Consider a case where a user deletes the secret, causing the route's
+				//    status to be updated to "reject". This status update triggers the
+				//    entire plugin chain again.  Without re-validating the external certificate
+				//    and re-syncing the secret here, the route could incorrectly transition
+				//    back to an "active" state and start serving the default certificate
+				//    even though its spec still references an external certificate.
+				//
+				// Therefore, it is essential to re-sync the secret to ensure the plugin chain correctly handles the route.
+
+				log.V(4).Info("ExternalCertificate is not updated", "namespace", route.Namespace, "route", route.Name)
```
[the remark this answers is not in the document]
chiragkyal: Updated, thanks

pkg/router/controller/route_secret_manager.go
```diff
@@ -76,37 +105,82 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+					return err
+				}
 			}
-		}
-		// register with secret monitor
-		if hasExternalCertificate(route) {
+
+		case newHasExt && !oldHadExt:
+			// New route has externalCertificate, old route did not
+			log.V(4).Info("New route has externalCertificate, old route did not", "namespace", route.Namespace, "route", route.Name)
```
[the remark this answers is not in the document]
chiragkyal: Updated to
```go
log.V(4).Info("Validating and registering new external certificate", "namespace", route.Namespace, "secret", route.Spec.TLS.ExternalCertificate.Name, "route", route.Name)
```

pkg/router/controller/route_secret_manager.go
```diff
@@ -76,37 +105,82 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+			// register with secret monitor
 			if err := p.validateAndRegister(route); err != nil {
 				return err
 			}
+
+		case !newHasExt && oldHadExt:
+			// Old route had externalCertificate, new route does not
+			log.V(4).Info("Old route had externalCertificate, new route does not", "namespace", route.Namespace, "route", route.Name)
```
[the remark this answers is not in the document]
chiragkyal: Thanks, updated

pkg/router/controller/route_secret_manager.go
```diff
@@ -118,106 +192,199 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
 	// secret handler
 	return cache.ResourceEventHandlerFuncs{
 
-		// AddFunc is intentionally left empty (only logs the event) because this handler is generated only after ensuring the existence of the secret.
-		// By leaving this empty, we prevent unnecessary triggering for the addition of the secret again. Additionally, GetSecret() method is called
-		// immediately after registering with the secretManager, to read the secret from the cache.
 		AddFunc: func(obj interface{}) {
 			secret := obj.(*kapi.Secret)
-			log.V(4).Info("secret added for route", "namespace", route.Namespace, "secret", secret.Name, "route", route.Name)
+			log.V(4).Info("secret added for route", "namespace", namespace, "secret", secret.Name, "route", routeName)
```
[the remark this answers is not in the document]
chiragkyal: Updated

pkg/router/controller/route_secret_manager.go
```diff
@@ -118,106 +192,199 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+
+			// Secret re-creation scenario
+			// Check if the route key exists in the deletedSecrets map, indicating that the secret was previously deleted for this route.
+			// If it exists, it means the secret is being recreated. Remove the key from the map and proceed with handling the route.
+			// Otherwise, no-op (new secret creation scenario and no race condition with that flow)
+			// This helps to differentiate between a new secret creation and a re-creation of a previously deleted secret.
+			key := generateKey(namespace, routeName)
+			if _, deleted := p.deletedSecrets.LoadAndDelete(key); deleted {
+				msg := fmt.Sprintf("secret %q recreated for route %q", secret.Name, key)
+				log.V(4).Info(msg)
```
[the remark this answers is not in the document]
chiragkyal: Okay, made the logs structured and kept it separete from the rejection message

pkg/router/controller/route_secret_manager.go
chiragkyal: Thanks, updated

pkg/router/controller/route_secret_manager.go
chiragkyal: Updated to
```go
route, err := p.routelister.Routes(namespace).Get(routeName)
if err != nil {
  log.Error(err, "failed to get route", "namespace", namespace, "route", routeName)
  return
}
```

pkg/router/controller/route_secret_manager.go
chiragkyal: Thanks for the suggestion. I've removed this function and using `p.routelister.Routes(namespace).Get(routeName)` directly.

pkg/router/controller/route_secret_manager.go
chiragkyal: Updated, thanks

pkg/router/controller/route_secret_manager.go
chiragkyal: This function is called in `HandleRoute` and also inside `validateAndRegister` function. Moving the logic from here and using them directly will reduce code readability. Moreover, the godoc added for this function is self explanatory that if validation fails, it records the route rejection and calls the next plugin. I would like to have them inside a function to reduce code redundancy.

pkg/router/controller/route_secret_manager.go
chiragkyal: Due to the same reason stated above, I would like to keep them inside a function to avoid code redundancy.

alebedev87:
I played with the fix on a live cluster:
```
$ oc version
Server Version: 4.18.0-0.test-2024-11-29-144149-ci-ln-42q6vwk-latest
```

Two things I noticed about the status of the route:
- After I recreated the secret for the second time the route's status got stuck with `ExternalCertificateSecretRecreated` reason not progressing to `Admitted=true`. The third recreated did the right thing: moved from `ExternalCertificateSecretRecreated`  to `Admitted=true`. Any subsequent recreation didn't fall into stuck `ExternalCertificateSecretRecreated`.
- When using 2 routes (with the same external certificate) the statuses differ after the secret deletion. The route which already experienced the recreation and the route which just got its first one: `ExternalCertificateSecretDeleted` (`secret "selfsigned" deleted for route "test/httpd-24"`) vs `ExternalCertificateValidationFailed` (`'spec.tls.externalCertificate: Not found: "secrets \"selfsigned\" not`). Also after the secret was recreated the status differed too: `Admitted=true`, `ExternalCertificateValidationFailed` (`spec.tls.externalCertificate: Not found: "secrets \"selfsigned\`).

I think that we need to test again those cases and try to find out whether there is any race or a glitch.

chiragkyal:
Thanks for doing the test. I also tried on a live cluster.

I think there is a race condition happening while updating the status of the route. The secret handler status update is racing with the status update of the plugin re-trigger.

The top level plugin has a lock : https://github.com/openshift/router/blob/be831fa5808df9facf7df4f5cdbf2472c504b60d/pkg/router/controller/router_controller.go#L210-L217 

which means all routes are getting processed serially, unlike secret handler.

As a solution, I think the `StatusAdmitter` plugin can have its own Mutex to lock the methods like `RecordRouteUpdate`, `RecordRouteRejection`, `HandleRoute`, etc., so that we can avoid race conditions.

chiragkyal:
@alebedev87 @Miciah As per our discussion over sync call, I've updated the contention tracker to ignore the status update done by the secret manager. Please take a final look at the PR, and let's try to get this merged. Thanks!

chiragkyal:
I did some testing on a real cluster with multiple routes using the same secret, and after the latest changes, the transition of the status were as expected and consistent through out the routes.

```
$ oc get route -n sandbox -w
NAME        HOST/PORT                                                                 PATH   SERVICES          PORT    TERMINATION   WILDCARD
example     example-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com            hello-openshift   <all>   edge          None
example1    example1-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example10   example10-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com          hello-openshift   <all>   edge          None
example11   example11-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com          hello-openshift   <all>   edge          None
example2    example2-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example3    example3-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example4    example4-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example5    example5-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example6    example6-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example7    example7-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example8    example8-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None


// Secret Deleted


example5    ExternalCertificateSecretDeleted                                                 hello-openshift   <all>   edge          None
example2    ExternalCertificateSecretDeleted                                                 hello-openshift   <all>   edge          None
example6    ExternalCertificateSecretDeleted                                                 hello-openshift   <all>   edge          None
example7    ExternalCertificateSecretDeleted                                                 hello-openshift   <all>   edge          None
example3    ExternalCertificateSecretDeleted                                                 hello-openshift   <all>   edge          None
example8    ExternalCertificateSecretDeleted                                                 hello-openshift   <all>   edge          None
example     ExternalCertificateSecretDeleted                                                 hello-openshift   <all>   edge          None
example11   ExternalCertificateSecretDeleted                                                 hello-openshift   <all>   edge          None
example1    ExternalCertificateSecretDeleted                                                 hello-openshift   <all>   edge          None
example10   ExternalCertificateSecretDeleted                                                 hello-openshift   <all>   edge          None
example4    ExternalCertificateSecretDeleted                                                 hello-openshift   <all>   edge          None
example5    ExternalCertificateValidationFailed                                              hello-openshift   <all>   edge          None
example2    ExternalCertificateValidationFailed                                              hello-openshift   <all>   edge          None
example6    ExternalCertificateValidationFailed                                              hello-openshift   <all>   edge          None
example7    ExternalCertificateValidationFailed                                              hello-openshift   <all>   edge          None
example3    ExternalCertificateValidationFailed                                              hello-openshift   <all>   edge          None
example8    ExternalCertificateValidationFailed                                              hello-openshift   <all>   edge          None
example     ExternalCertificateValidationFailed                                              hello-openshift   <all>   edge          None
example11   ExternalCertificateValidationFailed                                              hello-openshift   <all>   edge          None
example1    ExternalCertificateValidationFailed                                              hello-openshift   <all>   edge          None
example10   ExternalCertificateValidationFailed                                              hello-openshift   <all>   edge          None
example4    ExternalCertificateValidationFailed                                              hello-openshift   <all>   edge          None

// Secret Re-Created

example5    ExternalCertificateSecretRecreated                                               hello-openshift   <all>   edge          None
example10   ExternalCertificateSecretRecreated                                               hello-openshift   <all>   edge          None
example11   ExternalCertificateSecretRecreated                                               hello-openshift   <all>   edge          None
example     ExternalCertificateSecretRecreated                                               hello-openshift   <all>   edge          None
example4    ExternalCertificateSecretRecreated                                               hello-openshift   <all>   edge          None
example1    ExternalCertificateSecretRecreated                                               hello-openshift   <all>   edge          None
example7    ExternalCertificateSecretRecreated                                               hello-openshift   <all>   edge          None
example2    ExternalCertificateSecretRecreated                                               hello-openshift   <all>   edge          None
example8    ExternalCertificateSecretRecreated                                               hello-openshift   <all>   edge          None
example6    ExternalCertificateSecretRecreated                                               hello-openshift   <all>   edge          None
example3    ExternalCertificateSecretRecreated                                               hello-openshift   <all>   edge          None
example5    example5-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example10   example10-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com          hello-openshift   <all>   edge          None
example11   example11-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com          hello-openshift   <all>   edge          None
example     example-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com            hello-openshift   <all>   edge          None
example4    example4-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example1    example1-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example7    example7-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example2    example2-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example8    example8-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example6    example6-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None
example3    example3-sandbox.apps.ckyal-20241216-486d8f62.devcluster.openshift.com           hello-openshift   <all>   edge          None

```

alebedev87:
/approve

Thanks for [the testing](https://github.com/openshift/router/pull/614#issuecomment-2552133142), we should be good now.

chiragkyal:
tide/merge-method-squash

pkg/router/controller/contention.go
```diff
@@ -279,8 +292,11 @@ func ingressConditionsEqual(a, b []routev1.RouteIngressCondition) bool {
 	return true
 }
 
-// conditionsEqual compares two RouteIngressConditions, ignoring LastTransitionTime.
+// conditionsEqual compares two RouteIngressConditions, ignoring LastTransitionTime and any reason in ignoreIngressConditionReason..
```
Miciah, line 295:
```suggestion
// conditionsEqual compares two RouteIngressConditions, ignoring LastTransitionTime and any reason in ignoreIngressConditionReason.
```
No need to fix this now, but maybe we can fix it in the next PR that touches the contention code.

pkg/router/controller/contention.go
```diff
@@ -279,8 +292,11 @@ func ingressConditionsEqual(a, b []routev1.RouteIngressCondition) bool {
...
 func conditionsEqual(a, b *routev1.RouteIngressCondition) bool {
+	if ignoreIngressConditionReason.Has(a.Reason) || ignoreIngressConditionReason.Has(b.Reason) {
+		return true
+	}
```
Miciah, line 299: We discussed today that this condition may be broader than necessary, but we couldn't think of a realistic scenario in which it would cause problems.  However, we may need to revisit this condition if we ever observe any strange behavior with respect to status updates for routes with external certificates during, say, a rolling update of a router deployment.

pkg/router/controller/route_secret_manager.go
```diff
@@ -28,18 +38,33 @@ type RouteSecretManager struct {
 
 	secretManager secretmanager.SecretManager
 	secretsGetter corev1client.SecretsGetter
+	routelister   routelisters.RouteLister
 	sarClient     authorizationclient.SubjectAccessReviewInterface
+	// deletedSecrets tracks routes for which the associated secret was deleted after intial creation of the secret monitor.
```
Miciah, line 43:
```suggestion
	// deletedSecrets tracks routes for which the associated secret was deleted after initial creation of the secret monitor.
```

pkg/router/controller/route_secret_manager.go
```diff
@@ -28,18 +38,33 @@ type RouteSecretManager struct {
...
+	// This helps to differentiate between a new secret creation and a recreation of a previously deleted secret.
+	// Populated inside DeleteFunc, and consumed or cleaned inside AddFunc and unregister().
+	// It is thread safe and "namespace/routeName" is used as it's key.
```
Miciah, line 46:
```suggestion
	// It is thread safe and "namespace/routeName" is used as its key.
```

pkg/router/controller/route_secret_manager.go
```diff
@@ -60,8 +85,20 @@ func (p *RouteSecretManager) Commit() error {
...
 
 // HandleRoute manages the registration, unregistration, and validation of routes with external certificates.
+//
 // For Added events, it validates the route's external certificate configuration and registers it with the secret manager.
-// For Modified events, it first unregisters the route if it's already registered and then revalidates and registers it again.
+//
+// For Modified events, it checks if the route's external certificate configuration has changed and takes appropriate actions:
+//  1. Both the old and new routes have an external certificate:
+//     - If the external certificate has changed, it unregisters the old one and registers the new one.
+//     - If the external certificate has not changed, it revalidates and updates the in-memory TLS certificate and key.
```
Miciah, line 94: The comment is a little unclear.  Presumably by "the external certificate has changed" or "the external certificate has not changed", you mean that the external certificate secret _reference_ has or has not changed.  That is, if the secret _reference_ changes, you need to track the new _referent_, but if the reference is the same, you only need to check whether the referent secret's contents have changed.

pkg/router/controller/route_secret_manager.go
```diff
@@ -118,106 +201,207 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
-			// call the next plugin with watch.Modified
-			p.plugin.HandleRoute(watch.Modified, route)
-			// commit the changes
-			p.plugin.Commit()
+			msg := fmt.Sprintf("secret %q updated for route %q (oldSecretVersion=%v, newSecretVersion=%v)", secretNew.Name, key, secretOld.ResourceVersion, secretNew.ResourceVersion)
+			// Update the route status to notify plugins, including this plugin, for re-evaluation.
+			// - If the route is admitted (Admitted=True), record an update event.
+			// - If the route is not admitted, record a rejection event (keep it rejected).
+			if isRouteAdmittedTrue(route.DeepCopy()) {
+				p.recorder.RecordRouteUpdate(route, ExtCrtStatusReasonSecretUpdated, msg)
```
Miciah, line 302: `isRouteAdmittedTrue` checks whether _any_ router has admitted the route:
https://github.com/openshift/router/blob/3966d000324c5ddaa56d7ae26f4055bd3d15c7f9/pkg/router/controller/route_secret_manager.go#L396-L407

Suppose that the current router has previously rejected the route, but another router admitted it.  Then will this logic cause the current router to admit the route erroneously?  

The route status could have a stale status entry from a router that is no longer running, or from a router that is using an old image, or from a router that is configured differently (for example, it might allow wildcards while the current one does not).  Why wouldn't you check only whether _this_ router has admitted the route?

pkg/router/controller/route_secret_manager.go
```diff
@@ -118,106 +201,207 @@ func (p *RouteSecretManager) HandleRoute(eventType watch.EventType, route *route
...
+//     are propagated correctly to all plugins.
+//   - All plugins get a chance to react to the changes and make necessary
+//     in-memory modifications to the route object, ensuring consistent behavior.
+//
+// - AddFunc:
+//   - Invoked when a new secret is added.
+//   - Handles secret recreation: If a secret is recreated (created after being
+//     deleted), it fetches the latest route object associated with it using the
+//     RouteLister and updates the route's status to trigger a full
+//     re-evaluation of the route by the entire plugin chain.
```
Miciah, line 240: I think there is a gap here.  Consider the following scenario:

1. Start with a router deployment.
2. Create a route with an external certificate reference to secret.
3. Delete the secret.
4. Restart the router deployment's pods (for example, for a rolling update).
5. Recreate the secret.

The new router pods will reject the route at step 4 and will miss the secret recreate event at step 5, right?  

It might not be feasible to close this gap, but I want to raise this concern to make sure I understand correctly and to note that it is a potential issue.

chiragkyal: I think this is likely possible because the route pod is managing the lifecycle of the secret manager and the secret informers. Restarting the router deployment's pods will kill the previously stored watch, and new pods will not register the new watch because the secret won't be there. So, yeah, in this scenario, we will miss the secret recreate event. 

> It might not be feasible to close this gap, but I want to raise this concern to make sure I understand correctly and to note that it is a potential issue.

I agree on this, given the secret manager's dependency on the router pods.

pkg/router/controller/route_secret_manager.go
chiragkyal: > Why wouldn't you check only whether _this_ router has admitted the route?

Thanks for the suggestion. Yeah, it makes sense to me. I have updated the logic in https://github.com/openshift/router/pull/614/commits/62daef459fde17c27007cd8d0ac6bc24209b2120 commit to pass the [routerName](https://github.com/openshift/router/blob/282180b5d4599c00bdb1a85b919948d21146cf8d/pkg/cmd/infra/router/router.go#L35) to [RouteSecretManager](https://github.com/openshift/router/blob/282180b5d4599c00bdb1a85b919948d21146cf8d/pkg/cmd/infra/router/template.go#L818C26-L818C44) and check the status change for only this specific router.

pkg/router/controller/route_secret_manager.go
chiragkyal: Updated in https://github.com/openshift/router/commit/62daef459fde17c27007cd8d0ac6bc24209b2120

pkg/router/controller/route_secret_manager.go
chiragkyal: Thanks for the catch. Updated in https://github.com/openshift/router/commit/62daef459fde17c27007cd8d0ac6bc24209b2120

pkg/router/controller/contention.go
chiragkyal: Thanks for the catch. Updated in https://github.com/openshift/router/commit/62daef459fde17c27007cd8d0ac6bc24209b2120

Miciah:
Just an observation: It occurs to me that using `--update-status=false` or `ROUTER_UPDATE_STATUS=false` would prevent the router controller from updating the route status and thus prevent the secret manager from triggering reconciliation.  However, we don't expose an API for `--update-status=false` or `ROUTER_UPDATE_STATUS=false`.  We do use `--update-status=false` in some openshift/origin tests, but it would only become an issue if we had a test that used both external certificates and `--update-status=false`.

pkg/router/controller/route_secret_manager.go
Miciah: Thanks for confirming my understanding!  If the comment is unclear, we should try to make it clearer.  If not in this PR, maybe in the next one that touches this code.

pkg/router/controller/status.go
```diff
@@ -36,6 +37,7 @@ const (
 // RouteStatusRecorder is an object capable of recording why a route status condition changed.
 type RouteStatusRecorder interface {
 	RecordRouteRejection(route *routev1.Route, reason, message string)
+	RecordRouteUpdate(route *routev1.Route, reason, message string)
```
Miciah, line 40: As a future clean-up, `HandleRoute` could use  this new method.

Miciah:
Excellent work!  Thank you for your patience and effort with this PR!
/lgtm

chiragkyal:
> Excellent work! Thank you for your patience and effort with this PR!

Finally!! Thank you very much for the approval.

PR #619, opened 2024-08-28
OCPBUGS-38426: DO NOT MERGE: validate haproxy22-2.2.24-5.rhaos4.13.el8.x86_64.rpm

gcs278:
/hold
/wip

### DIFF
--- a/0001-Resolve-https-issues.redhat.com-browse-OCPBUGS-24455.patch
+++ b/0001-Resolve-https-issues.redhat.com-browse-OCPBUGS-24455.patch
@@ -1,0 +1,683 @@
+From 3f950735d5c8cca8d288b880811468e3d74d8be5 Mon Sep 17 00:00:00 2001
+From: Grant Spence <gspence@redhat.com>
+Date: Wed, 28 Aug 2024 08:30:00 -0400
+Subject: Resolve https://issues.redhat.com/browse/OCPBUGS-24455
+ (CVE-2023-45539)
+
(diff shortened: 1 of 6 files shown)


gcs278:
Everything worked first round.
/test all

gcs278:
2nd test run success:
/test all

gcs278:
One more round for sanity:
/test all

alebedev87:
It's a pity `e0c9008874b89621449f7ff3e9bc6db4e94fac6d BUG/MINOR: h2: reject more chars from the :path pseudo header`commit didn't mention `b12ab9c04 MINOR: ist: Add istend() function to return a pointer to the end of the string` as a dependency along with the others. Should we do it in our patch or we don't do it in haproxy distgit and [Jira comment](https://issues.redhat.com/browse/OCPBUGS-39119?focusedId=25489157&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25489157)  is enough?

alebedev87:
> Right, it's a pity, but I don't think we should edit the commit message, I think we should still take the commit as-is to avoid confusion. I think leaving a paper (comment) trail in the Jira is sufficient.

I thought of a README file rather the commit message. But I think Jira ticket reference (which is already there) is indeed enough.

gcs278:
See [comment](https://issues.redhat.com/browse/OCPBUGS-39119?focusedId=25513826&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25513826)
/close

PR #621, opened 2024-08-30
OCPBUGS-39403: Fix parseIPList to Continue Processing After Invalid IPs and Return Valid IPs

mJace:
### Summary

This PR updates the `parseIPList` function to handle IP lists containing invalid IPs or CIDRs more gracefully. Previously, the function would return an empty string as soon as it encountered an invalid IP or CIDR. This update ensures that the function will now continue processing the rest of the list and return a space-separated string of all valid IPs and CIDRs.

### Changes

- Modified `parseIPList` to:
  - Log all invalid entries for troubleshooting.
  - Return a string of valid IPs and CIDRs, or an empty string if none are valid.

### Benefits

- **Improved Robustness**: The function now handles mixed validity IP lists more gracefully, allowing valid entries to be processed even if there are some invalid entries.
- **Enhanced Debugging**: Logging of invalid IPs or CIDRs provides better visibility into issues with input data.

### Additional Notes

- The function continues to trim and validate input while logging detailed information about invalid entries.
- Ensure to review and test the function with various IP list scenarios to confirm the correctness of the changes.

### DIFF
--- a/pkg/router/template/template_helper.go
+++ b/pkg/router/template/template_helper.go
@@ -367,26 +367,40 @@ func parseIPList(list string) string {
 
 	trimmedList := strings.TrimSpace(list)
 	if trimmedList == "" {
+		log.V(0).Info("parseIPList empty list found")
 		return ""
 	}
 
 	// same behavior as the previous approach with regexp
 	if trimmedList != list {
-		log.V(7).Info("parseIPList leading/trailing spaces found")
+		log.V(0).Info("parseIPList leading/trailing spaces found")
 		return ""
 	}
 
+	var validIPs []string
+
 	ipList := strings.Fields(list)
 	for _, ip := range ipList {
-		if net.ParseIP(ip) == nil {
-			if _, _, err := net.ParseCIDR(ip); err != nil {
-				log.V(7).Info("parseIPList found not IP/CIDR item", "value", ip, "err", err)
-				return ""
-			}
+		// check if it's a valid IP
+		if net.ParseIP(ip) != nil {
+			validIPs = append(validIPs, ip)
+		} else if _, _, err := net.ParseCIDR(ip); err == nil {
+			// check if it's a valid CIDR
+			validIPs = append(validIPs, ip)
+		} else {
+			// Log invalid IP/CIDR
+			log.V(0).Info("parseIPList found invalid IP/CIDR", ip)
 		}
 	}
-	log.V(7).Info("parseIPList parsed the list", "value", list)
-	return list
+
+	if len(validIPs) == 0 {
+		log.V(0).Info("No valid IP/CIDR in the list")
+		return ""
+	}
+
+	result := strings.Join(validIPs, " ")
+	log.V(7).Info("parseIPList parsed the list", "validIPs", result)
+	return result
 }
 
 var helperFunctions = template.FuncMap{
--- a/pkg/router/template/template_helper_test.go
+++ b/pkg/router/template/template_helper_test.go
@@ -981,33 +981,40 @@ func TestGenerateHAProxyWhiteListFile(t *testing.T) {
 
 func TestParseIPList(t *testing.T) {
 	testCases := []struct {
-		name          string
-		input         string
-		expectedEmpty bool
+		name           string
+		input          string
+		expectedEmpty  bool
+		expectedReturn string
 	}{
 		{
-			name:  "All mixed",
-			input: "192.168.1.0 2001:0db8:85a3:0000:0000:8a2e:0370:7334 172.16.14.10/24 2001:0db8:85a3::8a2e:370:10/64 64:ff9b::192.168.0.1 2600:14a0::/40",
+			name:           "All mixed",
+			input:          "192.168.1.0 2001:0db8:85a3:0000:0000:8a2e:0370:7334 172.16.14.10/24 2001:0db8:85a3::8a2e:370:10/64 64:ff9b::192.168.0.1 2600:14a0::/40",
+			expectedReturn: "192.168.1.0 2001:0db8:85a3:0000:0000:8a2e:0370:7334 172.16.14.10/24 2001:0db8:85a3::8a2e:370:10/64 64:ff9b::192.168.0.1 2600:14a0::/40",
 		},
 		{
-			name:  "IPs only",
-			input: "192.168.1.0 2001:0db8:85a3:0000:0000:8a2e:0370:7334 64:ff9b::192.168.0.1 172.16.14.10",
+			name:           "IPs only",
+			input:          "192.168.1.0 2001:0db8:85a3:0000:0000:8a2e:0370:7334 64:ff9b::192.168.0.1 172.16.14.10",
+			expectedReturn: "192.168.1.0 2001:0db8:85a3:0000:0000:8a2e:0370:7334 64:ff9b::192.168.0.1 172.16.14.10",
 		},
 		{
-			name:  "CIDRs only",
-			input: "192.168.1.0/16 2001:0db8:85a3:0000:0000:8a2e:0370:7334/48 172.16.14.10/24 2001:0db8:85a3::8a2e:0370:10/64 2600:14a0::/40",
+			name:           "CIDRs only",
+			input:          "192.168.1.0/16 2001:0db8:85a3:0000:0000:8a2e:0370:7334/48 172.16.14.10/24 2001:0db8:85a3::8a2e:0370:10/64 2600:14a0::/40",
+			expectedReturn: "192.168.1.0/16 2001:0db8:85a3:0000:0000:8a2e:0370:7334/48 172.16.14.10/24 2001:0db8:85a3::8a2e:0370:10/64 2600:14a0::/40",
 		},
 		{
-			name:  "IPv6 only",
-			input: "2001:0db8:85a3:0000:0000:8a2e:0370:7334 2001:0db8:85a3::8a2e:370:10/64 2001:db8::2:1 ::ffff:192.168.0.1 2600:14a0::/40",
+			name:           "IPv6 only",
+			input:          "2001:0db8:85a3:0000:0000:8a2e:0370:7334 2001:0db8:85a3::8a2e:370:10/64 2001:db8::2:1 ::ffff:192.168.0.1 2600:14a0::/40",
+			expectedReturn: "2001:0db8:85a3:0000:0000:8a2e:0370:7334 2001:0db8:85a3::8a2e:370:10/64 2001:db8::2:1 ::ffff:192.168.0.1 2600:14a0::/40",
 		},
 		{
-			name:  "IPv4 only",
-			input: "192.168.10.10 10.168.12.10/8 8.8.8.8 172.16.0.0/24",
+			name:           "IPv4 only",
+			input:          "192.168.10.10 10.168.12.10/8 8.8.8.8 172.16.0.0/24",
+			expectedReturn: "192.168.10.10 10.168.12.10/8 8.8.8.8 172.16.0.0/24",
 		},
 		{
-			name:  "Single IP",
-			input: "192.168.15.15",
+			name:           "Single IP",
+			input:          "192.168.15.15",
+			expectedReturn: "192.168.15.15",
 		},
 		{
 			// as behavior as the previous (regexp) approach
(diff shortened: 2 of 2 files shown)


alebedev87:
@mJace : would it be possible for you to provide some unit tests for this change?

mJace:
@alebedev87 
I added more test case to function `TestParseIPList` in `pkg/router/template/template_helper_test.go`.

pkg/router/template/template_helper.go
```diff
@@ -376,17 +376,33 @@ func parseIPList(list string) string {
...
+			validIPs = append(validIPs, ip)
+		} else {
+			invalidIPs = append(invalidIPs, ip)
 		}
 	}
-	log.V(7).Info("parseIPList parsed the list", "value", list)
-	return list
+
+	if len(invalidIPs) > 0 {
+		log.V(7).Info("parseIPList found invalid IP/CIDR items", "invalidIPs", invalidIPs)
```
candita, line 396: I would keep the log line up at line 391 where you find an invalid IP, e.g. log it for each invalid ip. Then you don't need the invalidIPs slice, nor to take a len later.

mJace: Thank you Candace!

pkg/router/template/template_helper.go
```diff
@@ -376,17 +376,29 @@ func parseIPList(list string) string {
...
-			}
+		// check if it's a valid IP
+		if net.ParseIP(ip) != nil {
+			validIPs = append(validIPs, ip)
+		} else if _, _, err := net.ParseCIDR(ip); err == nil {
+			// check if it's a valid CIDR
+			validIPs = append(validIPs, ip)
+		} else {
+			// Log invalid IP/CIDR
+			log.V(0).Info("parseIPList found invalid IP/CIDR", ip)
```
candita, line 391: Since this changes behavior, we need to make sure this error is noticed, and indicate that the invalid IP is now ignored.  Something like this:
```suggestion
			log.Error("error parsing IP list, ignoring invalid IP/CIDR", ip)
```

pkg/router/template/template_helper_test.go
```diff
@@ -981,74 +981,125 @@ func TestGenerateHAProxyWhiteListFile(t *testing.T) {
...
 		},
 		{
 			// as behavior as the previous (regexp) approach
-			name:          "Leading and trailing spaces",
-			input:         " 192.168.10.10  ",
-			expectedEmpty: true,
+			name:           "Leading and trailing spaces",
+			input:          " 192.168.10.10  ",
+			expectedEmpty:  true,
+			expectedReturn: "",
```
candita, line 1024: `expectedReturn: ""` is not needed.  It is implicit that `expectedReturn: ""` when `expectedEmpty: true`.

pkg/router/template/template_helper_test.go
```diff
@@ -981,74 +981,125 @@ func TestGenerateHAProxyWhiteListFile(t *testing.T) {
...
+			name:           "Wrong IPv6 in a IPv6 list",
+			input:          "2001:0db8:85a3:0000:8a2e:0370:7334 2001:0db8:85a3::8a2e:370:10/64 2001:db8::2:1 ::ffff:192.168.0.1 :/40",
+			expectedEmpty:  false,
+			expectedReturn: "2001:0db8:85a3::8a2e:370:10/64 2001:db8::2:1 ::ffff:192.168.0.1",
 		},
 		{
-			name:          "Wrong IPv4 CIDR",
-			input:         "192.168.10.5/64",
-			expectedEmpty: true,
+			name:           "All mixed type with invlaid IPv4",
```
candita, line 1087:
```suggestion
			name:           "All mixed type with invalid IPv4",
```

pkg/router/template/template_helper_test.go
```diff
@@ -1061,7 +1112,7 @@ func TestParseIPList(t *testing.T) {
 				}
 				return
 			}
-			if got != tc.input {
+			if got != tc.expectedReturn {
```
candita, line 1115: Followup to https://github.com/openshift/router/pull/621/files#r1815351390.
```suggestion
			if expectedEmpty && got != "" || got != tc.expectedReturn {
```

pkg/router/template/template_helper_test.go
```diff
@@ -1061,7 +1112,7 @@ func TestParseIPList(t *testing.T) {
...
 				t.Errorf("Failure: expected %q, got %q", tc.input, got)
```
candita, line 1116:
```suggestion
				t.Errorf("Failure: expected %q, got %q", tc.expectedReturn, got)
```

pkg/router/template/template_helper_test.go
```diff
@@ -981,74 +981,125 @@ func TestGenerateHAProxyWhiteListFile(t *testing.T) {
...
+			name:           "Wrong IPv6 CIDR",
+			input:          "2600:14a0::/256",
+			expectedEmpty:  true,
+			expectedReturn: "",
 		},
 		{
-			name:          "Only white spaces",
-			input:         "   ",
-			expectedEmpty: true,
+			name:           "Wrong IPv4 in a IPs only list",
```
candita, line 1063: nit, here and elsewhere, use "an" before "IPs"
```suggestion
			name:           "Wrong IPv4 in an IPs only list",
```

pkg/router/template/template_helper.go
mJace: I added a customErr for IP/CIDR since only `net.parseCIDR` returns `err`.

mJace:
@candita 
I combined your suggestion 
https://github.com/openshift/router/pull/621#pullrequestreview-2393212003
https://github.com/openshift/router/pull/621#pullrequestreview-2393209390

so the test function is like
```
	for _, tc := range testCases {
			got := parseIPList(tc.input)
			if tc.expectedEmpty {
				if got != "" {
					t.Errorf("Expected empty got %q", got)
				}
				return
			}
			if tc.expectedEmpty && got != "" || got != tc.expectedReturn {
				t.Errorf("Failure: expected %q, got %q", tc.input, got)
			}
		})
	}
```

or do you think it is ok to make it more simple?
```
	for _, tc := range testCases {
			got := parseIPList(tc.input)
			if tc.expectedEmpty && got != "" || got != tc.expectedReturn {
				t.Errorf("Failure: expected %q, got %q", tc.input, got)
			}
		})
	}
```

pkg/router/template/template_helper.go
alebedev87: I'm wondering whether this one should be an error, taking into account that now we don't stop the parsing and there can be some valid IPs still. I think that the error would make more sense down the code where we find that the list is empty. However in the similar conditions in the same function (when we return `""`) we use `log.V(7).Info()`. So it seems to be inconsistent.

pkg/router/template/template_helper_test.go
```diff
@@ -1046,9 +1053,46 @@ func TestParseIPList(t *testing.T) {
 			expectedEmpty: true,
 		},
 		{
-			name:          "Wrong IP in a list",
-			input:         "192.168.1.0 2001:0db8:85a3:0000:0000:8a2e:0370:7334 172.16.14.10/24 2001:0db8:85a3::8a2e:370:10/64 64:ff9b::192.168.0.1 10.",
-			expectedEmpty: true,
+			name:           "Wrong IPv4 in an IPs only list",
+			input:          "192.168.1.0 2001:0db8:85a3:0000:0000:8a2e:0370:7334 172.16.14.10/24 2001:0db8:85a3::8a2e:370:10/64 64:ff9b::192.168.0.1 10.",
+			expectedEmpty:  false,
```
alebedev87, line 1058: When `expectedEmpty` is `false` it can be skipped from the test struct, similar to the other tests which were modified above.

alebedev87 (commented):
I think it's close to LGTM. I just wanted to discuss [this thread](https://github.com/openshift/router/pull/621#discussion_r1816935225) and I agree with Candace that [this `if` should be changed](https://github.com/openshift/router/pull/621/files#r1815367684) to be more straightforward.

alebedev87:
Sorry for the double post of the review, some buggy behavior from github.

pkg/router/template/template_helper.go
mJace: @alebedev87  That’s why I use log.V(0).Info at the beginning.
I reviewed all scenarios that call log.Error() in `template_helper.go`, and they all return "" immediately.
Since, in this case, we will continue parsing the rest of the IP, is it okay to use log.Info but at a higher log level?
If yes, what level should I use here?

alebedev87: [The default log level is 2](https://github.com/openshift/router/blob/master/images/router/haproxy/Dockerfile.ocp#L22) so if we want this message to be visible it has to be `<= 2`. From what I can see in `template_helper.go` the log level `0` is what's used in similar cases ([example](https://github.com/openshift/router/blob/40db90072b633e6d0ec7d064bd8441f12065a2c4/pkg/router/template/template_helper.go#L136)) - when the user needs to be informed but the flow may continue. Similarly, looking at the other functions - failing the parsing with an error if the list of valid IPs/CIDRs is empty seems to be the way to go.

alebedev87: Btw the condition about the trimmer list being compared to the given list doesn't seem to use the right logging level:
```go
	if trimmedList != list {
		log.V(7).Info("parseIPList leading/trailing spaces found")
		return ""
	}
```

I would prefer it to be a level `0` too as we want the user to see this.

pkg/router/template/template_helper_test.go
```diff
@@ -1061,9 +1098,24 @@ func TestParseIPList(t *testing.T) {
...
 				return
 			}
-			if got != tc.input {
+			if tc.expectedEmpty && got != "" || got != tc.expectedReturn {
 				t.Errorf("Failure: expected %q, got %q", tc.input, got)
 			}
 		})
 	}
+
+	for _, tc := range testCases {
```
alebedev87, line 1107: This version of `for` looks good to me. Can you remove the previous one (starts on the line 1055)? Because correctly we run the test for `parseIPList` twice?

alebedev87:
/retitle OCPBUGS-39403: Fix parseIPList to Continue Processing After Invalid IPs and Return Valid IPs

alebedev87:
/lgtm
/approve
/hold

Holding for @candita to have another look.

alebedev87:
https://github.com/openshift/origin/pull/29466 should fix the image registry failure of `e2e-aws-serial`.

gcs278:
https://github.com/openshift/origin/pull/29466 is merged
/test e2e-aws-serial

pkg/router/template/template_helper_test.go
candita: Looks like this was taken care of.

rhamini3:
Pre-merge verified on cluster version: 4.18.0-0.ci.test-2025-02-11-173230-ci-ln-j3bmgx2-latest
```
iamin@iamin-mac cluster-ingress-operator % oc get clusterversion
version   4.18.0-0.ci.test-2025-02-11-173230-ci-ln-j3bmgx2-latest   True        False         132m    Cluster version is 4.18.0-0.ci.test-2025-02-11-173230-ci-ln-j3bmgx2-latest
```
1. annotate the route with valid and invalid IPs
```
iamin@iamin-mac cluster-ingress-operator % oc annotate route edge-route --overwrite haproxy.router.openshift.io/ip_whitelist='0.0.0.0 192.abc.0.d 192.168.47.9'
route.route.openshift.io/edge-route annotated
```
2. view the pod logs to see the invalid IP
```
iamin@iamin-mac cluster-ingress-operator % oc -n openshift-ingress logs router-default-c949f7797-fdnvt

I0211 20:32:25.636475       1 template_helper.go:392] "msg"="parseIPList found invalid IP/CIDR" "192.abc.0.d"="(MISSING)" "logger"="template"
I0211 20:32:25.636735       1 template_helper.go:370] "msg"="parseIPList empty list found" "logger"="template"
```
3. check haProxy to make sure only valid logs are present and invalid logs aren't
```
iamin@iamin-mac cluster-ingress-operator % oc -n openshift-ingress rsh  router-default-c949f7797-fdnvt
sh-5.1$ cat haproxy.config | grep edge-route
backend be_edge_http:default:edge-route
sh-5.1$ cat haproxy.config | grep edge-route -A25
backend be_edge_http:default:edge-route
  mode http
  option redispatch
  option forwardfor
  balance random
  acl allowlist src 0.0.0.0 192.168.47.9
  tcp-request content reject if !allowlist
```
4. annotate route with all invalid IPs
```
iamin@iamin-mac cluster-ingress-operator % oc annotate route edge-route --overwrite haproxy.router.openshift.io/ip_whitelist='0.0.0.e 192.abc.0.d 192.168.47.'
route.route.openshift.io/edge-route annotated
```
5. check logs to make sure all IPs are flagged as invalid
```
iamin@iamin-mac cluster-ingress-operator % oc -n openshift-ingress logs router-default-c949f7797-fdnvt                                                        

I0211 20:34:18.097056       1 template_helper.go:392] "msg"="parseIPList found invalid IP/CIDR" "0.0.0.e"="(MISSING)" "logger"="template"
I0211 20:34:18.097074       1 template_helper.go:392] "msg"="parseIPList found invalid IP/CIDR" "192.abc.0.d"="(MISSING)" "logger"="template"
I0211 20:34:18.097080       1 template_helper.go:392] "msg"="parseIPList found invalid IP/CIDR" "192.168.47."="(MISSING)" "logger"="template"
I0211 20:34:18.097085       1 template_helper.go:397] "msg"="No valid IP/CIDR in the list" "logger"="template"
I0211 20:34:18.097308       1 template_helper.go:370] "msg"="parseIPList empty list found" "logger"="template"
```
6. haProxy should not contain any IPs
```
iamin@iamin-mac cluster-ingress-operator % oc -n openshift-ingress rsh  router-default-c949f7797-fdnvt                                                        
sh-5.1$ cat haproxy.config | grep edge-route -A25
backend be_edge_http:default:edge-route
  mode http
  option redispatch
  option forwardfor
  balance random
```
marking bug as verified
/label qe-approved

rhamini3:
/remove qe-approved
logs aren't showing expected behavior

PR #274, opened 2021-03-26
Add errorfile stanzas and dummy default html files to the router

miheer:
Add errorfile stanzas and dummy default html files to the router
https://issues.redhat.com/browse/NE-379

PR #628, opened 2024-10-15
[RFE-5098] Support for customizing HAProxy error code response page for http error code 504

natlibfi-jonollil:
**I'm not sure if this is enough to make this work. Or should this be against master or next release version?**

We have noticed problems regarding error response descriptions with HA-proxy. Problems have been encountered during timeouts and whitelisting as in these cases request produces blank response and this makes debugging work difficult. After finding out the reasons for the problems, we propose adding status codes 504 and 403 to the error responses to help identifying problems with timeouts and access rights in the future.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -152,8 +152,10 @@ defaults
   # backend service or alternatively, you can send a custom 503 or 404 error.
   #
   # server openshift_backend 127.0.0.1:8080
+  errorfile 504 {{ env "ROUTER_ERRORFILE_504" "/var/lib/haproxy/conf/error-page-504.http" }}
   errorfile 503 {{ env "ROUTER_ERRORFILE_503" "/var/lib/haproxy/conf/error-page-503.http" }}
   errorfile 404 {{ env "ROUTER_ERRORFILE_404" "/var/lib/haproxy/conf/error-page-404.http" }}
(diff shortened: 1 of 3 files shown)


candita:
@natlibfi-jonollil there's a little more needed than this.  https://github.com/openshift/router/pull/274 has an example, but you'd also need to make sure the two new environment variables are implemented in the ingress controller code.  https://github.com/openshift/cluster-ingress-operator/pull/588 has an example.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -152,8 +152,10 @@ defaults
   # backend service or alternatively, you can send a custom 503 or 404 error.
   #
   # server openshift_backend 127.0.0.1:8080
+  errorfile 504 {{ env "ROUTER_ERRORFILE_504" "/var/lib/haproxy/conf/error-page-504.http" }}
   errorfile 503 {{ env "ROUTER_ERRORFILE_503" "/var/lib/haproxy/conf/error-page-503.http" }}
   errorfile 404 {{ env "ROUTER_ERRORFILE_404" "/var/lib/haproxy/conf/error-page-404.http" }}
+  errorfile 403 {{ env "ROUTER_ERRORFILE_403" "/var/lib/haproxy/conf/error-page-403.http" }}
```
candita, line 158: Also even though the lines are not currently ordered, we would ask you to order these lines by error code number: 403 404 503 504.  Note that these capitalized env variables have exist so that people can customize their error pages if they don't want the default.

candita:
@natlibfi-jonollil please let us know if you intend to pursue this feature.
/label ok-to-test

PR #631, opened 2024-10-21
NE-1815: Enable ALPN for reencrypt routes when DCM is enabled

alebedev87:
This PR ensures that ALPN protocol negotiation is enabled for reencrypt routes when the Dynamic Configuration Manager (DCM) is active, aligning the behavior with [non-DCM configurations](https://github.com/openshift/router/blob/d095bdbb4777677f5317e49fb353f41b9c1a22e7/images/router/haproxy/conf/haproxy-config.template#L728-L729).

The gap to support HTTP2 workloads for reencrypt routes was found during [the smoke testing](https://github.com/openshift/cluster-ingress-operator/pull/1113#issuecomment-2315530761).


alebedev87:
/retest

This change may not impact anything because DCM is not enabled yet.

gcs278:
Thanks for the history @Miciah. Seems like a simple oversight and a straightforward update.
/lgtm
/unhold

PR #83, opened 2020-02-10
Bug 1801407: Omit brackets around IPv4 addresses in v4v6 mode

Miciah:
Do not put brackets around an IPv4 address in the Forwarded header. [RFC 7239](https://tools.ietf.org/html/rfc7239) dictates that the host name in the header should be an IPv4 address (without brackets) or an IPv6 address with brackets.

* `images/router/haproxy/conf/haproxy-config.template`: Don't quote IPv4 addresses in v4v6 mode.

----

Not thoroughly tested as I do not have an IPv6 cluster.

PR #633, opened 2024-10-22
OCPBUGS-26603: Listen for stats connections on v6 addresses when relevant.

rfredette:
When the router is configured for ipv6 or dual stack modes, it already listens on ipv6 ports for normal traffic. This change makes the router listen on ipv6 ports for stats traffic as well.

This is part of the fix for OCPBUGS-26603.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -186,7 +186,14 @@ defaults
 
   {{ if (gt .StatsPort -1) }}
 listen stats
+    {{ if eq "v4v6" $router_ip_v4_v6_mode }}
   bind :{{ if (gt .StatsPort 0) }}{{ .StatsPort }}{{ else }}1936{{ end }}
+  bind :::{{ if (gt .StatsPort 0) }}{{ .StatsPort }}{{ else }}1936{{ end }} v6only
+    {{- else if eq "v6" $router_ip_v4_v6_mode }}
+  bind :::{{ if (gt .StatsPort 0) }}{{ .StatsPort }}{{ else }}1936{{ end }} v6only
+    {{- else }}
+  bind :{{ if (gt .StatsPort 0) }}{{ .StatsPort }}{{ else }}1936{{ end }}
+    {{- end }}
   mode http
   # Health check monitoring uri.
   monitor-uri /healthz


Miciah:
/approve
/lgtm
/hold
in case you want a second reviewer.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -186,7 +186,14 @@ defaults
 
   {{ if (gt .StatsPort -1) }}
 listen stats
+    {{ if eq "v4v6" $router_ip_v4_v6_mode }}
   bind :{{ if (gt .StatsPort 0) }}{{ .StatsPort }}{{ else }}1936{{ end }}
+  bind :::{{ if (gt .StatsPort 0) }}{{ .StatsPort }}{{ else }}1936{{ end }} v6only
```
alebedev87, line 191: I see that this implementation is similar to what we have for port `80` and `443` frontends but just out of curisosity: will this be a shorter equivalent of these 2 bind instructions:
```
bind :::{{ if (gt .StatsPort 0) }}{{ .StatsPort }}{{ else }}1936{{ end }} v4v6
```
? Or there are some pitfalls?

alebedev87 (commented):
I see that dualstack and ipv6 only stacks are already supported on HTTP and HTTPS frontends. So, it seems to me that the tests should be working (except the one which is fixed in this PR). However I saw [your comment about the ruoter pod being used as unmanaged](https://issues.redhat.com/browse/OCPBUGS-26603?focusedId=25919259&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25919259). Will we plan to fix this in origin tests too?

rfredette:
>Will we plan to fix this in origin tests too?

The origin tests do need some fixes for ipv6 support, which is going to be part of this bug fix, but I think changing them to use managed router pods is out of the scope of this issue. I'm not sure what the reason is for using unmanaged pods; maybe it's because they're older tests from when unmanaged router pods were something we supported, but it could also be an intentional choice that I don't yet understand. I'll create a task to look into that, though.

rfredette:
origin test changes here: https://github.com/openshift/origin/pull/29224

alebedev87:
LGTM
/unhold

images/router/haproxy/conf/haproxy-config.template
Miciah: Using a single `bind` line with `v4v6` is not equivalent.  See https://github.com/openshift/router/pull/83/commits/74c2d9397cb8e92c28642f56e46fd29a08d8f64b.

PR #635, opened 2024-11-04
NE-1790: Enable setting zero pool size for blueprint routes via environment variable

alebedev87:
- Allows setting blueprint route pool size to zero through an environment variable.
- Aligns environment variable configuration with the `--blueprint-route-pool-size` flag for consistency.


Miciah:
Would you mind adding a note in the commit message that this change makes the `ROUTER_BLUEPRINT_ROUTE_POOL_SIZE` environment variable consistent the `--blueprint-route-pool-size` flag, which already allows 0 as the value?  That would make it clear that the commit does not add any new functionality.

Miciah:
Thanks!
/lgtm

alebedev87:
/label docs-approved
/label px-approved

Technical limitation lifted, nothing to be documented.

ShudiLi:
/label qe-approved
thanks

PR #636, opened 2024-11-05
NE-1815: Activate dynamic cookies for dynamic server slots (DCM)

alebedev87:
- Introduced a "dynamic" option for ["cookie" directive](https://docs.haproxy.org/2.8/configuration.html#4-cookie) used on plain HTTP, edge and re-encrypt routes. This option activates the dynamic cookie creation per server, applicable only for dynamic server slots (static server slots set cookie values explicitly). Without this option, dynamic servers will not set a cookie value.

- Added the ["dynamic-cookie-key" directive](https://docs.haproxy.org/2.8/configuration.html#dynamic-cookie-key) to re-ecrypt routes to align the dynamic cookie handling across all route types (plain HTTP and edge already have "dynamic-cookie-key").This directive is used to generate a unique cookie value for a each route.

`Set-Cookie` header sent from dynamic servers before this change:
```
set-cookie: a84fef2b01006bb7036c768d4a819999=; Expires=Thu, 01-Jan-1970 00:00:01 GMT; path=/; HttpOnly
```

and after:
```
set-cookie: a84fef2b01006bb7036c768d4a819999=424f3944a7ac32ec; path=/; HttpOnly
```

**Note**: the dynamic cookie value is a 16 chars long hash which is different from the cookie set on the static servers (32 chars).

Previous PR on the same topic (dynamic cookies for DCM slots): https://github.com/openshift/origin/pull/20557. This PR introduced "dynamic-cookie-key" directive for plain http/edge and passthrough backends in order to be able to disable the dynamic cookies for blueprint routes buy default and re-enable after a reload.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -703,6 +703,8 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
               {{- "" }} secure attr SameSite={{ $samesite }}
             {{- end }}
           {{- end }}
+          {{- with $dynamicConfigManager }} dynamic
+          {{- end }}
         {{- end }}{{/* end disable cookies check */}}
 
         {{- if matchValues (print $cfg.TLSTermination) "edge" "reencrypt" }}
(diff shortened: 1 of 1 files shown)


gcs278:
@alebedev87 Generally everything looks fine to me. Tested it out too.

But do you know why they [put](https://github.com/openshift/origin/pull/20557/files#diff-21bd64a4ad8a7927e9d59e99e09628efc884fe6d6291e50eb33c5deb3d2a79f0R570) `dynamic-cookie-key` in the passthrough backend section? Passthrough operates on layer 4 (except for reading SNI...), so it can't use cookies. What's the point of having that config in the backend for passthrough?

alebedev87:
> But do you know why they [put](https://github.com/openshift/origin/pull/20557/files#diff-21bd64a4ad8a7927e9d59e99e09628efc884fe6d6291e50eb33c5deb3d2a79f0R570) dynamic-cookie-key in the passthrough backend section? Passthrough operates on layer 4 (except for reading SNI...), so it can't use cookies. What's the point of having that config in the backend for passthrough?

No idea. I agree, I don't think it makes any sense on passthrough routes. I didn't want to remove anything though, just to make the PR easier and less "destructive".

gcs278:
> I didn't want to remove anything though, just to make the PR easier and less "destructive".

That's fair. We can clean up later, but it isn't hurting anything right now.

/lgtm
/approve

PR #627, opened 2024-10-10
OCPBUGS-42972: Support appProtocol: kubernetes.io/h2c

Miciah:
#### Rename `mustCreate` to `mustCreateRoute`

Rename the `mustCreate` type used in tests to `mustCreateRoute`.

#### Add test cases for `appProtocol`

Define a new `mustCreateEndpointSlice` type for tests, and use it to define new test cases where the endpointslice associated with the route specifies `appProtocol` with the recognized value "h2c" and unrecognized values "unknown-value" and "kubernetes.io/h2c".

#### Support `appProtocol: kubernetes.io/h2c`

Recognize both "h2c" and "kubernetes.io/h2c" as valid values for `appProtocol`.  Update the expectations for the "route with appProtocol: kubernetes.io/h2c" test case accordingly. 

--- 

Follow-up to  #328.

PR #637, opened 2024-11-08
NE-1815: Add verifyhost to dynamic server slots for re-encrypt routes (DCM)

alebedev87:
This PR introduces  [the "verifyhost" configuration setting](https://docs.haproxy.org/2.8/configuration.html#verifyhost) to dynamic server slots during template rendering. This ensures consistent behavior with static servers.

The "verifyhost" uses the FQDN of the primary service only, as it's not feasible to predict which service endpoint will be used when alternate backends are available.Consequently, dynamic servers for re-encrypt routes are not intended to handle alternate backend endpoints. A reload is forced when endpoint changes occur for alternate backend services in a re-encrypt route.

The commit which added the server certificate verification without `verifyhost`: https://github.com/openshift/router/commit/50a330de5a3da9636b36b5e9e62ddbc374b99524.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -769,6 +769,15 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
               {{- else }}
                 {{- if not (isTrue $router_disable_http2) }} alpn h2,http/1.1
                 {{- end }}
+                {{- /*
+                    Add "verifyhost" for the primary service only. This server is not intended
+                    to be used for alternate backend endpoints. A reload will be triggered if
+                    an alternate backend endpoint is added dynamically.
+                */ -}}
+                {{- with $serviceUnit := index $.ServiceUnits $cfg.PrimaryServiceUnitKey -}}
+                    {{- if $cfg.VerifyServiceHostname }} verifyhost {{ $serviceUnit.Hostname }}
+                    {{- end }}
+                {{- end }}
                 {{- if gt (len $defaultDestinationCA) 0 }} verify required ca-file {{ $defaultDestinationCA }}
                 {{- else }} verify none
                 {{- end }}
--- a/pkg/router/router_test.go
+++ b/pkg/router/router_test.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"reflect"
 	"strconv"
 	"strings"
 	"sync"
@@ -42,6 +41,7 @@ import (
 	"github.com/openshift/router/pkg/router"
 	"github.com/openshift/router/pkg/router/controller"
 	templateplugin "github.com/openshift/router/pkg/router/template"
+	haproxyconfigmanager "github.com/openshift/router/pkg/router/template/configmanager/haproxy"
 	"github.com/openshift/router/pkg/router/writerlease"
 )
 
@@ -153,9 +153,13 @@ u3YLAbyW/lHhOCiZu2iAI8AbmXem9lW6Tr7p/97s0w==
 -----END RSA PRIVATE KEY-----
 `,
 		DefaultCertificateDir: h.dirs["certs"],
-		ReloadFn:              func(shutdown bool) error { return nil },
-		TemplatePath:          "../../images/router/haproxy/conf/haproxy-config.template",
-		ReloadInterval:        reloadInterval,
+		// DefaultDestinationCAPath enables the host verification ("verifyhost")
+		// based on the service hostname for re-encrypt routes without
+		// DestinationCACertificate set.
+		DefaultDestinationCAPath: "dummy",
+		ReloadFn:                 func(shutdown bool) error { return nil },
+		TemplatePath:             "../../images/router/haproxy/conf/haproxy-config.template",
+		ReloadInterval:           reloadInterval,
 		HTTPResponseHeaders: []templateplugin.HTTPHeader{{
 			Name:   "x-foo",
 			Value:  "'bar'",
@@ -167,6 +171,10 @@ u3YLAbyW/lHhOCiZu2iAI8AbmXem9lW6Tr7p/97s0w==
 			Action: "Set",
 		}},
 		SecretManager: &fakesm.SecretManager{},
+		DynamicConfigManager: haproxyconfigmanager.NewHAProxyConfigManager(templateplugin.ConfigManagerOptions{
+			ConnectionInfo:    "unix:///var/lib/dymmy",
+			MaxDynamicServers: 1,
+		}),
 	}
 	plugin, err = templateplugin.NewTemplatePlugin(pluginCfg, svcFetcher)
 	if err != nil {
@@ -670,10 +678,12 @@ func TestConfigTemplate(t *testing.T) {
 		},
 		"route with appProtocol: unknown-value": {
 			mustCreateWithConfig{
-				mustCreateEndpointSlice: mustCreateEndpointSlice{
-					name:        "servicep1",
-					serviceName: "servicep1",
-					appProtocol: "unknown-value",
+				mustCreateEndpointSlices: []mustCreateEndpointSlice{
+					{
+						name:        "servicep1",
+						serviceName: "servicep1",
+						appProtocol: "unknown-value",
+					},
 				},
 				mustCreateRoute: mustCreateRoute{
 					name:              "p1",
@@ -692,10 +702,12 @@ func TestConfigTemplate(t *testing.T) {
 		},
 		"route with appProtocol: h2c": {
 			mustCreateWithConfig{
-				mustCreateEndpointSlice: mustCreateEndpointSlice{
-					name:        "servicep2",
-					serviceName: "servicep2",
-					appProtocol: "h2c",
+				mustCreateEndpointSlices: []mustCreateEndpointSlice{
+					{
+						name:        "servicep2",
+						serviceName: "servicep2",
+						appProtocol: "h2c",
+					},
 				},
 				mustCreateRoute: mustCreateRoute{
 					name:              "p2",
@@ -713,10 +725,12 @@ func TestConfigTemplate(t *testing.T) {
 		},
 		"route with appProtocol: kubernetes.io/h2c": {
 			mustCreateWithConfig{
-				mustCreateEndpointSlice: mustCreateEndpointSlice{
-					name:        "servicep3",
-					serviceName: "servicep3",
-					appProtocol: "kubernetes.io/h2c",
+				mustCreateEndpointSlices: []mustCreateEndpointSlice{
+					{
+						name:        "servicep3",
+						serviceName: "servicep3",
+						appProtocol: "kubernetes.io/h2c",
+					},
 				},
 				mustCreateRoute: mustCreateRoute{
 					name:              "p3",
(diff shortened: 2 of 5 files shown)


alebedev87:
> Force a reload if the endpoints change for a multi-service route. This is necessary to ensure that we write out server lines with verifyhost and the correct host name for multi-service routes.

This would be the safest, yes. Let me see whether I can find a way to do this.

alebedev87:
/retitle [WIP] NE-1815: Add verifyhost to dynamic server slots for re-encrypt routes (DCM)

Work In Progress until https://github.com/openshift/router/pull/637#issuecomment-2465041632 is addressed.

alebedev87:
> Force a reload if the endpoints change for a multi-service route. This is necessary to ensure that we write out server lines with verifyhost and the correct host name for multi-service routes.

Addressed [here](https://github.com/openshift/router/pull/637/files#diff-2d8eb06e37fa38846f1923194ac1079fb4d811d75e57c501a329dff5b886b0ea).

alebedev87:
Changed `TestConfigTemplate` to converge to the approach from https://github.com/openshift/router/pull/627.

alebedev87:
Rebased from `master` after https://github.com/openshift/router/pull/627 was merged.

pkg/router/template/router.go
```diff
@@ -878,6 +878,14 @@ func (r *templateRouter) dynamicallyReplaceEndpoints(id ServiceUnitKey, service
 			weight = 0
 		}
 
+		if cfg.TLSTermination == routev1.TLSTerminationReencrypt && cfg.VerifyServiceHostname && cfg.ActiveServiceUnits > 1 {
```
gcs278, line 881: Do you need to check for reencrypt? `cfg.VerifyServiceHostname` is only set to true if it's reencrypt via:
https://github.com/openshift/router/blob/9779c5eed8d4f327e7921eda09f011bc33b3e1c4/pkg/router/template/router.go#L1020-L1022

```suggestion
		if cfg.VerifyServiceHostname && cfg.ActiveServiceUnits > 1 {
```

gcs278: Also, is there any reason this can't be earlier in the function? it's within the for loop of `service.ServiceAliasAssociations`, but doesn't use anything from the iteration.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -769,6 +769,23 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
...
+                {{- /*
+                    Add "verifyhost" with service hostname only if both are true:
+                    - The route has a single active service.
+                    - The route does not have a destination CA certificate.
+
+                    Limiting "verifyhost" to a single service route is essential,
+                    as it's not possible to reliably predict which service’s endpoint will be used
+                    when alternate backends are available.
+                */ -}}
+                {{- if and $cfg.VerifyServiceHostname (eq $cfg.ActiveServiceUnits 1) -}}
```
gcs278, line 781: So if `cfg.ActiveServiceUnits > 1`, then this entire reencrypt backend is effectively not dynamic (that's unfortunate). However, the config still shows up with the dynamic servers:
```
  server _dynamic-pod-1 172.4.0.4:8765 weight 0 ssl disabled check inter 5000ms verify required ca-file /var/run/configmaps/service-ca/service-ca.crt
  server _dynamic-pod-2 172.4.0.4:8765 weight 0 ssl disabled check inter 5000ms verify required ca-file /var/run/configmaps/service-ca/service-ca.crt
  server _dynamic-pod-3 172.4.0.4:8765 weight 0 ssl disabled check inter 5000ms verify required ca-file /var/run/configmaps/service-ca/service-ca.crt
```

Did you consider moving this condition up so that we don't even print any dynamic server lines? They are totally unusable right? I know it's mostly a NO-OP, but it might be more explicit when debugging the config to not mislead folks into thinking this dynamic servers are actually useful.

EDIT: See my next comment since this might be OBE.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -769,6 +769,23 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
...
+                    {{- range $serviceUnitName, $weight := $cfg.ServiceUnitNames -}}
+                        {{- with $serviceUnit := index $.ServiceUnits $serviceUnitName -}}
+                            {{- print " verifyhost " $serviceUnit.Hostname }}
+                            {{- break }}
+                        {{- end }}
+                    {{- end }}
+                {{- end }}
```
gcs278, line 788: I don't know what our philosophy is on DCM, or how much efficiency we are trying to squeeze out of it, but did you consider the strategy of always assuming dynamic servers are reserved for the primary backend (not the alternates), and change your logic to only reload if an alternate backend endpoints needs syncing?

While it might be an asymmetrical efficiency gain, it'd be a net gain to avoid reloading for primary backend endpoint changes. Net positive, but yes, asymmetrical gains (primary wins while alternate lose). Also depends the added complexity and whether that provides reasonable ROI. Not trying to demand we need to be as perfect as possible, but I'm just curious.

pkg/router/router_test.go
```diff
@@ -732,23 +746,95 @@ func TestConfigTemplate(t *testing.T) {
...
+				},
+				mustMatchConfig: mustMatchConfig{
+					section:     "backend",
+					sectionName: reencryptBackendName(h.namespace, "q3"),
+					attribute:   "server",
+					value:       "_dynamic-pod-1 172.4.0.4:8765 weight 0 ssl disabled check inter 5000ms alpn h2,http/1.1 verify required ca-file dummy",
+					fullMatch:   true,
+				},
+			},
+		},
```
gcs278, line 828: Easy way to test VerifyHost doesn't get added with `destinationCACertificate`?

gcs278 (commented):
Nice catch finding this and I appreciate the code comments.

pkg/router/template/router.go
gcs278: > Right, it's redundant. I just wanted to highlight the fact that it only impacts reencrypt routes.

That's fine.

> It uses cfg (type ServiceAliasConfig) which is retrieved from the state. This loop is supposed to iterate over all the routes associated with a given service. Also, I wanted this condition to be close to the place where we call ReplaceRouteEndpoints of DCM.

Whoops, sorry I missed the `cfg`.

images/router/haproxy/conf/haproxy-config.template
gcs278: > Upd: just recalled that the main problem was to bring the notion of alternate service to the template, all service units of a route are the same for the moment.

I see, it'd require more refactoring than probably it's worth for the added efficiency since we don't distinguish between the primary backend and alternate backends. It's something we can keep in mind for later improvements.

gcs278:
I think the approach is reasonable given the constraints, and I've done some testing of my own with reencrypt routes and DCM to make sure it works as expected. I think there are some potential future refactoring we should do with the unit test in `router_test.go`, such as maybe separate the DCM plugin testing into a different package (might be some future overlap that cause issues), but it seems reasonable given the state of DCM.

/lgtm

@Miciah I'll yield to you for review and approval.

images/router/haproxy/conf/haproxy-config.template
alebedev87: [This commit](https://github.com/openshift/router/pull/637/commits/17d42d1cc1c16523420bf6c98d7932545e3d9bc5) can do what you proposed.

pkg/router/router_test.go
```diff
@@ -732,23 +746,69 @@ func TestConfigTemplate(t *testing.T) {
...
+				mustMatchConfig: mustMatchConfig{
+					section:     "backend",
+					sectionName: reencryptBackendName(h.namespace, "q"),
+					attribute:   "server",
+					value:       "_dynamic-pod-1 172.4.0.4:8765 weight 0 ssl disabled check inter 5000ms alpn h2,http/1.1 verifyhost serviceq.default.svc verify required ca-file dummy",
+					fullMatch:   true,
+				},
+			},
+		},
+		"No verifyhost for dynamic slot when alternate backend is used": {
```
gcs278, line 774:
```suggestion
		"Verifyhost for dynamic slot uses primary backend when alternate backend is present": {
```

images/router/haproxy/conf/haproxy-config.template
gcs278: :+1: Looks good, doesn't seem to add too much complexity, and it made the fix slightly more efficient.

gcs278:
@alebedev87 Your recent updates look good to me, just one last minor comment. You will need to update your PR description if you plan to keep the commit (and presumably squash if you want to keep it clean).

alebedev87:
> Your recent updates look good to me, just one last minor comment. You will need to update your PR description if you plan to keep the commit (and presumably squash if you want to keep it clean).

Thanks a lot! The commits are fixed up, the description is updated.

alebedev87:
/label docs-approved
/label px-approved

Technical gap, no need to ping the docs people.

alebedev87:
/label acknowledge-critical-fixes-only

This change impacts only the DCM enabled router which is not the case by default. [The featuregate](https://github.com/openshift/cluster-ingress-operator/pull/1159) needs to be enabled to activate DCM.

alebedev87:
/test e2e-aws-serial

`e2e-aws-serial` passed the test but the whole job took too long and timed out:

```
INFO[2024-11-20T13:41:05Z] Step phase test succeeded after 2h29m19s.    
INFO[2024-11-20T13:41:05Z] Running multi-stage phase post               
INFO[2024-11-20T13:41:05Z] Running step e2e-aws-serial-gather-aws-console. 
{"component":"entrypoint","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:169","func":"sigs.k8s.io/prow/pkg/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 4h0m0s timeout","severity":"error","time":"2024-11-20T13:41:22Z"}
```

gcs278:
Thanks for the updates & the responses.
/lgtm

Will let @Miciah approve.

pkg/router/router_test.go
```diff
@@ -732,23 +746,69 @@ func TestConfigTemplate(t *testing.T) {
...
+					{
+						name:        "serviceq",
+						serviceName: "serviceq",
+					},
+				},
+				mustCreateRoute: mustCreateRoute{
+					name:              "q",
+					host:              "qexample.com",
+					targetServiceName: "serviceq",
+					weight:            int32(100),
```
Miciah, line 761: Why is `weight` needed in this test case?  Is it used just to verify that the dynamic server does not start with the route weight?

alebedev87: No, I think I did it for consistency with the alternate backend case which I wanted to look "real". None of them really influence the template rendering: `weight` is always `0`.

pkg/router/router_test.go
```diff
@@ -842,6 +907,21 @@ func (e mustCreateRoute) Apply(h *harness) error {
...
+	alternateBackends := []routev1.RouteTargetReference{}
+	if e.alternateBackend != "" {
+		fifty := int32(50)
+		alternateBackends = []routev1.RouteTargetReference{
+			{
+				Name:   e.alternateBackend,
+				Weight: &fifty,
+			},
+		}
+		weight = &fifty
```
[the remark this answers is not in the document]
alebedev87: What you think is not correct? I wanted the 50-50 balance between 2 services.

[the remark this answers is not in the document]
alebedev87: Added a dedicated field for the alternate backend weight.

pkg/router/router_test.go
```diff
@@ -842,6 +911,23 @@ func (e mustCreateRoute) Apply(h *harness) error {
 	if e.targetServiceName != "" {
 		serviceName = e.targetServiceName
 	}
+	weight := new(int32)
```
Miciah, line 914: Not a big deal, but is there a reason you use `weight := new(int32)` instead of using `var weight int32`?

pkg/router/router_test.go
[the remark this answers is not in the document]
Miciah: I see, I didn't understand your intention before, so I'm glad I asked.  Thanks for changing the test expectation to be more explicit!

pkg/router/router_test.go
alebedev87: No, I was just inspired by the previous code which did this: `Weight: new(int32)`.

Miciah:
Thanks!
/approve
/lgtm

alebedev87:
/retest

Tests are not starting for an hour.

Miciah:
Thanks!  
/approve
/lgtm
(Hm, I thought I already posted this comment.)

PR #638, opened 2024-11-14
OCPBUGS-43745: force http-keep-alive to 1s

frobware:


frobware:
Another sample required.

/test perfscale-aws-ingress-perf

PR #640, opened 2024-11-27
[WIP] HAProxy config check

gcs278:
Swap TemplatePlugin and StatusAdmitter ordering so that StatusAdmitter is the last plugin so that the TemplatePlugin can reject a route based on whether HAProxy rejects the configuration.

When TemplatePlugin adds a route, it also writes the config file without reloading, and runs the configuration check script. If the configuration check script fails with the new route, the route will be removed from the state and it will be rejected.

Example failure for https://issues.redhat.com//browse/OCPBUGS-27741 looks like:
```yaml
status:
  ingress:
  - conditions:
    - lastTransitionTime: "2024-11-27T16:40:59Z"
      message: Failed to validate HAProxy config with route. Review the openshift-router
        logs for more details on the validation failure.
      reason: HAProxyCheckConfigFailed
      status: "False"
      type: Admitted
    host: route-service1-default.apps.gspence-2024-11-27-0943.devcluster.openshift.com
    routerCanonicalHostname: router-default.apps.gspence-2024-11-27-0943.devcluster.openshift.com
    routerName: default
    wildcardPolicy: None
```

WIP:
- Additional unit testing coverage
- Test & confirm that writing the config before reloading is OKAY
- Research impact of swapping StatusAdmitter and TemplatePlugin ordering
- Understand performance impact of running `haproxy -c` on every route add or change.

### DIFF
--- a/images/router/haproxy/Dockerfile
+++ b/images/router/haproxy/Dockerfile
@@ -17,5 +17,6 @@ USER 1001
 EXPOSE 80 443
 WORKDIR /var/lib/haproxy/conf
 ENV TEMPLATE_FILE=/var/lib/haproxy/conf/haproxy-config.template \
-    RELOAD_SCRIPT=/var/lib/haproxy/reload-haproxy
+    RELOAD_SCRIPT=/var/lib/haproxy/reload-haproxy \
+    CONFIG_CHECK_SCRIPT=/var/lib/haproxy/check-haproxy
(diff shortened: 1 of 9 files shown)


gcs278:
This was a fun experiment, but this is a complex and risky change that needs more work done understanding performance impacts, and it's likely this shouldn't merge in it's current state:
/close

PR #661, opened 2025-06-27
OCPBUGS-58040: haproxy-config.template: Allow empty ciphers

Miciah:
#### `haproxy-config.template`: Simplify `ROUTER_CIPHERS`

Simplify the default ciphers configuration in the configuration template by combining `{{else}}` and `{{if}}` stanzas into `{{else if}}`.

#### `haproxy-config.template`: Introduce `$routerCiphers`

Define a template variable `$routerCiphers` and use it to avoid unnecessarily calling the `env` helper repeatedly.

#### Introduce `$routerCiphersuites`

Define a template variable `$routerCiphersuites` for consistency with `$routerCiphers`.

#### `haproxy-config.template`: Allow empty ciphers

Allow the empty value for TLS ciphers (TLSv1.2 and earlier) and cipher suites (TLSv1.3).  Rather than filling in a default value, pass the empty value through to HAProxy, which in turn passes it through to OpenSSL.  Specifying the empty value means that the corresponding TLS version cannot be used.


gcs278:
Sorry, the PR sync tool assigned this one to yourself @Miciah in the initial syncing from the PR Scrub document. Removing, and moving back to new.

PR #667, opened 2025-08-26
NE-2116: Update haproxy-config.template to use HTTPS/TCP log formats

rohara:
The haproxy-config.template must be updated to allow configuration of both HTTP and TCP log formats. Previously, only HTTP log format was configurable, and if not set, the default 'option httplog' was used. This logic is now pushed down into each frontend. The basic idea is that if access logging is enabled, we then look at the respective log format environment variable, then use either the custom log format or the default log format.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -227,6 +227,12 @@ listen stats
 
   {{ if .BindPorts -}}
 frontend public
+    {{- if ne (env "ROUTER_SYSLOG_ADDRESS") "" }}
+      {{- if ne (env "ROUTER_HTTP_LOG_FORMAT") "" }}
+  log-format {{ env "ROUTER_HTTP_LOG_FORMAT" }}
+      {{- else }}
+  option httplog
+    {{- end }}
     {{ if eq "v4v6" $router_ip_v4_v6_mode }}
   bind :{{ env "ROUTER_SERVICE_HTTP_PORT" "80" }}{{ if isTrue (env "ROUTER_USE_PROXY_PROTOCOL") }} accept-proxy{{ end }}
   bind :::{{ env "ROUTER_SERVICE_HTTP_PORT" "80" }} v6only{{ if isTrue (env "ROUTER_USE_PROXY_PROTOCOL") }} accept-proxy{{ end }}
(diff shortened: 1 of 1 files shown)


rikatz:
@rohara I was having some chat with @alebedev87 and @Miciah about testing templates, are you considering adding some cases on https://github.com/openshift/router/blob/master/pkg/router/router_test.go using t.SetEnv to assert that the template rendering is working?

I don't have comments about the feature/directives, will leave this for someone with more context!

Thanks!

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -227,6 +227,12 @@ listen stats
 
   {{ if .BindPorts -}}
 frontend public
+    {{- if ne (env "ROUTER_SYSLOG_ADDRESS") "" }}
+      {{- if ne (env "ROUTER_HTTP_LOG_FORMAT") "" }}
+  log-format {{ env "ROUTER_HTTP_LOG_FORMAT" }}
+      {{- else }}
+  option httplog
+    {{- end }}
```
Miciah, line 235: This does not look syntactically correct: You add two `if`s and only one `end`.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -461,6 +476,12 @@ backend be_no_sni
   server fe_no_sni unix@/var/lib/haproxy/run/haproxy-no-sni.sock weight 1 send-proxy
 
 frontend fe_no_sni
+  {{- if ne (env "ROUTER_SYSLOG_ADDRESS") "" }}
+    {{- if ne (env "ROUTER_HTTPS_LOG_FORMAT") "" }}
+  log-format {{ env "ROUTER_HTTPS_LOG_FORMAT" }}
+    {{- else }}
+  option httsplog
+  {{- end }}
```
Miciah, line 484: Missing `end`.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -227,6 +227,12 @@ listen stats
 
   {{ if .BindPorts -}}
 frontend public
+    {{- if ne (env "ROUTER_SYSLOG_ADDRESS") "" }}
```
Miciah, line 230: Consider defining `{{- $syslogAddress := env "ROUTER_SYSLOG_ADDRESS" }}`, `{{- $httpLogFormat := env "ROUTER_HTTP_LOG_FORMAT" }}`, and so on for the following reasons:

- Less performance hit from repeatedly calling into Go helpers.
- Less risk of typos in the environment variable's name.
- Slightly simpler and more readable `if` conditions.

Miciah:
The PR description has some useful detail that is not in the commit message.  Please do make sure that the commit message has a clear description.

images/router/haproxy/conf/haproxy-config.template
ShudiLi, line 484: httsplog should be httpslog?

PR #669, opened 2025-09-01
OCPBUGS-49769: Validate the key and cert content before accepting

rikatz:
This change enforces the validation that a private key does not contain any certificate/public key, and that a public key does not contain any private key.

This is required because during the .pem file assembling for HAProxy, in case a certificate is present on the private key field, it will be ignored by the PEM sanitization function.

This lack of validation can cause a situation where the .pem file of HAProxy will be created as private key -> CA chain -> leaf certificate, which is invalid and rejected by HAProxy.

As the right order is private key -> leaf certificate -> CA Chain, and the proper ordering of the public key/CA certificate is already guaranteed by X509KeyPair(), we can reject the misusage of private keys or certificates on the wrong fields, that can cause HAProxy to stop

Additionally, some tests were added to validate that PEM files exported from a P12 and that contain bag attributes like friendlyNames are valid and that the process of sanitization of these files will clean them correctly and process they rest of the certificates and private keys.

As an example, the lack of validation would cause the following invalid PEM file to be added to HAProxy and make it get stuck on reconfiguration:

```
-----BEGIN RSA PRIVATE KEY-----
.....
-----END RSA PRIVATE KEY-----
-----BEGIN CERTIFICATE-----
.... this is the CA chain and not the leaf cert
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
... this is the leaf cert but it should happen right after the private key
-----END CERTIFICATE-----
```

EDIT: test executed:
* created an invalid ingress with invalid key (the key contains the CA chain)

```
router-default-6f5d7fcdfb-rs5bg router E0901 19:25:43.585955       1 extended_validator.go:52] "msg"="skipping route due to invalid configuration" "error"="spec.tls.key: Invalid value: \"redacted key data\": field contains invalid types CERTIFICATE, expecting only PRIVATE KEY" "logger"="controller" "route"="ricardo2/test2-jnszn"
router-default-6f5d7fcdfb-rs5bg router E0901 19:25:43.586000       1 router_controller.go:273] invalid route configuration
router-default-6f5d7fcdfb-bghqr router E0901 19:25:43.585863       1 router_controller.go:273] invalid route configuration
router-default-6f5d7fcdfb-rs5bg router E0901 19:25:43.599418       1 extended_validator.go:52] "msg"="skipping route due to invalid configuration" "error"="spec.tls.key: Invalid value: \"redacted key data\": field contains invalid types CERTIFICATE, expecting only PRIVATE KEY" "logger"="controller" "route"="ricardo2/test2-jnszn"
router-default-6f5d7fcdfb-rs5bg router E0901 19:25:43.599449       1 router_controller.go:273] invalid route configuration
router-default-6f5d7fcdfb-bghqr router E0901 19:25:43.598906       1 extended_validator.go:52] "msg"="skipping route due to invalid configuration" "error"="spec.tls.key: Invalid value: \"redacted key data\": field contains invalid types CERTIFICATE, expecting only PRIVATE KEY" "logger"="controller" "route"="ricardo2/test2-jnszn"
router-default-6f5d7fcdfb-bghqr router E0901 19:25:43.598939       1 router_controller.go:273] invalid route configuration
```

* check the route:
```
kubectl get route
NAME          HOST/PORT                  PATH   SERVICES     PORT    TERMINATION     WILDCARD
test2-jnszn   ExtendedValidationFailed   /      hello-node   <all>   edge/Redirect   None
....
status:
  ingress:
  - conditions:
    - lastTransitionTime: "2025-09-01T19:25:43Z"
      message: 'spec.tls.key: Invalid value: "redacted key data": field contains invalid
        types CERTIFICATE, expecting only PRIVATE KEY'
      reason: ExtendedValidationFailed
      status: "False"
      type: Admitted
    host: something2-ricardo2.apps.lab.tld
    routerCanonicalHostname: router-default.apps.lab.tld
    routerName: default
    wildcardPolicy: None
```

### DIFF
--- a/pkg/router/routeapihelpers/validation.go
+++ b/pkg/router/routeapihelpers/validation.go
@@ -9,6 +9,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"strings"
 
 	"k8s.io/apiserver/pkg/authentication/user"
 	"k8s.io/client-go/util/cert"
@@ -251,6 +252,14 @@ func ExtendedValidateRoute(route *routev1.Route) field.ErrorList {
 			if len(keyBytes) == 0 {
 				result = append(result, field.Invalid(tlsFieldPath.Child("key"), "", "no key specified"))
 			} else {
+				// Validate that final key contains only private key, and cert contains only public keys
+				if err := validatePEMContent(keyBytes, "PRIVATE KEY"); err != nil {
+					result = append(result, field.Invalid(tlsFieldPath.Child("key"), "redacted key data", err.Error()))
+				}
+				if err := validatePEMContent(certBytes, "CERTIFICATE"); err != nil {
+					result = append(result, field.Invalid(tlsFieldPath.Child("certificate"), "redacted certificate data", err.Error()))
+				}
+				// Validate if the keypair is valid (eg.: the leaf certificate should be the first on certBytes)
 				if _, err := tls.X509KeyPair(certBytes, keyBytes); err != nil {
 					result = append(result, field.Invalid(tlsFieldPath.Child("key"), "redacted key data", err.Error()))
 				} else {
@@ -279,6 +288,31 @@ func ExtendedValidateRoute(route *routev1.Route) field.ErrorList {
 	}
 
 	return result
+}
+
+// validatePEMContent takes content and pemType, PEM decodes content and
+// validates that the first block matches the expected pemType. This validation
+// 1. Ensures that the required pemType is first in the order
+// 2. Blocks attempts of passing certificates where keys should be used, and
+// 3. Blocks attempts of passing keys where certificates should be used.
+// Passing an out of order certificate, or key as certificate, or certificate as key
+// breaks HAProxy.  Note that the match of content and pemType is not exact, but
+// content must CONTAIN the expected pemType.
+func validatePEMContent(content []byte, pemType string) error {
+	if len(content) == 0 {
+		return fmt.Errorf("the PEM content cannot be null")
+	}
+	for len(content) > 0 {
+		var pemBlock *pem.Block
+		pemBlock, content = pem.Decode(content)
+		if pemBlock == nil {
+			break
+		}
+		if !strings.Contains(pemBlock.Type, pemType) {
+			return fmt.Errorf("field contains invalid types %s, expecting only %s", pemBlock.Type, pemType)
+		}
+	}
+	return nil
 }
 
 // validateTLS tests fields for different types of TLS combinations are set.  Called
--- a/pkg/router/routeapihelpers/validation_test.go
+++ b/pkg/router/routeapihelpers/validation_test.go
@@ -3,6 +3,7 @@ package routeapihelpers
 import (
 	"bytes"
 	"reflect"
+	"strings"
 	"testing"
 
 	routev1 "github.com/openshift/api/route/v1"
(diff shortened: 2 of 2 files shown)


pkg/router/routeapihelpers/validation_test.go
```diff
@@ -2316,6 +2479,61 @@ func TestExtendedValidateRoute(t *testing.T) {
...
+				Spec: routev1.RouteSpec{
+					TLS: &routev1.TLSConfig{
+						Termination: routev1.TLSTerminationEdge,
+						Certificate: leafCertExtractedFromP12,
+						Key:         leafKeyExtractedFromP12 + caChainExtractedFromP12,
+					},
+				},
+			},
+			expectedErrors: 1,
+		},
```
alebedev87, line 2523: So, this test case is supposed to address the bug in the title, right? The rest of tests cases are here just to confirm the behavior which existed before this PR.

alebedev87: > I have generated them as p12 with bag attributes to follow that very specific case as well.

> that creating the tls.crt usingthe wrong order will be rejected (my change doesn't do it, but it is done by X509 validation from Go, so just be sure this case is covered and doesn't regress)

> that the legacy bundled certificate doesn't regress

Thanks for adding more test scenarios!

> that the case from this bug is covered (private key containing certificates)

Right. I asked the question because from reading the bug comments it didn't seem 100% certain that this is the only issue with the customer configuration.

alebedev87: No, the PR description LGTM.

rikatz:
/hold cancel

As aligned with Miciah, the e2e and OTE are a separate story and implementation

pkg/router/routeapihelpers/validation_test.go
```diff
@@ -2316,6 +2479,92 @@ func TestExtendedValidateRoute(t *testing.T) {
...
+						Key:         leafKeyExtractedFromP12 + testCertificateRsaSha256Key,
+					},
+				},
+			},
+			expectedErrors: 0,
+		},
+		{
+			// A private key containing 2 keys, one wrong and one right will be rejected because
+			// only the first key is considered and it doesn't match the cert
+			name: "A private key containing 2 keys on wrong order should be rejected",
```
candita, line 2556:
```suggestion
			name: "A private key containing 2 keys is rejected if the first key is invalid",
```

pkg/router/routeapihelpers/validation.go
```diff
@@ -251,6 +252,14 @@ func ExtendedValidateRoute(route *routev1.Route) field.ErrorList {
 			if len(keyBytes) == 0 {
 				result = append(result, field.Invalid(tlsFieldPath.Child("key"), "", "no key specified"))
 			} else {
+				// Validate that final key contains only private key, and cert contains only public keys
+				if err := validatePEMContent(keyBytes, "PRIVATE KEY"); err != nil {
+					result = append(result, field.Invalid(tlsFieldPath.Child("key"), "redacted key data", err.Error()))
+				}
+				if err := validatePEMContent(certBytes, "CERTIFICATE"); err != nil {
+					result = append(result, field.Invalid(tlsFieldPath.Child("certificate"), "redacted key data", err.Error()))
```
candita, line 260: nit:
```suggestion
					result = append(result, field.Invalid(tlsFieldPath.Child("certificate"), "redacted cert data", err.Error()))
```

pkg/router/routeapihelpers/validation.go
```diff
@@ -281,6 +290,27 @@ func ExtendedValidateRoute(route *routev1.Route) field.ErrorList {
...
+// the expected type. This validation is necessary to block attempts of passing certificates
+// inside the field that should be private key only, and vice versa. This is required
+// because otherwise passing an out of order certificate to HAProxy will break it.
+// The function will take the array of bytes of a key/cert, and the expected type
+// that the PEM should content. The match here is not strict, but will check if the
+// passed byte type CONTAINS the expected pemType
+func validatePEMContent(content []byte, pemType string) error {
+	for {
+		var pemBlock *pem.Block
+		pemBlock, content = pem.Decode(content)
```
candita, line 303: Does this need a check, just in case the content is empty?
```suggestion
        if len(content) > 0 {
		pemBlock, content = pem.Decode(content)
```

pkg/router/routeapihelpers/validation_test.go
```diff
@@ -2316,6 +2479,92 @@ func TestExtendedValidateRoute(t *testing.T) {
...
+					},
+				},
+			},
+			expectedErrors: 0,
+		},
+		{
+			// A private key containing 2 keys, one right and one wrong will be accepted because
+			// only the first key is considered. Tested on HAProxy / Router and this also applies
+			// to how OpenSSL parses the keys
+			name: "A private key containing 2 keys is accepted",
```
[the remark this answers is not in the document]
rikatz: added

pkg/router/routeapihelpers/validation.go
rikatz: ah actually, this is inside a "for" that keeps decoding the content and reducing the slice, so at some point content will be 0, and the error will always happen if we add this validation. 

The right fix here will be `for len(content) > 0 { ....`

pkg/router/routeapihelpers/validation.go
```diff
@@ -251,6 +252,14 @@ func ExtendedValidateRoute(route *routev1.Route) field.ErrorList {
 			if len(keyBytes) == 0 {
 				result = append(result, field.Invalid(tlsFieldPath.Child("key"), "", "no key specified"))
 			} else {
+				// Validate that final key contains only private key, and cert contains only public keys
```
candita, line 255: Would it be worth it here to also call [cert.parseCertsPEM ](https://github.com/kubernetes/client-go/blob/master/util/cert/pem.go#L35) like we do for the other certs in the `tlsConfig`?

candita: Actually, never mind,  we already call it on Line 226.

pkg/router/routeapihelpers/validation.go
candita: Except that means that empty content is valid, because it would return nil.

pkg/router/routeapihelpers/validation.go
candita: I don't think the comment is describing what's happening.  The certBytes is being checked that it only contains CERTIFICATE blocks.  Doesn't [splitCertKey](https://github.com/openshift/router/blob/master/pkg/router/routeapihelpers/validation.go#L146-L169) do what you want to do here?

rikatz: not really, because on the "else" condition the keyBytes can be empty, and then you append the tlsConfig.Key `keyBytes = append(keyBytes, []byte(tlsConfig.Key)...)` which you didn't really validated yet.

The certBytes indeed after the split may have just CERTIFICATE blocks, but I still wanted to guarantee that we were not leaking any "bug" from the split into the PEM content, that's why I have executed it again

rikatz: (I mean, this is more like an additional defensive check)

pkg/router/routeapihelpers/validation.go
```diff
@@ -281,6 +290,27 @@ func ExtendedValidateRoute(route *routev1.Route) field.ErrorList {
 	return result
 }
 
+// validatePEMContent validates that the passed content, when PEM decoded contains
+// the expected type. This validation is necessary to block attempts of passing certificates
+// inside the field that should be private key only, and vice versa. This is required
+// because otherwise passing an out of order certificate to HAProxy will break it.
+// The function will take the array of bytes of a key/cert, and the expected type
+// that the PEM should content. The match here is not strict, but will check if the
+// passed byte type CONTAINS the expected pemType
```
candita, line 299: I'm being picky about this, but the comment here isn't clearly describing what's going on.  Let me know if you think this works.  If I've misunderstood, we may need to change the code.

```suggestion
// validatePEMContent takes content and pemType, PEM decodes content and 
// validates that the first block matches the expected pemType. This validation
// 1. Ensures that the required pemType is first in the order
// 2. Blocks attempts of passing certificates where keys should be used, and
// 3. Blocks attempts of passing keys where certificates should be used.
// Passing an out of order certificate, or key as certificate, or certificate as key
// breaks HAProxy.  Note that the match of content and pemType is not exact, but
// content must CONTAIN the expected pemType.
```

pkg/router/routeapihelpers/validation.go
candita: I still think you need to check if keyBytes and/or certBytes is nil before calling validatePEMContent.  If either is nil, there is no error result appended that would report that.

pkg/router/routeapihelpers/validation_test.go
```diff
@@ -2316,6 +2479,92 @@ func TestExtendedValidateRoute(t *testing.T) {
...
+				Spec: routev1.RouteSpec{
+					TLS: &routev1.TLSConfig{
+						Termination: routev1.TLSTerminationEdge,
+						Certificate: leafCertExtractedFromP12,
+						Key:         testCertificateRsaSha256Key + leafKeyExtractedFromP12,
+					},
+				},
+			},
+			expectedErrors: 1,
+		},
```
candita, line 2567: Can you also add a test for missing CERTIFICATE and KEY?

pkg/router/routeapihelpers/validation.go
rikatz: added a check for len < 0 on validatePEMContent

lihongan:
/verified by @rhamini3 

see https://github.com/openshift/router/pull/669#issuecomment-3286364673

lihongan:
/retest-required

```
could not run steps: step e2e-aws-serial failed: failed to acquire lease for "aws-2-quota-slice": status 502 Bad Gateway, status code 502 
```

PR #656, opened 2025-05-28
OCPBUGS-55506: Prevent startup failures due to name resolution

rohara:
When servers are listed using FQDN names rather than IP addresses, HAProxy will attempt to resolve these names at startup. In the event that a name cannot be resolve, HAProxy will fail. To overcome this, use the 'init-addr' such that even if name resolution fails, HAProxy will still start normally. Any server that fails name resolution will be set in the DOWN state. Note that the key here is the 'none' keyword used with 'init-addr'

Also note that this option is being set via the 'default-server' option in the 'defaults' section and therefore applied to all servers listed in the configuration.

PR #672, opened 2025-09-09
OCPBUGS-61224: Revert Prevent startup failures due to name resolution"

sdodson:
Reverts openshift/router#656

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -184,22 +184,6 @@ defaults
   option contstats
   {{- end }}
 
-  # Prevent failures at startup due to name resolution
-  default-server init-addr last,none resolvers ingress_dns
-
-# Define a resolver to be used when servers are specified by name and
-# therefore requires name resolution. Server name resolution typically
-# occurs only at startup. With this configuration, name resolution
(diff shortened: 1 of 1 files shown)


sdodson:
/payload-job periodic-ci-openshift-release-master-ci-4.21-upgrade-from-stable-4.20-e2e-azure-ovn-upgrade 10

dgoodwin:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.21-upgrade-from-stable-4.20-e2e-azure-ovn-upgrade 10

This is the command for multiple runs I believe.

With this kind of thing we need to be super precise examining the results, we're looking for a very specific pattern of disruption, new connections going down, during node update, to all three ingress related backends, image-registry, oauth, and console, usually about 14-18 seconds total for each. 

I do not see the pattern in these runs above so far, but I think we're facing a 3-4% fail rate so it's going to be tough to be sure we've fixed it.

sdodson:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.21-upgrade-from-stable-4.20-e2e-azure-ovn-upgrade 10

I think also if we want to be 90% sure that we've figured out that this is fixed we need something like > 40 runs without the error assuming it manifested approximately 5% of the time prior. So queue another 10, don't want to go too wide all at once.

PR #677, opened 2025-09-22
Reapply "OCPBUGS-55506: Prevent startup failures due to name resolution"

rfredette:
This reverts commit ac22e2224c8d0980310050766a1576d34d3a658a.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -184,6 +184,22 @@ defaults
   option contstats
   {{- end }}
 
+  # Prevent failures at startup due to name resolution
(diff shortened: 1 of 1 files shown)


bentito:
@coderabbitai are the current check fails due to infra flakes or code changes in this PR?

Miciah:
We have not been able to reproduce the panic that the change was causing the first time we merged it.  Because the panic is so rare, we likely need a very large number of CI runs to observe the issue.  The plan therefore is to merge the PR again and monitor CI for coredumps, and to do the merge **ASAP after branch cut** (which was more than 3 weeks ago) so we can fix this early in the development cycle and minimize disruption to other teams.  

/approve

PR #685, opened 2025-10-28
OCPBUGS-49908: dynamicallyReplaceEndpoints: Check for idled svc

Miciah:
When dynamically replacing a route's endpoints, force a reload if the new endpoints include an endpoint for an idled service.  

Before this change, the Dynamic Configuration Manager could replace an active endpoint that had the health check enabled with the endpoint for an idled service, leaving the health check enabled for the idled service.  The health check could then cause the idled service to be unidled.  This commit inhibits dynamic updates for a route that has an idled endpoint; this means that the router follows the conventional update logic, which does not enable health checks for idled services.  

---

Once this PR merges, we need to post a revert of https://github.com/openshift/api/pull/2552 to move DCM back to TPNU.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -872,6 +872,38 @@ func (r *templateRouter) dynamicallyReplaceEndpoints(id ServiceUnitKey, service 
 
 		newEndpoints := endpointsForAlias(cfg, service)
 
+		// If a service is idled, createRouterEndpoints returns a slice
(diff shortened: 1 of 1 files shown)


rikatz (approved):
looks good to me, thanks for the detailed explanation.

Will leave the lgtm for @grzpiotrowski

grzpiotrowski (approved):
/lgtm

PR #730, opened 2026-02-10
OCPBUGS-74511: remove RouteExternalCertificate feature gate

jcmoraisjr:
RouteExternalCertificate featuregate now is enabled by default. We used to have this option defaulting to False, cluster-ingress-operator is responsible to always configure the correct value, which is now moving to a hardcoded True. Router on the other side continues having it parameterized.

We are changing its behavior as True. This better aligns with the value that is provided by cluster-ingress-operator, in case the router isrunning standalone / development environment, and the option is not configured.

https://issues.redhat.com/browse/OCPBUGS-74511


melvinjoseph86:
Reapplying verified label https://github.com/openshift/router/pull/730#issuecomment-3879084929
/verified by @mjoseph

alebedev87 (commented):
/lgtm
/approve

PR #733, opened 2026-02-13
NE-2664: parameterize haproxy version

jcmoraisjr:
add two new envvars used to parameterize how to start or reload haproxy.

* `ROUTER_HAPROXY_VERSION`: defines the haproxy version to use. I must be one of the installed versions. If missing, it defaults to not use any suffix, preserving the same behavior before the change.

* `ROUTER_HAPROXY_MASTER_UNIX_SOCKET`: defines the full qualified path of master's unix socket. If provided, a reload command is issued to master instead of using the reload script, so configuring this option makes ROUTER_HAPROXY_VERSION to be ignored. If missing, the previous behavior is used, which is to call the reload script.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -19,7 +19,7 @@ $ make
 * An admin-scoped `KUBECONFIG` for the cluster.
 * Install [imagebuilder](https://github.com/openshift/imagebuilder)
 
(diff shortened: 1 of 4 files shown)


jcmoraisjr:
Closing this branch on behalf of https://github.com/openshift/router/pull/772

/close

PR #736, opened 2026-02-23
OCPBUGS-49738: fix handling of host conflict

jcmoraisjr:
Routes are grouped by host on three distinct hashmaps when allowWildcardRoutes is enabled. These hashmaps are updated whenever routes are added, modified or deleted, and their content is used to evaluate host conflicts. This hashmaps handling is done using the current host of the route resource, so, if the host value changes, the router does not find the entry on the hashmaps because it is using the current - and still not used - host value as the key.

The approach used on this update is to ignore the host from the router, and remove the resource from all the hosts instead, since the only match on all the hashmaps should be found on the key corresponding to the former host value.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale host and wildcard conflicts when routes are updated, removed, or reassigned by ensuring prior host ownership is cleared before reclaiming.

* **Refactor**
  * Centralized and simplified route/host cleanup for more consistent claiming and release behavior.

* **Tests**
  * Added tests covering wildcard host lifecycle (add → modify → delete) to ensure hosts can be reclaimed without errors.

### DIFF
--- a/pkg/router/controller/host_admitter.go
+++ b/pkg/router/controller/host_admitter.go
@@ -2,8 +2,10 @@ package controller
 
 import (
 	"fmt"
+	"slices"
 
 	kapi "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/watch"
 
@@ -19,30 +21,17 @@ type RouteAdmissionFunc func(*routev1.Route) error
 type RouteMap map[string][]*routev1.Route
 
 // RemoveRoute removes any existing routes that match the given route's namespace and name for a key
-func (srm RouteMap) RemoveRoute(key string, route *routev1.Route) bool {
-	k := 0
-	removed := false
+func (srm RouteMap) RemoveRoute(key string, route *routev1.Route) {
+	source := srm[key]
+	remaining := slices.DeleteFunc(source, func(r *routev1.Route) bool {
+		return r.Namespace == route.Namespace && r.Name == route.Name
+	})
 
-	m := srm[key]
-	for i, v := range m {
-		if m[i].Namespace == route.Namespace && m[i].Name == route.Name {
-			removed = true
-		} else {
-			m[k] = v
-			k++
-		}
-	}
-
-	// set the slice length to the final size.
-	m = m[:k]
-
-	if len(m) > 0 {
-		srm[key] = m
+	if len(remaining) > 0 {
+		srm[key] = remaining
 	} else {
 		delete(srm, key)
 	}
-
-	return removed
 }
 
 func (srm RouteMap) InsertRoute(key string, route *routev1.Route) {
@@ -94,6 +83,10 @@ type HostAdmitter struct {
 	claimedHosts     RouteMap
 	claimedWildcards RouteMap
 	blockedWildcards RouteMap
+
+	// routeHosts caches the previous host of a route, used to lookup
+	// old entries in the claimed hosts and wildcards hashmaps.
+	routeHosts map[types.UID]string
 }
 
 // NewHostAdmitter creates a plugin wrapper that checks whether or not to
@@ -111,6 +104,7 @@ func NewHostAdmitter(plugin router.Plugin, fn RouteAdmissionFunc, allowWildcards
 		claimedHosts:     RouteMap{},
 		claimedWildcards: RouteMap{},
 		blockedWildcards: RouteMap{},
+		routeHosts:       make(map[types.UID]string),
 	}
 }
 
@@ -149,10 +143,16 @@ func (p *HostAdmitter) HandleRoute(eventType watch.EventType, route *routev1.Rou
 			}
 
 		case watch.Deleted:
-			p.claimedHosts.RemoveRoute(route.Spec.Host, route)
-			wildcardKey := routeapihelpers.GetDomainForHost(route.Spec.Host)
-			p.claimedWildcards.RemoveRoute(wildcardKey, route)
-			p.blockedWildcards.RemoveRoute(wildcardKey, route)
+			// Remove claimed and blocked entries based on the old host and old wildcard key:
+			// the current one (route.Spec.Host) could have been rejected due to a host conflict,
+			// in that case the claimed and blocked maps are not updated.
+			if oldHost, found := p.routeHosts[route.UID]; found {
+				oldWildcardKey := routeapihelpers.GetDomainForHost(oldHost)
+				p.claimedHosts.RemoveRoute(oldHost, route)
+				p.blockedWildcards.RemoveRoute(oldWildcardKey, route)
+				p.claimedWildcards.RemoveRoute(oldWildcardKey, route)
+				delete(p.routeHosts, route.UID)
+			}
 		}
 	}
 
@@ -214,28 +214,31 @@ func (p *HostAdmitter) addRoute(route *routev1.Route) error {
 	// Add the new route
 	wildcardKey := routeapihelpers.GetDomainForHost(route.Spec.Host)
 
+	// Restore the old host and old wildcard key, removing from the previously claimed ones
+	if oldHost, found := p.routeHosts[route.UID]; found {
+		oldWildcardKey := routeapihelpers.GetDomainForHost(oldHost)
+		// ensure the route doesn't exist as a claimed host or blocked wildcard - in case it previously was
+		p.claimedHosts.RemoveRoute(oldHost, route)
+		p.blockedWildcards.RemoveRoute(oldWildcardKey, route)
+		p.claimedWildcards.RemoveRoute(oldWildcardKey, route)
+	}
+
 	switch route.Spec.WildcardPolicy {
 	case routev1.WildcardPolicyNone:
 		// claim the host, block wildcards that would conflict with this host
 		p.claimedHosts.InsertRoute(route.Spec.Host, route)
 		p.blockedWildcards.InsertRoute(wildcardKey, route)
-		// ensure the route doesn't exist as a claimed wildcard (in case it previously was)
-		p.claimedWildcards.RemoveRoute(wildcardKey, route)
-
 	case routev1.WildcardPolicySubdomain:
 		// claim the wildcard
 		p.claimedWildcards.InsertRoute(wildcardKey, route)
-		// ensure the route doesn't exist as a claimed host or blocked wildcard
-		p.claimedHosts.RemoveRoute(route.Spec.Host, route)
-		p.blockedWildcards.RemoveRoute(wildcardKey, route)
 	default:
-		p.claimedHosts.RemoveRoute(route.Spec.Host, route)
-		p.claimedWildcards.RemoveRoute(wildcardKey, route)
-		p.blockedWildcards.RemoveRoute(wildcardKey, route)
 		err := fmt.Errorf("unsupported wildcard policy %s", route.Spec.WildcardPolicy)
 		p.recorder.RecordRouteRejection(route, "RouteNotAdmitted", err.Error())
 		return err
 	}
+
+	// adding/updating host succeeded, update the current host for this route's UID
+	p.routeHosts[route.UID] = route.Spec.Host
 
 	return nil
 }
--- a/pkg/router/controller/host_admitter_test.go
+++ b/pkg/router/controller/host_admitter_test.go
@@ -16,6 +16,7 @@ import (
 
 	routev1 "github.com/openshift/api/route/v1"
 	"github.com/openshift/client-go/route/clientset/versioned/fake"
+	"github.com/stretchr/testify/require"
 )
 
 const (
@@ -155,10 +156,11 @@ func TestHostAdmit(t *testing.T) {
 
 func TestWildcardHostDeny(t *testing.T) {
 	p := &fakePlugin{}
-	admitter := NewHostAdmitter(p, wildcardRejecter, false, false, LogRejections)
+	admitter := NewHostAdmitter(p, wildcardAdmitter, true, false, LogRejections)
 	tests := []struct {
 		name   string
 		host   string
+		path   string
 		policy routev1.WildcardPolicyType
 		errors bool
 	}{
@@ -174,6 +176,7 @@ func TestWildcardHostDeny(t *testing.T) {
 		{
 			name:   "allowed2",
 			host:   "www.host.admission.test",
+			path:   "/api",
 			policy: routev1.WildcardPolicyNone,
 			errors: false,
 		},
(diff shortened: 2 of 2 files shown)


gcs278:
@coderabbit can you review this PR?

pkg/router/controller/host_admitter.go
gcs278, line 1: Could we add a unit test in `pkg/router/controller/host_admitter_test.go`, maybe something like `TestHostChangeRemovesStaleEntry`?

jcmoraisjr: Sure, just added a new test, and used this opportunity to revisit the problem, the approach, and how a test could explore it. This new test fails without the fix and succeed after it.

pkg/router/controller/host_admitter_test.go
```diff
@@ -258,6 +259,76 @@ func TestWildcardHostDeny(t *testing.T) {
...
+			require.NoError(t, err)
+
+			route1.Spec.Host = tc.newHostRoute1
+			route2.Spec.Host = tc.newHostRoute2
+			err = admitter.HandleRoute(watch.Modified, route1)
+			require.NoError(t, err)
+			err = admitter.HandleRoute(watch.Modified, route2)
+			require.NoError(t, err)
+		})
+	}
```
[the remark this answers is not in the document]
jcmoraisjr: Good improvements, thanks. I'm skipping just the route3 scenario because it adds premises to be followed on how hostnames are named. The way the test is designed now makes the test in one of the routes, using the second one just as a way to reach the failure scenario.

pkg/router/controller/host_admitter_test.go
```diff
@@ -258,6 +262,79 @@ func TestWildcardHostDeny(t *testing.T) {
...
+			err = admitter.HandleRoute(watch.Added, route2)
+			require.NoError(t, err)
+
+			route1.Spec.Host = tc.newHostRoute1
+			route2.Spec.Host = tc.newHostRoute2
+			err = admitter.HandleRoute(watch.Modified, route1)
+			require.NoError(t, err)
+			err = admitter.HandleRoute(watch.Modified, route2)
+			require.NoError(t, err)
+		})
```
gcs278, line 373: *nit* For complete coverage, consider adding a delete then recreate here with another route that is using the same hostname, since the bug https://redhat.atlassian.net/browse/OCPBUGS-49738 called out a "phantom" hostname claim even while the route was deleted.


```suggestion
      route3name := tc.namePrefix + "-route3"
      route3 := &routev1.Route{                                                                                                                                                                   
          ObjectMeta: metav1.ObjectMeta{                                                                                                                                                          
              Name:      route3name,                                                                                                                                                              
              Namespace: "host-update",                                                                                                                                                           
              UID:       types.UID(route3name),                                                                                                                                                   
          }, 
          Spec: routev1.RouteSpec{Host: tc.oldHostRoute1, WildcardPolicy: tc.policy},                                                                                                             
      }
      [...]
      // Delete both routes                                                                                                                                                                       
      err = admitter.HandleRoute(watch.Deleted, route1)                                                                                                                                           
      require.NoError(t, err)                                                                                                                                                                     
      err = admitter.HandleRoute(watch.Deleted, route2)                                                                                                                                           
      require.NoError(t, err)                                                                                                                                                                     
                                                                                                                                                                                                  
      // New route should be able to claim route1's original host                                                                                                                                                                                                                                                                                                                  
      err = admitter.HandleRoute(watch.Added, route3)
      require.NoError(t, err) 
		})
```

pkg/router/controller/host_admitter.go
```diff
@@ -207,6 +202,15 @@ func (p *HostAdmitter) addRoute(route *routev1.Route) error {
...
 	}
 
+	// Internal hashmaps store all hostnames used to calculate displaced routes.
+	// This is misbehaving because the hostname is used as key, which leaves dangling
+	// entries behind in case of a host update. Since it is not neither possible nor
+	// safe to use the old hostname value, we're simply removing the hashmap entries
+	// from the route being updated, the rest of the code adds the entry again.
+	p.claimedHosts.RemoveRouteFromAllKeys(route)
+	p.blockedWildcards.RemoveRouteFromAllKeys(route)
+	p.claimedWildcards.RemoveRouteFromAllKeys(route)
```
gcs278: I had claude do a rough benchmarking for me.

Master (before)         PR (after)           Ratio
routes-100          ~3,800 ns/op            ~7,600 ns/op         2x slower
routes-1,000        ~8,500 ns/op           ~85,500 ns/op        10x slower
routes-10,000     ~170,000 ns/op        ~1,100,000 ns/op         6x slower

Nothing unexpected here - iterating over the routes can be expensive for large clusters.

My experience is that the router is particularly performance sensitive - and sometimes have customers with an enormous amount of routes (10k+).

The alternative to looping through is to use a data structure to do some "bookkeeping" of the previous route (key) to hostname (value), so that you could look up the route's previous hostname quickly and remove that without scanning.

Could look something like:
```suggestion
  if oldHost, ok := p.routeHosts[route.UID]; ok {                           
        oldWildcardKey := routeapihelpers.GetDomainForHost(oldHost)                                                                                                                             
        p.claimedHosts.RemoveRoute(oldHost, route)                                                                                                                                              
        p.blockedWildcards.RemoveRoute(oldWildcardKey, route)                                                                                                                                   
        p.claimedWildcards.RemoveRoute(oldWildcardKey, route)                                                                                                                                   
    }                                                                                                                                                                                           
    p.routeHosts[route.UID] = route.Spec.Host 
```

Did you consider adding a cache of route hostnames like this for better performance?

gcs278 (commented):
Nice PR - generally looks good, I just have a concern about performance and a minor unit test coverage addition.

pkg/router/controller/host_admitter_test.go
jcmoraisjr: Good idea, just added.

pkg/router/controller/host_admitter.go
jcmoraisjr: This is a fair point, in fact I was focusing more in the problem instead of in the solution. Just changed pretty much like you suggested, changing a bit in naming and location of the code.

pkg/router/controller/host_admitter.go
```diff
@@ -153,6 +147,7 @@ func (p *HostAdmitter) HandleRoute(eventType watch.EventType, route *routev1.Rou
 			wildcardKey := routeapihelpers.GetDomainForHost(route.Spec.Host)
```
gcs278: I think there is another nuanced variation of this bug with delete behavior that we should consider fixing while we are here. Consider the scenario:

1. RouteX added with hostA: maps have entries under A, `routeHosts[UID] = A`                                                                                                                 
2. RouteX modified to hostB, but rejected by `displacedRoutes`: maps still have entries under hostA, `routeHosts[UID] still = A`
3. RouteX deleted: `route.Spec.Host = B`, but maps have the route under A: `RemoveRoute("B", route)` is a no-op, stale entry leaks

In this scenario, we'd be attempting to clean up entries in `claimedHosts`, `claimedWildcards`, and `blockedWildcards` using a host that was never admitted. `routeHosts` is effectively the "last successfully admitted host" while the route might have failed to admit it's current host.

Should we also use `routeHosts` here to fix that bug? Also, if you do end up fixing this, consider a unit test, like `TestDeleteAfterRejectedHostChange` to verify the bug I'm describing & the fix.

pkg/router/controller/host_admitter.go
```diff
@@ -214,24 +209,25 @@ func (p *HostAdmitter) addRoute(route *routev1.Route) error {
...
 
+	// Restore the old host and old wildcard key, removing from the previously claimed ones
+	if oldHost, found := p.routeHosts[route.UID]; found {
+		oldWildcardKey := routeapihelpers.GetDomainForHost(oldHost)
+		// ensure the route doesn't exist as a claimed host or blocked wildcard - in case it previously was
+		p.claimedHosts.RemoveRoute(oldHost, route)
+		p.blockedWildcards.RemoveRoute(oldWildcardKey, route)
+		p.claimedWildcards.RemoveRoute(oldWildcardKey, route)
+	}
+	p.routeHosts[route.UID] = route.Spec.Host
```
gcs278: Should this go below the switch statement? The switch statement's `default` case is a rejection for an unsupported wildcard policy, which means we'd be storing a route host in our map that was rejected.

Probably more of defensive coding comment - I don't know if it's possible to trigger the `Default` case with API validation.

pkg/router/controller/host_admitter.go
jcmoraisjr: Good catch, got it, makes sense, just updated as suggested.

I reused the same tests to cover this new scenario, since this is a complement of what the test already implements, how do the changes sound?

pkg/router/controller/host_admitter.go
jcmoraisjr: In fact I think the API validation blocks this path to happen, but it worth to use a defensive code here. Just changed as suggested.

gcs278:
LGTM. Code Rabbit did [point out](https://github.com/openshift/router/pull/736#pullrequestreview-4220943275) that  there is an minor inconsistency with removing the old routes from the hashmaps lookups before the wildcard policy is validated. Practically this can't happen today since the API validates the policy before it reaches the router, so I'm okay with moving forward. Feel free to address if you want.

Thanks for the quick turn around!

/approve
/lgtm

melvinjoseph86:
Re verifying this fix

```
➜  Downloads oc get clusterversion
version   5.0.0-0-2026-05-05-073241-test-ci-ln-7tgybmb-latest   True        False         22m     Cluster version is 5.0.0-0-2026-05-05-073241-test-ci-ln-7tgybmb-latest
➜  Downloads oc patch ingresscontroller default -n openshift-ingress-operator --type=merge -p '{"spec":{"routeAdmission":{"wildcardPolicy":"WildcardsAllowed"}}}'
ingresscontroller.operator.openshift.io/default patched
➜  Downloads oc create -f webrc.yaml 
replicationcontroller/web-server-rc created
service/service-secure created
service/service-unsecure created
➜  oc get svc
NAME               TYPE           CLUSTER-IP      EXTERNAL-IP                            PORT(S)     AGE
kubernetes         ClusterIP      172.30.0.1      <none>                                 443/TCP     56m
openshift          ExternalName   <none>          kubernetes.default.svc.cluster.local   <none>      50m
service-secure     ClusterIP      172.30.233.72   <none>                                 27443/TCP   9s
service-unsecure   ClusterIP      172.30.71.21    <none>                                 27017/TCP   9s
➜  oc expose service service-unsecure --name httpd-cm1 
route.route.openshift.io/httpd-cm1 exposed
➜  oc expose service service-unsecure --name httpd-cm2
route.route.openshift.io/httpd-cm2 exposed
➜  oc get svc
NAME               TYPE           CLUSTER-IP      EXTERNAL-IP                            PORT(S)     AGE
kubernetes         ClusterIP      172.30.0.1      <none>                                 443/TCP     60m
openshift          ExternalName   <none>          kubernetes.default.svc.cluster.local   <none>      54m
service-secure     ClusterIP      172.30.233.72   <none>                                 27443/TCP   4m3s
service-unsecure   ClusterIP      172.30.71.21    <none>                                 27017/TCP   4m3s
➜  oc get routes                                       
NAME        HOST/PORT                                                           PATH   SERVICES           PORT   TERMINATION   WILDCARD
httpd-cm1   httpd-cm1-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org          service-unsecure   http                 None
httpd-cm2   httpd-cm2-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org          service-unsecure   http                 None
➜  oc patch route httpd-cm1 --type=merge -p '{"spec":{"host":"httpd-cm2-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org"}}'  
route.route.openshift.io/httpd-cm1 patched
➜  oc get routes                                                                                                                     
NAME        HOST/PORT                                                           PATH   SERVICES           PORT   TERMINATION   WILDCARD
httpd-cm1   httpd-cm2-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org          service-unsecure   http                 None
httpd-cm2   HostAlreadyClaimed                                                         service-unsecure   http                 None
➜  oc patch route httpd-cm2 --type=merge -p '{"spec":{"host":"httpd-cm1-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org"}}' 
route.route.openshift.io/httpd-cm2 patched
➜  oc get routes                                                                                                                    
NAME        HOST/PORT                                                           PATH   SERVICES           PORT   TERMINATION   WILDCARD
httpd-cm1   httpd-cm2-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org          service-unsecure   http                 None
httpd-cm2   httpd-cm1-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org          service-unsecure   http                 None
➜  oc get routes -o json | jq -r '.items[] |.metadata.name,.spec.host,.status.ingress[].conditions[].message'
httpd-cm1
httpd-cm2-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org
null
httpd-cm2
httpd-cm1-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org
null

➜  oc patch route httpd-cm2 --type=merge -p '{"spec":{"host":"httpd-cm1-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org"}}' 
route.route.openshift.io/httpd-cm2 patched
➜  oc get routes                                                                                                                    
NAME        HOST/PORT                                                           PATH   SERVICES           PORT   TERMINATION   WILDCARD
httpd-cm1   httpd-cm1-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org          service-unsecure   http                 None
httpd-cm2   HostAlreadyClaimed                                                         service-unsecure   http                 None
➜  oc delete route httpd-cm1  
route.route.openshift.io "httpd-cm1" deleted

➜  oc get routes              
NAME        HOST/PORT            PATH   SERVICES           PORT   TERMINATION   WILDCARD
httpd-cm2   HostAlreadyClaimed          service-unsecure   http                 None
➜  oc get routes
NAME        HOST/PORT            PATH   SERVICES           PORT   TERMINATION   WILDCARD
httpd-cm2   HostAlreadyClaimed          service-unsecure   http                 None
➜  oc delete route httpd-cm2
route.route.openshift.io "httpd-cm2" deleted
➜  oc get routes            
No resources found in default namespace.
➜  oc expose service service-unsecure --name httpd-cm2
route.route.openshift.io/httpd-cm2 exposed
➜  oc get routes                                      
NAME        HOST/PORT                                                           PATH   SERVICES           PORT   TERMINATION   WILDCARD
httpd-cm2   httpd-cm2-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org          service-unsecure   http                 None
➜  oc get routes -o json | jq -r '.items[] |.metadata.name,.spec.host,.status.ingress[].conditions[].message'
httpd-cm2
httpd-cm2-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org
null
➜  oc expose service service-unsecure --name httpd-cm1                                                       
route.route.openshift.io/httpd-cm1 exposed
➜  oc get routes                                                                                             
NAME        HOST/PORT                                                           PATH   SERVICES           PORT   TERMINATION   WILDCARD
httpd-cm1   httpd-cm1-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org          service-unsecure   http                 None
httpd-cm2   httpd-cm2-default.apps.ci-ln-7tgybmb-76ef8.aws-4.ci.openshift.org          service-unsecure   http                 None
```
Hence marking as verified
/verified by @mjoseph

PR #740, opened 2026-02-26
OCPBUGS-77263: create self-signed crt if failing to read default

jcmoraisjr:
Ingress operator manages and exposes a default certificate for Router. A default certificate is used if the route does not provide one, or a request is made for a non matching route.

Current approach uses a hardcoded certificate in the router image, which should lead to image scanners reporting the storage of its private key.

This implementation removes the hardcoded certificate, as well as the template code that ensures a default certificate is provided. In the Go code side, we are ensuring that the default certificate is always provided by generating one on the fly. If this last resort option fails due to permission, the whole router process fail to start.

## Summary by CodeRabbit

* **New Features**
  * Automatic self-signed certificate creation when no default is configured.
  * Live certificate reloads via filesystem watch with atomic replacement.

* **Bug Fixes**
  * Fail-fast handling during certificate initialization (no silent fallback).
  * TLS termination now uses the configured certificate directly.
  * Default certificate and key files created with stricter permissions (0600).
  * Plugin construction now returns errors instead of producing invalid instances.


pkg/router/template/plugin.go
```diff
@@ -166,7 +166,10 @@ func NewTemplatePlugin(cfg TemplatePluginConfig, lookupSvc ServiceLookup) (*Temp
 		httpRequestHeaders:            cfg.HTTPRequestHeaders,
 	}
 	router, err := newTemplateRouter(templateRouterCfg)
-	return newDefaultTemplatePlugin(router, cfg.IncludeUDP, lookupSvc), err
+	if err != nil {
+		return nil, err
+	}
+	return newDefaultTemplatePlugin(router, cfg.IncludeUDP, lookupSvc), nil
```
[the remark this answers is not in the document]
jcmoraisjr: It's skipping the call to the later in the case the former fails.

rikatz:
@coderabbitai rise and shine, validate this PR, check for consistency, be an adversarial reviewer of Joao and tell me major and critical findings

pkg/router/template/router.go
```diff
@@ -416,13 +416,22 @@ func (r *templateRouter) writeDefaultCert() error {
...
+		// There is no default cert path as well, maybe there is a dir ...
+		if len(r.defaultCertificateDir) != 0 {
+			// use certificate from the provided dir, and return (finishes the process) in case of a failure
+			if err := secretToPem(r.defaultCertificateDir, outPath); err != nil {
+				return fmt.Errorf("failed to write default cert: %w", err)
+			}
 		} else {
-			r.defaultCertificatePath = outPath
+			// generate one on the fly only in case we don't have neither a path nor a directory configured
+			if err := generateSelfSignCert(outPath); err != nil {
```
rikatz: I think there are two things here that can be disrupting but maybe they are already disrupting:
* If an error happens, previously we would use a default certificate, but now we will error. What does this imply? That our proxy will be out? Is this "acceptable"?
* From the 'user' perspective, the worst case fallback was accepting an insecure self signed certificate once (meh) and then never caring about it, while here we will always generate a new certificate

I understand this is the edge of the edge case, just some food for thought, although I agree that it is better to have a "safe" (not much, but still better) fallback than having someone being able to do MITM in this case because the private key is there and common

jcmoraisjr: This was previously [raised](https://github.com/openshift/router/pull/740#discussion_r2884522657) and we ended up agreeing that it should fail in case there is an attempt to configure, so better to have a failing/crashing router instead of incorrectly configuring it with a distinct (the fake one) certificate.

Anyway this is an ingress operator task, so failures here would only happen in case of a bug in the operator. Users could not (should not) cause a disruption in the router due to this change in the code.

rikatz: ok, I see, thanks!

jcmoraisjr:
@coderabbitai good catch. Note that I used a simplified approach in the max key, all the others are addressed as suggested. Can you confirm all is good now?

pkg/router/template/router.go
```diff
@@ -415,13 +415,22 @@ func (r *templateRouter) writeDefaultCert() error {
 			// Just use the provided path
 			return nil
 		}
+
+		// There is no default cert path as well. If cert dir is also missing ...
+		if len(r.defaultCertificateDir) == 0 {
+			// ... use the self signed certificate. Finish here since it won't be updated during the router lifecycle.
+			return generateSelfSignedCert(outPath)
+		}
```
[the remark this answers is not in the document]
jcmoraisjr: good catch, fixed now

rikatz:
sorry for the delay, and thanks for the fixes
/lgtm
/approve

melvinjoseph86:
Tested using clusterbot

```
➜  oc get clusterversion
version   5.0.0-0-2026-05-05-131949-test-ci-ln-qxlg8lb-latest   True        False         10m     Cluster version is 5.0.0-0-2026-05-05-131949-test-ci-ln-qxlg8lb-latest

➜  oc get po -n openshift-ingress
NAME                              READY   STATUS    RESTARTS      AGE
router-default-6c45d55b44-qt55k   1/1     Running   2 (23m ago)   34m
router-default-6c45d55b44-qz8fj   1/1     Running   1 (25m ago)   34m
➜  oc exec router-default-6c45d55b44-qt55k   -n openshift-ingress -- cat /var/lib/haproxy/conf/default_pub_keys.pem|openssl x509 -out - --text 
cat: /var/lib/haproxy/conf/default_pub_keys.pem: No such file or directory
command terminated with exit code 1
Could not find certificate from <stdin>

Making the CVO unmanged
➜  oc patch clusterversions/version --type=json --patch='[{"op":"add","path":"/spec/overrides","value":[{"kind":"Deployment","group":"apps","name":"ingress-operator","namespace":"openshift-ingress-operator","unmanaged":true}]}]'
clusterversion.config.openshift.io/version patched
➜  oc -n openshift-ingress-operator scale deployment.apps/ingress-operator --replicas=0
Warning: spec.template.spec.nodeSelector[node-role.kubernetes.io/master]: use "node-role.kubernetes.io/control-plane" instead
deployment.apps/ingress-operator scaled

Customize router deployment by removing envvar DEFAULT_CERTIFICATE_DIR and ROUTER_USE_PROXY_PROTOCOL as false
➜  oc -n openshift-ingress edit deployment/router-default
deployment.apps/router-default edited

➜  oc get po -n openshift-ingress
NAME                              READY   STATUS        RESTARTS       AGE
router-default-6c45d55b44-qt55k   0/1     Terminating   2 (116m ago)   127m
router-default-6c45d55b44-qz8fj   1/1     Terminating   1 (118m ago)   127m
router-default-f5f55998f-7gkmf    0/1     Pending       0              39s
router-default-f5f55998f-xlffv    1/1     Running       0              39s
➜  oc rsh -n openshift-ingress    router-default-f5f55998f-7gkmf 
E0505 21:20:51.529604   27668 websocket.go:296] Unknown stream id 1, discarding message
sh-5.1$ 
sh-5.1$ :|openssl s_client -connect 127.0.0.1:443 |openssl x509 -noout -dates
Connecting to 127.0.0.1
Can't use SSL_get_servername
depth=0 CN=openshift-router
verify error:num=18:self-signed certificate
verify return:1
depth=0 CN=openshift-router
verify return:1
DONE
notBefore=May  5 14:49:11 2026 GMT
notAfter=May  4 14:49:11 2036 GMT
sh-5.1$ 
sh-5.1$ 
sh-5.1$ exit
```

The notBefore date is 1h before the router started to run, and notAfter is 10 years in the future comparing the router logs and also the dummy cert file is not there and checked the auto generated certificate, hence marking as verified
/verified by @mjoseph

PR #763, opened 2026-04-10
NE-2572: use add/del haproxy api calls

jcmoraisjr:
DCM currently works by creating blueprint backends and empty backend server slots upfront. These empty slots are updated later with new endpoints, created on scale out operations, without the need to fork-and-reload haproxy. Scale in operations work in a similar way: backend servers are disabled when endpoints are not available anymore.

HAProxy has add/del apis since 2.5, which adds the benefit of not having to create empty slots. This means that we don't need to predict the ideal number of empty servers per backend, so backends can be configured with the currently active endpoints only, and it can scale to any number of new backend servers without the need to reload.

This update is being made in at least two phases:

Phase 1: update config manager (manager.go) and client's backend (backend.go) the minimum to support the new api calls. The priorities are: new approach fully functional and an easier code update to revise.

Phase 2: cleanup dead code, eventually combining types into a single, coherent and simplified struct.

New phases should happen as a revisit on other parts of the code that can be optimized with the new approach, or the code cleanup, like changing route resources and the handling of the haproxy lookup maps.

## Summary by CodeRabbit

* **New Features**
  * Direct runtime backend server management (add/update/delete) with per-server health-check control.

* **Deprecated**
  * The --max-dynamic-servers option is deprecated (flag kept for parsing only).

* **Changes**
  * Router now emits a dynamic routing cookie key when dynamic management is enabled and no longer generates template-based server-templates; per-service endpoint tracking and endpoint-replacement behavior improved.
  * New working-directory and default destination CA options for dynamic management.

* **Tests**
  * New comprehensive tests for backend dynamic operations (add/update/delete, health checks).

### DIFF
--- a/pkg/router/template/configmanager/haproxy/backend.go
+++ b/pkg/router/template/configmanager/haproxy/backend.go
@@ -4,9 +4,12 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"os"
+	"path"
 	"strconv"
 	"strings"
 
+	v1 "github.com/openshift/api/route/v1"
 	templaterouter "github.com/openshift/router/pkg/router/template"
 )
 
@@ -75,11 +78,9 @@ type serverStateInfo struct {
 // BackendServerInfo represents a server [endpoint] for a haproxy backend.
 type BackendServerInfo struct {
 	Name          string
-	FQDN          string
 	IPAddress     string
 	Port          int
 	CurrentWeight int32
-	InitialWeight int32
 }
 
 // Backend represents a specific haproxy backend.
@@ -87,7 +88,7 @@ type Backend struct {
 	name    templaterouter.ServiceAliasConfigKey
 	servers map[string]*backendServer
 
-	client *Client
+	client HAProxyClient
 }
 
 // backendServer is internally used for managing a haproxy backend server.
@@ -118,7 +119,7 @@ func buildHAProxyBackends(c *Client) ([]*Backend, error) {
 }
 
 // newBackend returns a new Backend representing a haproxy backend.
-func newBackend(name templaterouter.ServiceAliasConfigKey, c *Client) *Backend {
+func newBackend(name templaterouter.ServiceAliasConfigKey, c HAProxyClient) *Backend {
 	return &Backend{
 		name:    name,
 		servers: make(map[string]*backendServer),
@@ -152,9 +153,7 @@ func (b *Backend) Refresh() error {
 			Name:          v.Name,
 			IPAddress:     v.IPAddress,
 			Port:          v.Port,
-			FQDN:          v.FQDN,
 			CurrentWeight: v.UserVisibleWeight,
-			InitialWeight: v.InitialWeight,
 		}
 
 		b.servers[v.Name] = newBackendServer(info)
@@ -304,6 +303,148 @@ func (b *Backend) FindServer(id string) (*backendServer, error) {
 	return nil, fmt.Errorf("no server found for id: %s", id)
 }
 
+func (b *Backend) AddServer(cfg *templaterouter.ServiceAliasConfig, svc *templaterouter.ServiceUnit, ep templaterouter.Endpoint, weight int32, workingDir, defaultDestinationCA string) error {
+	if err := b.innerAddServer(cfg, svc, ep, weight, workingDir, defaultDestinationCA); err != nil {
+		if !strings.Contains(err.Error(), "Already exists a server ") {
+			return err
+		}
+		// Failed due to already existing server left behind, in maintenance mode, due to in-flight connections.
+		// Let's just give it another chance to be deleted.
+		if err := b.innerDeleteServer(ep); err != nil {
+			// No way, need to fail which will ask for a fork-and-reload. This will leave the existing connections in the old process.
+			return err
+		}
+		if err := b.innerAddServer(cfg, svc, ep, weight, workingDir, defaultDestinationCA); err != nil {
+			return err
+		}
+	}
+	if err := b.innerSetServerState(ep, true, weight); err != nil {
+		return err
+	}
+
+	// health check is disabled by default on new backend servers, its enablement is handled via cm.ReplaceRouteEndpoints(),
+	// since that method has a better view of former and current active backend servers.
+	return nil
+}
+
+func (b *Backend) UpdateServer(ep templaterouter.Endpoint, weight int32, isPassthrough bool) error {
+	// missing to properly populate the current servers when created, should be done in the next phase.
+	// After that we can update only changed attributes.
+	if err := b.innerUpdateServerAddr(ep); err != nil {
+		return err
+	}
+	return b.innerUpdateServerWeight(ep, weight, isPassthrough)
+}
+
+func (b *Backend) EnableHealthCheck(ep templaterouter.Endpoint) error {
+	return b.innerSetHealthCheck(ep, true)
+}
+
+func (b *Backend) DisableHealthCheck(ep templaterouter.Endpoint) error {
+	return b.innerSetHealthCheck(ep, false)
+}
+
+func (b *Backend) DeleteServer(ep templaterouter.Endpoint) error {
+	if err := b.innerSetServerState(ep, false, 0); err != nil {
+		return err
+	}
+	if err := b.innerDeleteServer(ep); err != nil {
+		log.Info("disabling backend server instead of deleting due to a delete failure", "server", ep.ID, "error", err.Error())
+	}
+	return nil
+}
+
+func (b *Backend) innerAddServer(cfg *templaterouter.ServiceAliasConfig, svc *templaterouter.ServiceUnit, ep templaterouter.Endpoint, weight int32, workingDir, defaultDestinationCA string) error {
+	// This should always follow the template, changes here should be reflected there, both regular and passthrough backends
+	//
+	// TODO: either read this configuration from the template, or instead make the template read from here.
+	// For the former, note that creating a new template definition should conflict with the for-loop in templateRouter.writeConfig()
+	// that assumes that all the definitions should be written to disk.
+
+	cmd := fmt.Sprintf("add server %s/%s %s:%s weight %d", b.name, ep.ID, ep.IP, ep.Port, weight)
+
+	switch cfg.TLSTermination {
+	case v1.TLSTerminationReencrypt:
+		cmd += " ssl"
+		if disableHTTP2, _ := strconv.ParseBool(os.Getenv("ROUTER_DISABLE_HTTP2")); !disableHTTP2 {
+			cmd += " alpn h2,http/1.1"
+		}
+		if cfg.VerifyServiceHostname {
+			cmd += " verifyhost " + svc.Hostname
+		}
+		if cert := cfg.Certificates[cfg.Host+"_pod"]; len(cert.Contents) > 0 {
+			cmd += " verify required ca-file " + path.Join(workingDir, "router/cacerts", cert.ID+".pem")
+		} else if len(defaultDestinationCA) > 0 {
+			cmd += " verify required ca-file " + defaultDestinationCA
+		} else {
+			cmd += " verify none"
+		}
+	case "", v1.TLSTerminationEdge:
+		if ep.AppProtocol == "h2c" || ep.AppProtocol == "kubernetes.io/h2c" {
+			cmd += " proto h2"
+		}
+	case v1.TLSTerminationPassthrough:
+		// passthrough is a TCP listener and does not use ssl or proto related config
+	}
+
+	if !ep.NoHealthCheck {
+		// health check is always configured, being enabled depending on `cfg.ActiveEndpoints > 1 `
+		inter := templaterouter.FirstMatch(`[1-9][0-9]*(us|ms|s|m|h|d)?`,
+			cfg.Annotations["router.openshift.io/haproxy.health.check.interval"],
+			os.Getenv("ROUTER_BACKEND_CHECK_INTERVAL"),
+			"5000ms")
+		cmd += " check inter " + inter
+	}
+
+	podMaxConn := cfg.Annotations["haproxy.router.openshift.io/pod-concurrent-connections"]
+	if _, err := strconv.Atoi(podMaxConn); err == nil {
+		cmd += " maxconn " + podMaxConn
+	}
+
+	return execCommand(b.client, apiAddServer, cmd)
+}
+
+func (b *Backend) innerUpdateServerAddr(ep templaterouter.Endpoint) error {
+	cmd := fmt.Sprintf("set server %s/%s addr %s port %s", b.name, ep.ID, ep.IP, ep.Port)
+	return execCommand(b.client, apiSetServerAddr, cmd)
+}
+
+func (b *Backend) innerUpdateServerWeight(ep templaterouter.Endpoint, weight int32, isPassthrough bool) error {
+	cmd := fmt.Sprintf("set server %s/%s", b.name, ep.ID)
+	if isPassthrough {
+		// https://github.com/openshift/router/blob/896390778ebe15f57f87e6ca78f11c96e64c2652/pkg/router/template/configmanager/haproxy/manager.go#L446-L454
+		cmd += " weight 100%"
+	} else {
+		cmd = fmt.Sprintf("%s weight %d", cmd, weight)
+	}
+	return execCommand(b.client, apiSetServerWeight, cmd)
+}
+
+func (b *Backend) innerSetHealthCheck(ep templaterouter.Endpoint, enable bool) error {
+	enableStr := "enable"
+	if !enable {
+		enableStr = "disable"
+	}
+	cmd := fmt.Sprintf("%s health %s/%s", enableStr, b.name, ep.ID)
+	return execCommand(b.client, apiSetHealth, cmd)
+}
+
+func (b *Backend) innerSetServerState(ep templaterouter.Endpoint, ready bool, weight int32) error {
+	state := "ready"
+	if !ready {
+		state = "maint"
+	} else if weight <= 0 {
+		state = "drain"
+	}
+	cmd := fmt.Sprintf("set server %s/%s state %s", b.name, ep.ID, state)
+	return execCommand(b.client, apiSetServerState, cmd)
+}
+
+func (b *Backend) innerDeleteServer(ep templaterouter.Endpoint) error {
+	cmd := fmt.Sprintf("del server %s/%s", b.name, ep.ID)
+	return execCommand(b.client, apiDelServer, cmd)
+}
+
 // newBackendServer returns a BackendServer representing a haproxy backend server.
 func newBackendServer(info BackendServerInfo) *backendServer {
 	return &backendServer{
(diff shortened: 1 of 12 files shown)


pkg/router/template/configmanager/haproxy/backend.go
```diff
@@ -304,6 +303,142 @@ func (b *Backend) FindServer(id string) (*backendServer, error) {
...
+		cmd += " ssl"
+		if disableHTTP2, _ := strconv.ParseBool(os.Getenv("ROUTER_DISABLE_HTTP2")); !disableHTTP2 {
+			cmd += " alpn h2,http/1.1"
+		}
+		if cfg.VerifyServiceHostname {
+			cmd += " verifyhost " + svc.Hostname
+		}
+		if len(cfg.Certificates[cfg.Host+"_pod"].Contents) > 0 {
+			cmd += " verify required ca-file " + path.Join(workingDir, "router/cacerts", string(b.name)+".pem")
+		} else if len(defaultDestinationCA) > 0 {
```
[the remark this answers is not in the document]
jcmoraisjr: Good point, and your suggestion makes sense to me. Updated.

pkg/router/template/configmanager/haproxy/manager.go
```diff
@@ -471,135 +464,86 @@ func (cm *haproxyConfigManager) ReplaceRouteEndpoints(id templaterouter.ServiceA
...
+		var newEPs []templaterouter.Endpoint
+		for _, ep := range addedEndpoints {
+			// enabling for all the new added endpoints
+			newEPs = append(newEPs, ep)
+		}
+		if len(entry.activeEndpoints) == 1 {
+			// enabling also for the former single endpoint as well
+			for _, ep := range entry.activeEndpoints {
+				newEPs = append(newEPs, ep)
 			}
```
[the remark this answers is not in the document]
jcmoraisjr: Fair point, updated.

Note that I fully removed activeEndpoints for now, it was not helping the way I've figured it could.

Note also the approach I used to populate oldEndpoints and please give this update a special attention. oldEndpoints need to be accurate, having the actual haproxy state, otherwise we'd end up having wrong add and delete operations. When checking this update, note that a backend can have more than one service, and also a service can be present in more than one backend, so we need to be sure that we are not cross updating neither of them.

pkg/router/template/configmanager/haproxy/backend.go
```diff
@@ -304,6 +303,146 @@ func (b *Backend) FindServer(id string) (*backendServer, error) {
...
+
+func (b *Backend) DeleteServer(ep templaterouter.Endpoint) error {
+	if err := b.innerSetServerState(ep, false); err != nil {
+		return err
+	}
+	if err := b.innerDeleteServer(ep); err != nil {
+		log.Info("disabling backend server %s instead of deleting due to a delete failure", "error", err.Error())
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
jcmoraisjr: The most important operation is disabling the server, which already succeeded in the lines above. Having the server disabled makes it not send neither requests nor health checks to the underlying server, just use a small amount of memory. So we don't have a reason to return an error, which would lead to a reload, this is an expensive operation for the router.

I'm however puzzled about `V(0)` in the log, shouldn't level `0` behaves the same as not having it at all?

pkg/router/template/configmanager/haproxy/manager.go
```diff
@@ -410,9 +398,12 @@ func (cm *haproxyConfigManager) RemoveRoute(id templaterouter.ServiceAliasConfig
...
-	log.V(4).Info("disabling all servers for backend", "backend", backendName)
-	if err := backend.Disable(); err != nil {
-		return err
+
+	log.V(4).Info("deleting all servers for backend", "backend", backendName)
+	for _, server := range backend.servers {
+		if err := backend.DeleteServer(templaterouter.Endpoint{ID: server.Name}); err != nil {
+			return err
+		}
 	}
```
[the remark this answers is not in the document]
jcmoraisjr: Good one. Fixed by calling instead backend.Servers() method, which in turn calls Refresh().

pkg/router/template/configmanager/haproxy/backend.go
```diff
@@ -304,6 +303,148 @@ func (b *Backend) FindServer(id string) (*backendServer, error) {
...
+func (b *Backend) DisableHealthCheck(ep templaterouter.Endpoint) error {
+	return b.innerSetHealthCheck(ep, false)
+}
+
+func (b *Backend) DeleteServer(ep templaterouter.Endpoint) error {
+	if err := b.innerSetServerState(ep, false, 0); err != nil {
+		return err
+	}
+	if err := b.innerDeleteServer(ep); err != nil {
+		log.Info("disabling backend server instead of deleting due to a delete failure", "server", ep.ID, "error", err.Error())
```
davidesalerno, line 364: We are returning `nil` even when `innerDeleteServer()` fails, so callers cannot distinguish “server was drained/put in maint” from “server was fully removed from HAProxy”. 
This makes error handling less explicit and can hide partial cleanup failures. 
If keeping the backend in maint is intentional to avoid reloads, I’d suggest either returning a typed/non-fatal error to the caller or documenting this degraded-success contract in the method comment so the behavior is explicit to maintainers.

pkg/router/template/configmanager/haproxy/backend.go
```diff
@@ -304,6 +303,148 @@ func (b *Backend) FindServer(id string) (*backendServer, error) {
 	return nil, fmt.Errorf("no server found for id: %s", id)
 }
 
+func (b *Backend) AddServer(cfg *templaterouter.ServiceAliasConfig, svc *templaterouter.ServiceUnit, ep templaterouter.Endpoint, weight int32, workingDir, defaultDestinationCA string) error {
```
davidesalerno, line 309: I think that it could be useful add a small documentation to all the functions of the backend struct that are exposed/public.

pkg/router/template/configmanager/haproxy/backend.go
```diff
@@ -396,3 +537,45 @@ func stripVersionNumber(data []byte) ([]byte, error) {
...
+
+	var valid bool
+	switch api {
+	case apiAddServer:
+		valid = response == "New server registered."
+	case apiDelServer:
+		valid = response == "Server deleted."
+	case apiSetServerAddr:
+		valid = response == "nothing changed" || strings.HasPrefix(response, "IP changed from ") || strings.HasPrefix(response, "port changed from ") || strings.HasPrefix(response, "no need to change ")
+	case apiSetHealth, apiSetServerWeight, apiSetServerState:
```
davidesalerno, line 585: Minor readability nit: this branch seems to be unreachable as written, because empty responses already return at the top of `execCommand().` 

The behavior is correct, but the switch suggests those commands can still validate response == "" here. I’d simplify this case so non-empty responses for those commands always fall through to the generic error path; that makes the success contract clearer.

pkg/router/template/configmanager/haproxy/backend.go
```diff
@@ -304,6 +303,148 @@ func (b *Backend) FindServer(id string) (*backendServer, error) {
...
+	if err := b.innerDeleteServer(ep); err != nil {
+		log.Info("disabling backend server instead of deleting due to a delete failure", "server", ep.ID, "error", err.Error())
+	}
+	return nil
+}
+
+func (b *Backend) innerAddServer(cfg *templaterouter.ServiceAliasConfig, svc *templaterouter.ServiceUnit, ep templaterouter.Endpoint, weight int32, workingDir, defaultDestinationCA string) error {
+	// This should always follow the template, changes here should be reflected there, both regular and passthrough backends
+	//
+	// TODO: either read this configuration from the template, or instead make the template read from here.
```
davidesalerno, line 373: I think the `TODO` here is worth treating as real follow-up feedback: the server definition is now duplicated between the template and `innerAddServer()`, which is fragile for a code path this sensitive. Any future change to reencrypt options, proto h2, health checks, cookies, or maxconn has to be updated in two places and kept byte-for-byte aligned. 
I’d strongly prefer a single source of truth soon after this lands—either generate the runtime command from shared logic, or have the template consume the same data/renderer so static and dynamic backends cannot drift.

davidesalerno (commented):
The changes look almost good even if I have some feedbacks to share that could help us to improve the final work.

pkg/router/template/configmanager/haproxy/backend.go
jcmoraisjr: Sure makes sense. Just added.

pkg/router/template/configmanager/haproxy/backend.go
jcmoraisjr: This is an expected scenario, but your concern and comments make sense. Just changed the method signature to return whether a removal fully succeeded or not, added doc, improved test, but ignored from the caller since we don't need this state right now.

pkg/router/template/configmanager/haproxy/backend.go
jcmoraisjr: Sure, just added https://redhat.atlassian.net/browse/NE-2646 to handle this.

pkg/router/template/configmanager/haproxy/backend.go
jcmoraisjr: In fact we need those entries just here, otherwise it would panic as invalid IDs. So I changed it to a hard coded false instead and added a comment, wondering if it sounds better now.

pkg/router/template/configmanager/haproxy/backend.go
```diff
@@ -304,6 +303,163 @@ func (b *Backend) FindServer(id string) (*backendServer, error) {
...
+// It returns a failure in case HAProxy refuses to dynamically add the server for any reason, or if the existing server
+// cannot be removed, e.g., it still have active or steady and established connection(s) to its backend server endpoint.
+func (b *Backend) AddServer(cfg *templaterouter.ServiceAliasConfig, svc *templaterouter.ServiceUnit, ep templaterouter.Endpoint, weight int32, workingDir, defaultDestinationCA string) error {
+	if err := b.innerAddServer(cfg, svc, ep, weight, workingDir, defaultDestinationCA); err != nil {
+		if !strings.Contains(err.Error(), "Already exists a server ") {
+			return err
+		}
+		// Failed due to already existing server left behind, in maintenance mode, due to in-flight connections.
+		// Let's just give it another chance to be deleted.
+		if err := b.innerDeleteServer(ep); err != nil {
```
bentito, line 316: Should we first put the server into `maint` mode (like `DeleteServer` does)?

pkg/router/template/configmanager/haproxy/manager.go
```diff
@@ -471,135 +466,74 @@ func (cm *haproxyConfigManager) ReplaceRouteEndpoints(id templaterouter.ServiceA
...
-		if ep, ok := modifiedEndpoints[relatedEndpointID]; ok {
-			configChanged = true
-			log.V(4).Info("enabling server for modified endpoint", "endpoint", relatedEndpointID, "server", s.Name, "ip", ep.IP, "port", ep.Port, "appProtocol", ep.AppProtocol, "weight", weight)
-			backend.UpdateServerInfo(s.Name, ep.IP, ep.Port, ep.AppProtocol, weight, weightIsRelative)
-			backend.EnableServer(s.Name)
-
-			delete(modifiedEndpoints, relatedEndpointID)
+		if len(oldEndpoints) == 1 {
+			// enabling also for the former single endpoint as well
+			newEPs = append(newEPs, oldEndpoints[0])
```
bentito, line 518: I think this ignores `modifiedEndpoints`. If an existing endpoint was modified in a way that requires its health check state to change, it will be missed, no?

bentito (commented):
~Adversarial review findings~

pkg/router/template/configmanager/haproxy/backend.go
```diff
@@ -304,6 +303,163 @@ func (b *Backend) FindServer(id string) (*backendServer, error) {
...
+		}
+	case "", v1.TLSTerminationEdge:
+		if ep.AppProtocol == "h2c" || ep.AppProtocol == "kubernetes.io/h2c" {
+			cmd += " proto h2"
+		}
+	case v1.TLSTerminationPassthrough:
+		// passthrough is a TCP listener and does not use ssl or proto related config
+	}
+
+	if !ep.NoHealthCheck {
```
bentito: If we remove this if `!ep.NoHealthCheck`, the health check parameters are always initialized. As-is, I think HAProxy rejects future dynamic `enable health` commands for this endpoint because the check was never configured.

pkg/router/template/configmanager/haproxy/backend.go
jcmoraisjr: This is not the case here, the only scenario that makes sense is an already maint server that failed to be deleted before due to established connections, now we're trying to remove it again, and it should fail if there are still active connections and the only option is to reload - by returning an error.

pkg/router/template/configmanager/haproxy/manager.go
jcmoraisjr: This condition is only to catch the former "single endpoint" which by design doesn't have its health check enabled. Modifications don't disable health checks, and additions are already covered elsewhere.

pkg/router/template/configmanager/haproxy/backend.go
jcmoraisjr: Enable health shouldn't happen since `NoHealthCheck` denies it, so we should be good. Anyway you're right, in case `NoHealthCheck` changes we'd need a reload just to configure check and inter, and a dynamically added endpoint already starts with health check disabled. Lemme better explore this, I think we can in fact drop this condition and always add check and inter. Thanks for the heads up.

jcmoraisjr: Addressed this one. Note that I also removed `NoHealthCheck` from the backend tests since it is not used in this part of the client.

pkg/router/template/configmanager/haproxy/backend.go
```diff
@@ -304,6 +303,162 @@ func (b *Backend) FindServer(id string) (*backendServer, error) {
...
+		// Failed due to already existing server left behind, in maintenance mode, due to in-flight connections.
+		// Let's just give it another chance to be deleted.
+		if err := b.innerDeleteServer(ep); err != nil {
+			// No way, need to fail which will ask for a fork-and-reload. This will leave the existing connections in the old process.
+			return err
+		}
+		if err := b.innerAddServer(cfg, svc, ep, weight, workingDir, defaultDestinationCA); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
jcmoraisjr: Good point. This is however an unfortunate consequence of how HAProxy API is designed, and our whole client code is designed to handle HAProxy particularities. This is not only related with the responses you pointed out, but also behaviors depending on current state and what changes. We are already dealing with those particularities in two fronts: ensure correct implementation via unit tests as you already found, and ensure correct behavior via e2e tests that cover changes on messages and behaviors.

pkg/router/template/configmanager/haproxy/backend.go
bentito: ah okay

pkg/router/template/configmanager/haproxy/manager.go
bentito: I understand it now, thanks, seems fine

melvinjoseph86:
cc @rhamini3

rhamini3:
Running pre-merge tests on this feature

Check dynamic cookie, scale deployment to check API calls, confirm health check enablement, and no pid changes on all route types
```
cat haproxy.config | grep unsec-route -A20
backend be_http:default:unsec-route
  mode http
  option redispatch
  option forwardfor
  balance random

  timeout check 5000ms
  # See the quoting rules in https://tools.ietf.org/html/rfc7239 for IPv6 addresses (v4 addresses get translated to v6 when in hybrid mode)
  acl ipv6_addr src -m sub :
  http-request add-header Forwarded for=\"[%[src]]\";host=%[req.hdr(host)];proto=%[req.hdr(X-Forwarded-Proto)] if ipv6_addr
  http-request add-header Forwarded for=%[src];host=%[req.hdr(host)];proto=%[req.hdr(X-Forwarded-Proto)] if !ipv6_addr
  cookie 2b6829b92844a407b15271fc05f99815 insert indirect nocache httponly dynamic 
  dynamic-cookie-key 2b6829b92844a407b15271fc05f99815

------
be_http:default:unsec-route
pod:web-server-deploy-85b57889f5-wzd6v:service-unsecure:http:10.129.2.17:8080
MAINT
10.129.2.17:8080
------
be_http:default:unsec-route
pod:web-server-deploy-85b57889f5-kdvdl:service-unsecure:http:10.131.0.22:8080
UP
L4OK
10.131.0.22:8080
------
be_http:default:unsec-route
pod:web-server-deploy-85b57889f5-brjvf:service-unsecure:http:10.128.2.10:8080
UP <-- health check enablement
L4OK 
10.128.2.10:8080

% oc -n openshift-ingress exec router-default-546b854ff7-4hj8d -c router -- bash -c 'echo show info | socat - /var/lib/haproxy/run/haproxy.sock |grep Pid'
Pid: 296
iamin@iamin-mac openshift-tests-private % oc scale deployment web-server-deploy --replicas=5                                                                                                      
deployment.apps/web-server-deploy scaled
iamin@iamin-mac openshift-tests-private % oc scale deployment web-server-deploy --replicas=4
deployment.apps/web-server-deploy scaled
iamin@iamin-mac openshift-tests-private % oc scale deployment web-server-deploy --replicas=2
deployment.apps/web-server-deploy scaled
iamin@iamin-mac openshift-tests-private % oc -n openshift-ingress exec router-default-546b854ff7-4hj8d -c router -- bash -c 'echo show info | socat - /var/lib/haproxy/run/haproxy.sock |grep Pid'
Pid: 296
```

reencrypt route
```
------
be_secure:default:reen-route
pod:web-server-deploy-85b57889f5-xd2hk:service-secure:https:10.128.2.20:8443
UP
L6OK
10.128.2.20:8443
------
be_secure:default:reen-route
pod:web-server-deploy-85b57889f5-x484t:service-secure:https:10.129.2.12:8443
UP
L6OK
10.129.2.12:8443
------
be_secure:default:reen-route
pod:web-server-deploy-85b57889f5-4ccn7:service-secure:https:10.131.0.24:8443
UP 
L6OK
10.131.0.24:8443

 % oc -n openshift-ingress exec router-default-546b854ff7-4hj8d -c router -- bash -c 'echo show info | socat - /var/lib/haproxy/run/haproxy.sock |grep Pid'
Pid: 3326
iamin@iamin-mac openshift-tests-private % oc scale deployment web-server-deploy --replicas=5                                                                                                      
deployment.apps/web-server-deploy scaled
iamin@iamin-mac openshift-tests-private % oc scale deployment web-server-deploy --replicas=3
deployment.apps/web-server-deploy scaled
iamin@iamin-mac openshift-tests-private % oc -n openshift-ingress exec router-default-546b854ff7-4hj8d -c router -- bash -c 'echo show info | socat - /var/lib/haproxy/run/haproxy.sock |grep Pid'
Pid: 3326

backend be_secure:default:reen-route
  mode http
  option redispatch
  option forwardfor
  balance random

  timeout check 5000ms
  # See the quoting rules in https://tools.ietf.org/html/rfc7239 for IPv6 addresses (v4 addresses get translated to v6 when in hybrid mode)
  acl ipv6_addr src -m sub :
  http-request add-header Forwarded for=\"[%[src]]\";host=%[req.hdr(host)];proto=%[req.hdr(X-Forwarded-Proto)] if ipv6_addr
  http-request add-header Forwarded for=%[src];host=%[req.hdr(host)];proto=%[req.hdr(X-Forwarded-Proto)] if !ipv6_addr
  cookie eb87d57f64e27cef92ef1344501762c5 insert indirect nocache httponly secure attr SameSite=None dynamic 
  dynamic-cookie-key eb87d57f64e27cef92ef1344501762c5
```
/verified by @rhamini3

PR #764, opened 2026-04-24
OCPBUGS-84323: log the reason why dcm failed

jcmoraisjr:
Template plugin calls DCM to try dynamically update HAProxy state. In case the dynamic update fails, an internal flag is raised to instruct the template plugin to reload the new configuration on a new HAProxy process. The reason of the failure is logged, however its level is above the default log verbosity, making it not to be added in the router logs. Since this is an exception scenario, those failures are being added on the default log verbosity in order to help debug some environment misbehavior.

https://redhat.atlassian.net/browse/OCPBUGS-84323

## Summary by CodeRabbit

* **Chores**
  * Improved logging visibility for configuration management errors during route operations. Error-related log messages now appear at higher verbosity levels to aid in troubleshooting when dynamic configuration updates fail.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -799,7 +799,7 @@ func (r *templateRouter) dynamicallyAddRoute(backendKey ServiceAliasConfigKey, r
 
 	err := r.dynamicConfigManager.AddRoute(backendKey, backend.RoutingKeyName, route)
 	if err != nil {
-		log.V(4).Info("router will reload as the ConfigManager could not dynamically add route for backend", "backendKey", backendKey, "error", err)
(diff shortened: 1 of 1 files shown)


gcs278:
@jcmoraisjr sorry for letting this one languish. The concern was that increasing this verbosity might flood the logs of customers because these logs are not naturally rate limited by the reload interval.

I think there is value in this PR, but I somewhat agree that we should probably coalesce these logs to be safe. I also see you have opened https://github.com/openshift/origin/pull/31555, so maybe this is not as critical anymore.

Let me know how you want to proceed.

jcmoraisjr:
No worries, I think we can close this one and its Jira on behalf of https://github.com/openshift/origin/pull/31555, and filling the gaps via #836.

/close

PR #772, opened 2026-05-07
NE-2664: add option to separate router and haproxy containers

jcmoraisjr:
Add an opt-in configuration that informs the router that HAProxy runs on a sidecar container, so its process should be managed via its admin socket.

Add also a script that manages signals sent from kubelet to HAProxy. The script should be the entry point of the container in order to capture terminating signals, converting to SIGUSR1, and than into SIGTERM when the timeout expires. The former asks HAProxy to finish as soon as all the current connections close, and the later asks HAProxy to close (FIN) connections and finish just after receiving FIN+ACK from both sides of all connections.

The script should be moved later to a new HAProxy image, separated from the router one.

https://redhat.atlassian.net/browse/NE-2664

## Why

[enhancements#1965](https://github.com/openshift/enhancements/pull/1965/changes) describes a new feature allowing users to choose a HAProxy version from IngressController API. Separating HAProxy on its own container is a simple and secure way to achieve this, with minimal disruption on the topology.

## Summary by CodeRabbit

* **New Features**
  * Graceful HAProxy shutdown with configurable timeout and staged signal handling
  * Supervisor entrypoint that runs HAProxy in foreground, forwards signals, and ensures correct exit semantics
  * Router reload now supports both embedded and external HAProxy control paths
  * New endpoint-aware HAProxy command runner for remote/local control

* **Improvements**
  * Application-wide cancellation context propagated for coordinated shutdowns
  * Metrics collection can determine process PID via a provided function
  * Reloads emit structured success info and stricter success validation

### DIFF
--- a/images/router/haproxy/start-haproxy
+++ b/images/router/haproxy/start-haproxy
@@ -1,0 +1,57 @@
+#!/bin/bash
+set -euo pipefail
+
+##
+## This script expects HAProxy running in foreground, do not initialize it in daemon mode!
+## Also, haproxy binary should be in the PATH.
+
+shutdownTimeoutSecs=${HAPROXY_MAX_SHUTDOWN_TIMEOUT:-30}
+if ! [[ "$shutdownTimeoutSecs" =~ ^[0-9]+$ ]]; then
+    echo "Invalid timeout: $shutdownTimeoutSecs"
+    exit 1
+fi
+
+signaled=0
+
+stopHAProxy() {
+    signaled=1
+
+    # HAProxy handles SIGUSR1 by finishing its process as soon as all the current connections,
+    # active or not, are closed by either the client or the backend server.
+    echo "Sending SIGUSR1 to HAProxy process $haproxyPID"
+    kill -s USR1 $haproxyPID
+
+    # Poll the process, retuning as soon as it is not alive anymore.
+    for i in $(seq 1 $shutdownTimeoutSecs); do
+        sleep 1
+        if ! kill -0 $haproxyPID 2>/dev/null; then
+            echo "All connections are closed"
+            return
+        fi
+    done
+
+    # HAProxy handles SIGTERM by closing all the TCP connections and waiting for the full TCP handshake
+    # (FIN / ACK / FIN-ACK) from both sides, and only after that it finishes. `kill` runs asynchronous,
+    # and `wait` maintains this script alive until haproxy finishes.
+    echo "SIGUSR1 timed out, sending SIGTERM to HAProxy process $haproxyPID"
+    kill -s TERM $haproxyPID 2>/dev/null || true
+}
+
+echo "Starting HAProxy. SIGUSR1 timeout is ${shutdownTimeoutSecs}s"
+haproxy "$@" &
+haproxyPID=$!
+
+trap stopHAProxy SIGTERM SIGUSR1 SIGINT
+trap "kill -s HUP $haproxyPID" SIGHUP
+
+exit_code=0
+while kill -0 $haproxyPID 2>/dev/null; do
+    # Start `wait` again in case it returned due to a non terminate signal, e.g. SIGHUP
+    wait $haproxyPID || exit_code=$?
+done
+
+# Received signal (usually SIGTERM) takes precedence during `wait`, overriding with a clean exit
+[ "$signaled" = 1 ] && exit 0
+
+echo "haproxy exited with status code $exit_code"
+exit $exit_code
--- a/pkg/cmd/infra/router/template.go
+++ b/pkg/cmd/infra/router/template.go
@@ -39,6 +39,7 @@ import (
 	"github.com/openshift/library-go/pkg/route/secretmanager"
 
 	"github.com/openshift/router/pkg/router"
+	"github.com/openshift/router/pkg/router/client"
 	"github.com/openshift/router/pkg/router/controller"
 	"github.com/openshift/router/pkg/router/metrics"
 	"github.com/openshift/router/pkg/router/metrics/haproxy"
@@ -563,6 +564,9 @@ func (o *TemplateRouterOptions) Run(stopCh <-chan struct{}) error {
 
 	var reloadCallbacks []func()
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
 	adminSocketURL := &url.URL{Scheme: "unix", Path: "/var/lib/haproxy/run/haproxy.sock"}
 	statsPort := o.StatsPort
 	switch {
@@ -606,11 +610,17 @@ func (o *TemplateRouterOptions) Run(stopCh <-chan struct{}) error {
 			}
 		}
 
+		adminUnixSocket := os.Getenv("ROUTER_HAPROXY_ADMIN_UNIX_SOCKET")
+		hasHAProxySidecar := len(adminUnixSocket) > 0
+
+		pidFn := haproxyPidEmbedded
+		if hasHAProxySidecar {
+			pidFn = haproxyPidSidecar(ctx, adminUnixSocket)
+		}
 		collector, err := haproxy.NewPrometheusCollector(haproxy.PrometheusOptions{
 			// Only template router customizers who alter the image should need this
-			ScrapeURI: env("ROUTER_METRICS_HAPROXY_SCRAPE_URI", adminSocketURL.String()),
-			// Only template router customizers who alter the image should need this
-			PidFile:            env("ROUTER_METRICS_HAPROXY_PID_FILE", ""),
+			ScrapeURI:          env("ROUTER_METRICS_HAPROXY_SCRAPE_URI", adminSocketURL.String()),
+			PidFn:              pidFn,
 			Timeout:            timeout,
 			ServerThreshold:    serverThreshold,
 			BaseScrapeInterval: baseScrapeInterval,
@@ -639,7 +649,9 @@ func (o *TemplateRouterOptions) Run(stopCh <-chan struct{}) error {
 		checkController := metrics.ControllerLive()
 		checkSocket := metrics.AdminSocketAvailable(adminSocketURL)
 		liveChecks := []healthz.HealthChecker{checkController}
-		if !(isTrue(env("ROUTER_BIND_PORTS_BEFORE_SYNC", ""))) {
+		if !hasHAProxySidecar && !(isTrue(env("ROUTER_BIND_PORTS_BEFORE_SYNC", ""))) {
+			// Do not configure checkSocket if HAProxy is running on a sidecar,
+			// since its container already has liveness probe.
 			liveChecks = append(liveChecks, checkSocket)
 		}
 
@@ -760,6 +772,7 @@ func (o *TemplateRouterOptions) Run(stopCh <-chan struct{}) error {
 	secretManager := secretmanager.NewManager(kc, nil)
 
 	pluginCfg := templateplugin.TemplatePluginConfig{
+		AppCtx:                        ctx,
 		WorkingDir:                    o.WorkingDir,
 		TemplatePath:                  o.TemplateFile,
 		ReloadScriptPath:              o.ReloadScript,
@@ -841,6 +854,7 @@ func (o *TemplateRouterOptions) Run(stopCh <-chan struct{}) error {
 
 	select {
 	case <-stopCh:
+		cancel()
 		// 45s is the default interval that almost all cloud load balancers require to take an unhealthy
 		// endpoint out of rotation.
 		delay := getIntervalFromEnv("ROUTER_GRACEFUL_SHUTDOWN_DELAY", 45)
@@ -983,4 +997,51 @@ func getStatsAuth(statsUsernameFile, statsPasswordFile, statsUsername, statsPass
 	}
 
 	return statsUsername, statsPassword, nil
+}
+
+func haproxyPidEmbedded() (int, error) {
+	pidFile := env("ROUTER_METRICS_HAPROXY_PID_FILE", "/var/lib/haproxy/run/haproxy.pid")
+	content, err := os.ReadFile(pidFile)
+	if err != nil {
+		return 0, fmt.Errorf("can't read haproxy pid file: %s", err)
+	}
+	value, err := strconv.Atoi(strings.TrimSpace(string(content)))
+	if err != nil {
+		return 0, fmt.Errorf("can't parse haproxy pid file: %s", err)
+	}
+	return value, nil
+}
+
+func haproxyPidSidecar(ctx context.Context, unixSocket string) func() (int, error) {
+
+	// TODO: the `show proc` parsing should be part of the new HAProxy client instead.
+	// https://redhat.atlassian.net/browse/NE-2746
+
+	return func() (int, error) {
+		// `show proc` layout on 2.8+:
+		//
+		// #<PID>          <type>          <reloads>       <uptime>        <version>
+		// 420054          master          0 [failed: 0]   0d00h07m14s     2.8.18-ae90be6
+		// # workers
+		// 420056          worker          0               0d00h07m14s     2.8.18-ae90be6
+		// # programs
+		//
+		output, err := client.RunCommand(ctx, "unix://"+unixSocket, "show proc", client.ClientOpts{})
+		if err != nil {
+			return 0, fmt.Errorf("error reading the current haproxy PID: %w", err)
+		}
+		lines := strings.Split(output, "\n")
+		if len(lines) < 4 {
+			return 0, fmt.Errorf("show proc command returned an unexpected response: %s", output)
+		}
+		workerFields := strings.Fields(lines[3])
+		if len(workerFields) == 0 || strings.HasPrefix(workerFields[0], "#") {
+			return 0, fmt.Errorf("show proc command returned no worker instance: %s", output)
+		}
+		pid, err := strconv.Atoi(workerFields[0])
+		if err != nil {
+			return 0, fmt.Errorf("show proc command returned an invalid PID number: %q: %w", workerFields[0], err)
+		}
+		return pid, nil
+	}
 }
--- a/pkg/router/client/haproxyclient.go
+++ b/pkg/router/client/haproxyclient.go
@@ -1,0 +1,90 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"regexp"
+	"strings"
+	"time"
+)
+
+// ClientOpts has options for the HAProxy client. All options have a proper
+// default value, so the zero value is also a valid configuration.
+type ClientOpts struct {
+	// Timeout of the API call. The default value is 5s if not declared.
+	Timeout time.Duration
+}
+
+var haproxyEndpointRegex = regexp.MustCompile(`^(tcp|unix)://(.*)$`)
+
+// RunCommand sends the specified `cmd` to the HAProxy endpoint, which could be either
+// the API or the admin/master one. `endpoint` has the "<schema>://<address>" syntax
+// which could be either "tcp://ip-or-name:port-number" or "unix:///path/to/unix-socket".
+func RunCommand(ctx context.Context, endpoint, cmd string, opts ClientOpts) (string, error) {
+
+	// TODO: this is the starting point of a new HAProxy client, managed by the NI&D team.
+	// Further improvements already tracked via https://redhat.atlassian.net/browse/NE-2746
+
+	if opts.Timeout == 0 {
+		opts.Timeout = 5 * time.Second
+	}
+
+	var d net.Dialer
+	endpointParts := haproxyEndpointRegex.FindStringSubmatch(endpoint)
+	if len(endpointParts) != 3 {
+		return "", fmt.Errorf("invalid endpoint: %s", endpoint)
+	}
+	conn, err := d.DialContext(ctx, endpointParts[1], endpointParts[2])
+	if err != nil {
+		return "", err
+	}
+	defer conn.Close()
+	rawConn, ok := conn.(interface{ CloseWrite() error })
+	if !ok {
+		// Just a safeguard, since validation enforces TCP and Unix only, and both implements CloseWrite().
+		return "", fmt.Errorf("connection type %T does not implement CloseWrite(), cannot half close the connection", conn)
+	}
+
+	err = conn.SetDeadline(time.Now().Add(opts.Timeout))
+	if err != nil {
+		return "", err
+	}
+	// Immediately cancel the connection in case the context is cancelled.
+	stop := context.AfterFunc(ctx, func() {
+		conn.SetDeadline(time.Now())
+	})
+	defer stop()
+
+	// Adding a line-break if missing, HAProxy expects it to run the command.
+	if !strings.HasSuffix(cmd, "\n") {
+		cmd += "\n"
+	}
+	_, err = conn.Write([]byte(cmd))
+	if err != nil {
+		return "", err
+	}
+
+	// Half close the connection, this tells HAProxy to close its write counterpart
+	// (our read) as soon as it finishes to respond.
+	err = rawConn.CloseWrite()
+	if err != nil {
+		return "", err
+	}
+
+	// Reading response into a loop and populating a buffer with the response chunks.
+	// io.EOF response means the read socket was closed by haproxy after all the data being transmitted.
+	buf := make([]byte, 1024)
+	output := strings.Builder{}
+	for {
+		count, err := conn.Read(buf)
+		output.Write(buf[:count])
+		if err == io.EOF {
+			return strings.TrimSpace(output.String()), nil
+		}
+		if err != nil {
+			return "", err
+		}
+	}
+}
--- a/pkg/router/client/haproxyclient_test.go
+++ b/pkg/router/client/haproxyclient_test.go
@@ -1,0 +1,45 @@
+package client
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/openshift/router/pkg/router/client/client_testutils"
+)
+
+func TestRunCommand(t *testing.T) {
+
+	// TODO: improve with new scenarios as part of https://redhat.atlassian.net/browse/NE-2746
+	// Non exhaustive list of examples:
+	// - half-closed connection
+	// - timeouts
+	// - read/write failures
+	// - response bigger than the buffer (1024 bytes)
+
+	ctx := context.Background()
+
+	t.Run("should fail on unexpected protocol", func(t *testing.T) {
+		_, err := RunCommand(ctx, "someproto://thecontent", "valid command", ClientOpts{})
+		require.EqualError(t, err, `invalid endpoint: someproto://thecontent`)
+	})
+
+	t.Run("should fail on invalid tcp address", func(t *testing.T) {
+		_, err := RunCommand(ctx, "tcp://missingport", "valid command", ClientOpts{})
+		require.EqualError(t, err, `dial tcp: address missingport: missing port in address`)
+	})
+
+	t.Run("should fail on unix socket not found", func(t *testing.T) {
+		_, err := RunCommand(ctx, "unix:///file/not/found", "valid command", ClientOpts{})
+		require.EqualError(t, err, `dial unix /file/not/found: connect: no such file or directory`)
+	})
+
+	t.Run("should receive expected command and return expected output", func(t *testing.T) {
+		socket, stopServer := client_testutils.CreateServerMock(t, map[string]string{"valid command": "output\n\n"})
+		defer stopServer()
+		output, err := RunCommand(ctx, "unix://"+socket, "valid command", ClientOpts{})
+		require.NoError(t, err)
+		require.Equal(t, "output", output)
+	})
+}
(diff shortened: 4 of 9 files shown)


images/router/haproxy/start-haproxy
```diff
@@ -0,0 +1,57 @@
...
+timeout="${timeoutStr%s}"
+signaled=0
+
+stopHAProxy() {
+    signaled=1
+
+    # HAProxy handles SIGUSR1 by finishing its process as soon as all the current connections,
+    # active or not, are closed by either the client or the backend server.
+    echo "Sending SIGUSR1 to HAProxy process $haproxyPID"
+    kill -s USR1 $haproxyPID
```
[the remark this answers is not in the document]
jcmoraisjr: The first one should be considered a failure, since this callback should be the only one stopping HAProxy. In case of kill failing, this means that HAProxy is already stopped so we'd be just raising the failure to logs.

I can see a race in the second one though, on a really small time window, the HAProxy process should stop between kill -0 and the kill -s term, raising a strange error to the logs. So just fixed the second one, how does it look now?

pkg/router/template/router.go
```diff
@@ -679,7 +696,34 @@ func (r *templateRouter) reloadRouter(shutdown bool) error {
...
+func (r *templateRouter) reloadRouterExternal(adminSocket string) error {
+	// TODO missing application's context
+	_ = wait.PollUntilContextCancel(context.Background(), 2*time.Second, true, func(ctx context.Context) (done bool, err error) {
+		_, errstat := os.Lstat(adminSocket)
+		if errstat != nil {
+			log.Info("waiting for haproxy socket", "message", errstat.Error())
+			return false, nil
+		}
+		return true, nil
+	})
```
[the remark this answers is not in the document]
jcmoraisjr: Well, there was a todo to implement the context, so I just did it. It's better than just adding a timeout, since not having a socket should be considered a failure and probes from the outside would will our process anyway. Better now with a proper context.

gcs278:
i'd like to review this too
/assign

pkg/router/template/router.go
```diff
@@ -15,10 +16,12 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/bcicen/go-haproxy"
```
gcs278: I noticed this library hasn't been updated in 8 years. I also see PRs open that haven't gotten any attention like this one for HAProxy 3.2 fixes: https://github.com/bcicen/go-haproxy/pull/15

What are your thoughts here? I know we are only using it to send reload commands, but I wonder if it's just safer to implement the logic in the router to avoid using a stale library.

pkg/cmd/infra/router/template.go
gcs278, line 570: this is an existing problem, but made even more confusing now because we are adding the admin socket. We might want to update in this PR because this is misleading.


```suggestion
	statsSocketURL := &url.URL{Scheme: "unix", Path: "/var/lib/haproxy/run/haproxy.sock"}
```

pkg/router/template/router.go
```diff
@@ -679,7 +699,37 @@ func (r *templateRouter) reloadRouter(shutdown bool) error {
...
+			log.Info("waiting for haproxy socket", "message", errstat.Error())
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		return fmt.Errorf("router is terminating")
+	}
+
+	client := haproxy.HAProxyClient{Addr: "unix://" + adminSocket, Timeout: 10 /*seconds*/}
```
gcs278: so i see the library times out on connection `Timeout: 10`, but I don't see a read timeout [here](https://github.com/bcicen/go-haproxy/blob/master/client.go#L40).

It feels a bit dangerous that the library doesn't have a deadline/timeout for reading. If HAProxy fails to respond, the router process hangs. Maybe an argument to implement this library ourselves?

images/router/haproxy/start-haproxy
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
gcs278, line 1: Could we also adapt `reload-haproxy` to work in sidecar mode?

The script already has a mode switch for `ROUTER_SHUTDOWN`. We could add a similar check for `ROUTER_HAPROXY_ADMIN_UNIX_SOCKET` at the top that sends `reload` via `socat` instead of spawning a new haproxy process. Existing embedded behavior stays untouched, but it'd be useful for debugging in sidecar mode.


For example in `reload-haproxy`:
```
if [ -n "${ROUTER_HAPROXY_ADMIN_UNIX_SOCKET-}" ]; then
    if [ -n "${ROUTER_SHUTDOWN-}" ]; then
        echo " - Sidecar mode: shutdown handled by sidecar container"
        exit 0
    fi
    echo " - Reloading HAProxy via admin socket"
    echo "reload" | socat -t 10 - UNIX-CONNECT:${ROUTER_HAPROXY_ADMIN_UNIX_SOCKET}
    exit $?
fi
```

images/router/haproxy/start-haproxy
```diff
@@ -0,0 +1,57 @@
...
+    echo "SIGUSR1 timed out, sending SIGTERM to HAProxy process $haproxyPID"
+    kill -s TERM $haproxyPID 2>/dev/null || true
+}
+
+echo "Starting HAProxy. SIGUSR1 timeout is ${timeout}s"
+/usr/sbin/haproxy "$@" &
+haproxyPID=$!
+
+trap stopHAProxy SIGTERM SIGUSR1 SIGINT
+trap "kill -s HUP $haproxyPID" SIGHUP
```
gcs278, line 45: Interesting - this will pass SIGHUP to haproxy, causing it to reload. I think that's fine, but it's new behavior.

pkg/router/template/router.go
gcs278: Also related: The library also doesn't take `appCtx` - that would also be nice to have, incase the router is stuck in reload when shutdown is called.

pkg/router/template/router.go
```diff
@@ -679,7 +699,37 @@ func (r *templateRouter) reloadRouter(shutdown bool) error {
...
 	}
-	log.V(0).Info("router reloaded", "output", string(out))
+	log.V(0).Info("router reloaded", "mode", "embedded", "output", string(out))
+	return nil
+}
+
+// reloadRouterExternal sends a reload command to the external haproxy.
+func (r *templateRouter) reloadRouterExternal(adminSocket string) error {
+	err := wait.PollUntilContextCancel(r.appCtx, 2*time.Second, true, func(ctx context.Context) (done bool, err error) {
+		_, errstat := os.Lstat(adminSocket)
```
gcs278: *nit* we only check if the file exists, should we check if we can connect to the socket before proceeding?

gcs278 (commented):
Very interesting changes! I tried this out on a cluster and experimented with long-lived connections across reloads: the old HAProxy workers correctly stayed alive draining connections before exiting, as expected. Reloads seemed to work fine.

Getting this in CI will be the next interesting step.

images/router/haproxy/start-haproxy
jcmoraisjr: The sidecar branch is now fully managed via the socket call in the go code. My plan instead is to remove `reload-haproxy` along with the embedded haproxy, after double checking if there's something in it that worth to be migrated to the new approach. What do you think?

images/router/haproxy/start-haproxy
jcmoraisjr: This is not exactly a new behavior, but instead letting one to issue SIGHUP to haproxy since this is a signal it recognizes. It's more like not blocking the signal instead of adding something new.

pkg/cmd/infra/router/template.go
jcmoraisjr: I see the problem, but this goes deeper inside the code, e.g.

https://github.com/openshift/router/blob/35c9eaa00a6520df664a9fa3bb7ebd3e8714edc1/pkg/router/metrics/health.go#L121-L122

I talked about a proper name with @Miciah and we ended up with admin for now, at least in the public parameters. Since admin is already a public facing term (metrics), wondering about a proper one for the master socket.

pkg/router/template/router.go
jcmoraisjr: Fully agreed. I opted to make it simple for now, but I support dropping it on behalf of a dialer instead. I'd instead add another Jira so we can do this on the whole router, which includes DCM calls, just after we merge all the already planned DCM improvements. How does it sound?

pkg/router/template/router.go
jcmoraisjr: HAProxy creates the unix socket only after it can handle calls to it. Anyway it's a simple improvement that adds some value - we can also log the version being used, and we'll need this info in order to validate minimum supported versions when HAProxy version selection evolves.

Just added the HAProxy availability check here.

pkg/router/template/router.go
jcmoraisjr: The dialer is created with the configured timeout, and this should fail the copy in case it does not happen in the expected amount of time.

In fact the client does not use a context based dialer so we cannot benefit of the application context here. We can fix this as soon as we move to our own client implementation.

images/router/haproxy/start-haproxy
jcmoraisjr: Hmm. Maybe you meant a new possibility from the user. In fact yes, and maybe it makes sense since HAProxy now has its own container.

gcs278:
Will the opposite direction work?
/testwith openshift/cluster-ingress-operator/master/e2e-aws-operator openshift/cluster-ingress-operator#1439

pkg/router/template/router.go
jcmoraisjr: This update stuck the startup, since the master socket does not close the connection on other commands than reload. Just commenting for now, it's safe, and adding it again as soon as we move from the current client.

jcmoraisjr:
/hold

The current changes are mostly to run some e2e along with https://github.com/openshift/cluster-ingress-operator/pull/1439. We are however missing either some unit tests or separate the client improvement on another PR - maybe both. The new client is needed because the admin/master socket does not close the connection, making the current client to block and never close its connection.

pkg/router/template/router.go
gcs278: Yea that sounds good - I'm for getting stuff in early for CI soak time. We can merge as is - maybe with a TODO in the code. A jira tracker to update/rewrite would be good too.

pkg/router/template/router.go
gcs278: Sounds good - would be good to capture these improvements we find a a follow up Jira so we don't forget.

images/router/haproxy/start-haproxy
gcs278: Yea I realized that shortly after I posted this. It makes sense that we remove the reload-haproxy script since it seems like we are committing to the side car architecture.

images/router/haproxy/start-haproxy
gcs278: Right - I don't think the container previously passed the SIGHUP signal to HAProxy. As long as there are some rogue SIGHUP signals coming from users clusters - this should be fine.

pkg/cmd/infra/router/template.go
```diff
@@ -982,3 +991,35 @@ func getStatsAuth(statsUsernameFile, statsPasswordFile, statsUsername, statsPass
...
+		}
+		lines := strings.Split(output, "\n")
+		if len(lines) < 4 {
+			return 0, fmt.Errorf("")
+		}
+		pidStr := strings.SplitN(lines[3], " ", 2)
+		pid, err := strconv.Atoi(pidStr[0])
+		if err != nil {
+			return 0, err
+		}
```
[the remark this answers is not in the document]
jcmoraisjr: @coderabbitai this was just a prototype, now I finished the implementation, please have a look if I missed anything

pkg/router/template/router.go
jcmoraisjr: Just reverted and added this check, since now we have a client that talks better with the master client.

pkg/router/template/router.go
jcmoraisjr: Sure just added https://redhat.atlassian.net/browse/NE-2746 so we won't forget.

pkg/router/template/router.go
jcmoraisjr: Cool thanks! The final version of the client depends on PRs still in the queue, so starting with this simplified code is a way to anticipate more integration tests.

jcmoraisjr:
@coderabbitai help me being an adversarial reviewer of this PR.

Give a special attention to haproxyclient.go that I just added, since we're planning to use it on other parts of the code on future PRs. Note that I chose the half-closed connection approach to finish the read loop, since the haproxy's master server does not close its write counterpart after writing its response. Wondering if you can spot some corner cases where this approach could not work as we are expecting.

jcmoraisjr:
@coderabbitai here are the updates from your review:

* 1: Good point, implemented with `context.AfterFunc()` instead.
* 2: Closing the write side of the connection is the idiomatic way for haproxy handling its write side in the same way. This is also how socat works, already explored on HAProxy provided script like https://github.com/haproxy/haproxy/blob/master/admin/cli/haproxy-reload .
* 3: Addressed.
* 4: Added some test, we should extend them later when updating other parts of the code to use this client.
* 5: Addressed.
* 6: Good point. Addressed by removing the check altogether, since HAProxy runs as a native sidecar and should always be up and running. Added a comment about this.
* 7: Addressed.
* 8: Not a problem. A HAProxy not running is not expected and should cause a failure.
* 9: Good point, variable renamed.

jcmoraisjr:
/unhold

Code is already in a good shape, should be good to merge.

pkg/cmd/infra/router/template.go
```diff
@@ -639,7 +649,9 @@ func (o *TemplateRouterOptions) Run(stopCh <-chan struct{}) error {
 		checkController := metrics.ControllerLive()
 		checkSocket := metrics.AdminSocketAvailable(adminSocketURL)
 		liveChecks := []healthz.HealthChecker{checkController}
-		if !(isTrue(env("ROUTER_BIND_PORTS_BEFORE_SYNC", ""))) {
+		if !hasHAProxySidecar && !(isTrue(env("ROUTER_BIND_PORTS_BEFORE_SYNC", ""))) {
```
gcs278, line 652: nice 👍

pkg/router/client/haproxyclient.go
```diff
@@ -0,0 +1,82 @@
...
+	defer conn.Close()
+	rawConn, ok := conn.(interface{ CloseWrite() error })
+	if !ok {
+		// Just a safeguard, since validation enforces TCP and Unix only, and both implements CloseWrite().
+		return "", fmt.Errorf("connection type %T does not implement CloseWrite(), cannot half close the connection", conn)
+	}
+
+	// TODO: timeout should be as long as the time to reload HAProxy
+	// https://redhat.atlassian.net/browse/NE-2746
+	timeout := 1 * time.Minute
```
gcs278: Yes - 1 minute seems pretty reasonable for a HAProxy reload - previously it was 30s + curl until responding.

However, keep in mind this function also is called for the `show proc` command too. 1 minute is quite a high timeout for just getting the process - the metrics scrape interval is only [30s](https://github.com/openshift/cluster-ingress-operator/blob/723664c40e22e0e2bd3b270fa1d24be7d40d1726/pkg/operator/controller/ingress/monitoring.go#L87) I believe. I'd consider parameterizing the timeout - show proc maybe 5-10 seconds?

pkg/router/client/haproxyclient_test.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+func TestRunCommand(t *testing.T) {
+
+	// TODO: improve with new scenarios as part of https://redhat.atlassian.net/browse/NE-2746
+	// Non exhaustive list of examples:
+	// - half-closed connection
+	// - timeouts
+	// - read/write failures
+	// - response bigger than the buffer (1024 bytes)
```
gcs278, line 19: These tests are pretty important to have before 5.0 since the sidecar will be core router functionality, but a follow-up PR is fine.

pkg/router/client/haproxyclient_test.go
```diff
@@ -0,0 +1,84 @@
...
+	stop := context.AfterFunc(ctx, func() {
+		listener.Close()
+	})
+	defer stop()
+
+	go func() {
+		buf := make([]byte, 4096)
+		for {
+			conn, err := listener.Accept()
+			require.NoError(t, err)
```
davidesalerno: Using `require` into a gorutine can panic or hang the test when the goroutine outlives the test. 
It will be more stable replacing with `assert` or channel-based error reporting.

pkg/router/template/router.go
```diff
@@ -679,7 +698,25 @@ func (r *templateRouter) reloadRouter(shutdown bool) error {
 	if err != nil {
 		return fmt.Errorf("error reloading router: %v\n%s", err, string(out))
 	}
-	log.V(0).Info("router reloaded", "output", string(out))
+	log.V(0).Info("router reloaded", "mode", "embedded", "output", string(out))
+	return nil
+}
+
+// reloadRouterSidecar sends a reload command to the haproxy running as sidecar.
+func reloadRouterSidecar(ctx context.Context, adminUnixSocket string) error {
```
davidesalerno, line 706: Since there is some branching logic inside this method it could be useful to add tests

pkg/cmd/infra/router/template.go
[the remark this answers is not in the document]
davidesalerno: @jcmoraisjr have you evaluated the chance to add some unit tests?

pkg/router/client/haproxyclient.go
```diff
@@ -0,0 +1,82 @@
...
+	err = conn.SetDeadline(time.Now().Add(timeout))
+	if err != nil {
+		return "", err
+	}
+	// Immediatelly cancel the connection in case the context is cancelled.
```
davidesalerno: Immediatelly -> Immediately ?

pkg/router/client/haproxyclient.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	// Half close the connection, this tells HAProxy to close its write counterpart
+	// (our read) as soon as it finishes to respond.
+	err = rawConn.CloseWrite()
+	if err != nil {
+		return "", err
+	}
+
+	// Reading response into a loop and populating a buffer with the response chuncks.
```
davidesalerno: chuncks -> chunks ?

pkg/router/template/router.go
```diff
@@ -679,7 +698,25 @@ func (r *templateRouter) reloadRouter(shutdown bool) error {
...
+	// HAProxy runs as a native sidecar, to it should always be up and running
```
davidesalerno: Is there a typo here "to" -> "so" ?

images/router/haproxy/start-haproxy
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+set -euo pipefail
+
+##
+## This script expects HAProxy running in foreground, do not initialize it in daemon mode!
+## Also, haproxy binary should be in the PATH.
+
+gracefulTimeoutStr=${ROUTER_GRACEFUL_SHUTDOWN_DELAY:-45s}
```
gcs278: So this is a bit confusing because `ROUTER_GRACEFUL_SHUTDOWN_DELAY` already exist as a router env variable which adds a sleep to the router shutdown so load balancers can remove the endpoint from rotation.

However, this is about sending HAProxy a SIGTERM - previously it was called [`ROUTER_MAX_SHUTDOWN_TIMEOUT`](https://github.com/openshift/router/blob/master/images/router/haproxy/reload-haproxy#L16) in the old script. 

Should we just reuse the ROUTER_MAX_SHUTDOWN_TIMEOUT or HAPROXY_MAX_SHUTDOWN_TIMEOUT name? Was this intentional to reuse this name?

gcs278 (commented):
Looking good - all of my previous comments were addressed - and the new haproxyclient with context support looks good!

The only thing I'm slightly concerned with is the timeout and I do want to emphasize that adding some edge case unit tests is essential as soon as possible, but I won't block the PR if you want to do it in a follow up.

pkg/router/template/router.go
jcmoraisjr: Thanks to the new client, it's now easy to mock and add a proper unit tests. Just added.

pkg/router/client/haproxyclient.go
jcmoraisjr: Added an opts param which defaults to 5s, overriding it to 1m on the reload call. How does it sound?

pkg/router/client/haproxyclient_test.go
jcmoraisjr: Nice, and agreed. The plan is to spend a few more time in the client during the migration, having a good balance for now between simplicity and a proper coverage.

images/router/haproxy/start-haproxy
jcmoraisjr: Ah my fault, I misunderstood the meaning of the graceful shutdown and incorrectly chose it here. Note that I didn't change the envvar used, I just changed the var in the script to not reuse the `timeout` keyword.

Regarding the new name, I'd say I liked the `HAPROXY_MAX_SHUTDOWN_TIMEOUT` the most, and since neither of them are currently declared by the operator to the router, we can start with the one exclusively to HAProxy. How about it?

gcs278: I think that makes sense. I believe it's the first "HAPROXY" side car container env var - so I like the idea of pre-fixing the env vars this way :+1:

pkg/router/client/haproxyclient.go
gcs278: Yep seems reasonable!

gcs278:
Overall the PR looks good! I think there's room for more unit test coverage (maybe around the `show proc` parsing), but that's not a blocker. Thanks for all of the hard work @jcmoraisjr!

I'll defer to @davidesalerno or @bentito to see if they have any more comments, and they can apply the LGTM

/approve

davidesalerno:
@bentito PR now lgtm, would you like to review it?

rhamini3:
Tested on a 5.0 cluster and confirmed the new topology behavior along with some basic route tests and haproxy sanity tests

running on HAProxy version 2.8.18-ae90be6
```
Router pods show 2 containers

oc -n openshift-ingress get pods
router-default-5d5f7f9c69-2nh8f   2/2     Running   0          3h30m
router-default-5d5f7f9c69-z868l   2/2     Running   0          3h30m
```

Router logs show reloaded with `mode=sidecar` 
```
I0616 15:08:44.188486      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0616 15:08:59.209241      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0616 15:09:04.204632      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0616 15:09:24.010979      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0616 15:09:28.983486      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0616 15:10:07.309205      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0616 15:10:29.100705      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0616 15:10:34.118989      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0616 15:10:39.109624      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
I0616 15:11:01.500647      17 router.go:719] "msg"="router reloaded" "logger"="template" "mode"="sidecar"
```

router pod creates and starts 3 containers with the init-container terminated (HAProxy and router running)
```
Init Containers:
  init-router:
    Container ID:  cri-o://9e47e9a3587d69f27b21f4c3dfad144a9c0478526c36737d6a663847741d7b4f
    Image:         registry.build10.ci.openshift.org/ci-ln-9wnzt0k/stable@sha256:9f4db018b7aaf6763739c599786e92c3267e56e5f302ff01b861a663ba02df6a
    Image ID:      registry.build10.ci.openshift.org/ci-ln-9wnzt0k/stable@sha256:074ef5f089114f89fd01b704380093d7b95311aa5c312c56e7312a9ddb0093b7
    Port:          <none>
    Host Port:     <none>
    Command:
      /bin/bash
      -c
      cp -R -p /var/lib/haproxy/* /mnt/config/
    State:          Terminated
      Reason:       Completed
...
  haproxy:
    Container ID:  cri-o://56de5699ddbce7557afbc9a59540c99b2e634977d5bb1af010295d7c936948ab
    Image:         registry.build10.ci.openshift.org/ci-ln-9wnzt0k/stable@sha256:9f4db018b7aaf6763739c599786e92c3267e56e5f302ff01b861a663ba02df6a
    Image ID:      registry.build10.ci.openshift.org/ci-ln-9wnzt0k/stable@sha256:074ef5f089114f89fd01b704380093d7b95311aa5c312c56e7312a9ddb0093b7
    Port:          <none>
    Host Port:     <none>
    Command:
      /var/lib/haproxy/start-haproxy
    Args:
      -W
      -db
      -S
      /var/lib/haproxy/run/admin.sock,mode,600
      -f
      /var/lib/haproxy/conf/haproxy.config
    State:          Running
      Started:      Tue, 16 Jun 2026 11:06:48 -0400
    Ready:          True
...
Containers:
  router:
    Container ID:   cri-o://55b2f9975175941822014f98b79be72ca240a0b6a853b61b95cef9f69eafc4fe
    Image:          registry.build10.ci.openshift.org/ci-ln-9wnzt0k/stable@sha256:9f4db018b7aaf6763739c599786e92c3267e56e5f302ff01b861a663ba02df6a
    Image ID:       registry.build10.ci.openshift.org/ci-ln-9wnzt0k/stable@sha256:074ef5f089114f89fd01b704380093d7b95311aa5c312c56e7312a9ddb0093b7
    Ports:          80/TCP, 443/TCP, 1936/TCP
    Host Ports:     0/TCP, 0/TCP, 0/TCP
    State:          Running
      Started:      Tue, 16 Jun 2026 11:06:50 -0400

 Type     Reason            Age                  From               Message
  ----     ------            ----                 ----               -------
  Warning  FailedScheduling  128m (x2 over 133m)  default-scheduler  0/3 nodes are available: 3 node(s) had untolerated taint(s). no new claims to deallocate, preemption: 0/3 nodes are available: 3 Preemption is not helpful for scheduling.
  Normal   Scheduled         126m                 default-scheduler  Successfully assigned openshift-ingress/router-default-5d5f7f9c69-z868l to ip-10-0-17-234.ec2.internal
  Warning  FailedScheduling  132m (x2 over 137m)  default-scheduler  0/3 nodes are available: 3 node(s) had untolerated taint(s). no new claims to deallocate, preemption: 0/3 nodes are available: 3 Preemption is not helpful for scheduling.
  Normal   AddedInterface    126m                 multus             Add eth0 [10.129.2.10/23] from ovn-kubernetes
  Normal   Pulling           126m                 kubelet            Pulling image "registry.build10.ci.openshift.org/ci-ln-9wnzt0k/stable@sha256:9f4db018b7aaf6763739c599786e92c3267e56e5f302ff01b861a663ba02df6a"
  Normal   Created           126m                 kubelet            Container created
  Normal   Pulled            126m                 kubelet            Successfully pulled image "registry.build10.ci.openshift.org/ci-ln-9wnzt0k/stable@sha256:9f4db018b7aaf6763739c599786e92c3267e56e5f302ff01b861a663ba02df6a" in 18.397s (18.397s including waiting). Image size: 387823953 bytes.
  Normal   Started           126m                 kubelet            Container started
  Normal   Pulled            126m                 kubelet            Container image "registry.build10.ci.openshift.org/ci-ln-9wnzt0k/stable@sha256:9f4db018b7aaf6763739c599786e92c3267e56e5f302ff01b861a663ba02df6a" already present on machine and can be accessed by the pod
  Normal   Created           126m                 kubelet            Container created
  Normal   Started           126m                 kubelet            Container started
  Normal   Pulled            126m                 kubelet            Container image "registry.build10.ci.openshift.org/ci-ln-9wnzt0k/stable@sha256:9f4db018b7aaf6763739c599786e92c3267e56e5f302ff01b861a663ba02df6a" already present on machine and can be accessed by the pod
  Normal   Created           126m                 kubelet            Container created
  Normal   Started           126m                 kubelet            Container started
```

/verified by @rhamini3

pkg/cmd/infra/router/template.go
```diff
@@ -984,3 +998,50 @@ func getStatsAuth(statsUsernameFile, statsPasswordFile, statsUsername, statsPass
...
+		//
+		output, err := client.RunCommand(ctx, "unix://"+unixSocket, "show proc", client.ClientOpts{})
+		if err != nil {
+			return 0, fmt.Errorf("error reading the current haproxy PID: %w", err)
+		}
+		lines := strings.Split(output, "\n")
+		if len(lines) < 4 {
+			return 0, fmt.Errorf("show proc command returned an unexpected response: %s", output)
+		}
+		workerFields := strings.Fields(lines[3])
```
bentito, line 1037: I think this could be just a follow up tiny PR:

Hardcodes the assumption that the worker process will always be on line 4 (index 3). If a future HAProxy version adds an extra header line, or if the configuration allows multiple workers, this will either parse the wrong PID or fail, entirely breaking Prometheus process metric collection

PR #780, opened 2026-05-27
NE-2218: Add HAProxy standalone images for 2.8 and 3.2

jcmoraisjr:
Adding HAProxy standalone image for the 2.8 and 3.2 branches, without router binary and configurations files. This image is intended to be part of the OCP bundle and should run as a sidecar of the Router pod.

These images are intended to be light weight, without scripts or base configuration. Router pod is responsible to configure one of them via a shared volume instead.

https://redhat.atlassian.net/browse/NE-2218

## Summary by CodeRabbit

* **New Features**
  * Added a new OpenShift HAProxy 2.8 container image (RHEL 9–based) with included runtime tooling.
  * Added a new OpenShift HAProxy 3.2 container image.
  * Both images include security hardening measures, run as a non-root user, and expose standard HTTP/HTTPS ports (80, 443).

### DIFF
--- a/images/router/haproxy28/Dockerfile.ocp
+++ b/images/router/haproxy28/Dockerfile.ocp
@@ -1,0 +1,12 @@
+FROM registry.ci.openshift.org/ocp/5.0:base-rhel9
+RUN INSTALL_PKGS="socat haproxy28" && \
+    yum install -y --setopt=install_weak_deps=0 $INSTALL_PKGS && \
+    rpm -V $INSTALL_PKGS && \
+    yum clean all && \
+    setcap 'cap_net_bind_service=ep' /usr/sbin/haproxy && \
+    sed -i 's/SECLEVEL=2/SECLEVEL=1/g' /etc/crypto-policies/back-ends/opensslcnf.config
+LABEL io.k8s.display-name="OpenShift HAProxy 2.8" \
+      io.k8s.description="This component exposes the Ingress and Route rules, running as a sidecar of the Router component." \
+      io.openshift.tags="openshift,haproxy"
+USER 1001
+EXPOSE 80 443
--- a/images/router/haproxy32/Dockerfile.ocp
+++ b/images/router/haproxy32/Dockerfile.ocp
@@ -1,0 +1,12 @@
+FROM registry.ci.openshift.org/ocp/5.0:base-rhel9
+RUN INSTALL_PKGS="socat haproxy32" && \
+    yum install -y --setopt=install_weak_deps=0 $INSTALL_PKGS && \
+    rpm -V $INSTALL_PKGS && \
+    yum clean all && \
+    setcap 'cap_net_bind_service=ep' /usr/sbin/haproxy && \
+    sed -i 's/SECLEVEL=2/SECLEVEL=1/g' /etc/crypto-policies/back-ends/opensslcnf.config
+LABEL io.k8s.display-name="OpenShift HAProxy 3.2" \
+      io.k8s.description="This component exposes the Ingress and Route rules, running as a sidecar of the Router component." \
+      io.openshift.tags="openshift,haproxy"
+USER 1001
+EXPOSE 80 443


images/router/haproxy28/Dockerfile.ocp
Miciah, line 1: I think of `images/router/foo` as corresponding to the "foo" image and `images/router/foo.bar` and `images/router/foo.baz` as corresponding to alternative builds of the "foo" image.  For example, you could have separate Dockerfiles for building for Centos versus RHEL, or for building using UBI versus a RHEL base image.  However, those Dockerfiles build variants of the same logical "foo" image.

Based on that reasoning, would it make sense to rename this file to `images/router/haproxy28/Dockerfile`?  We could later add a `images/router/haproxy32/Dockerfile` file, and keep `images/router/haproxy` for now as the image that combines the controller and proxy in the same image.

images/router/haproxy/Dockerfile.ocp.haproxy28
```diff
@@ -0,0 +1,12 @@
+FROM registry.ci.openshift.org/ocp/4.22:base-rhel9
```
Miciah: Any reason not to bump this to `FROM registry.ci.openshift.org/ocp/4.23:base-rhel9` instead of 4.22?

images/router/haproxy28/Dockerfile.ocp
jcmoraisjr: Separating router+haproxy and haproxy only looks good to me. However we are going to create a new image every year if we choose to closely follow LTS, so we'd end up creating a new directory every year. Does this change anything in the proposal, like, separate images on another dir but still distinguish them via the filename suffix?

Miciah: We should check with ART.  The alternative would be to name the images something like "haproxy" and "haproxy-next", or "haproxy-old" and "haproxy".  However, in any case, I would expect each image to have its own directory entry.

images/router/haproxy/Dockerfile.ocp.haproxy28
jcmoraisjr: I just made a copy of the router's base one, didn't realize 4.23 is already in place. Just updated.

images/router/haproxy28/Dockerfile.ocp
jcmoraisjr: We talked about this and agreed on using `haproxyNN` names, `NN` being major.minor version from HAProxy, and separate in distinct directories. This follows the approach ART asked us to use in dist-git repos. Also, we can remove old directories as soon as they are not part of the OCP bundle anymore, this will maintain a limited number of `haproxyNN` directories, probably up to 3.

images/router/haproxy/Dockerfile.ocp.haproxy28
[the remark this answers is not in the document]
jcmoraisjr: Just updated to 4.23. Regarding the base image, this is the one we already use for other OCP components, so we are just following what others already use.

gcs278:
I don't have any major concern. Loosing the `ps` and `kill` commands is a bit of a inconvenience for debugging - but we are sharing process namespaces so we can debug from the router container. I personally use it often in debugging our containers. If we drop processnamespace sharing - then I think it will be much more of a need to add it back.

The Code Rabbit comments are actually not specific to your new Dockerfiles - but apply to our existing Dockerfile. I don't think it's something we need to block the PR over, but I'm interested in someone following up (should we be using the UBI images)? Feel free to research and create bugs on these if you want to follow up.

For testing - you could merge this with no issues, but we don't have a HAProxy 3.2 RPM yet, nor the openshift/release or ocp-build-data PRs are uP (correct me if I'm wrong). So I don't think there is a rush.

@Miciah i can leave the final review and approval to you.

/lgtm

jcmoraisjr:
> [...] we don't have a HAProxy 3.2 RPM yet, nor the openshift/release or ocp-build-data PRs are uP (correct me if I'm wrong). So I don't think there is a rush.

I'm specially interested on an agreement about the approach we should use. About merging, it should be noop until ART configures `ocp-build-data`.

On the `ocp-build-data` side, to the best of my knowledge, we should only ask ART to update it after having this PR merged and `haproxy32` RPM built.

gcs278:
This is a pretty straightforward PR, and @Miciah has reviewed it once already.

/approve

gcs278:
this PR is a a NO-OP so CI failures are unrelated:
/retest

gcs278:
`e2e-aws-fips` has been very flakey lately, I [followed](https://redhat-internal.slack.com/archives/C01CQA76KMX/p1782329725173379) up previously on it - and got agreement on overriding on one of my origin PRs.

Let's let the current job finish, but I think we should override FIPs.

I also followed up with the `e2e-agnostic` failures in https://redhat.atlassian.net/browse/OCPBUGS-59496?focusedCommentId=17426904. It looks to be related to OCPBUGS-59496, however much more flakey. If it fails again, I think we can override as well.

gcs278:
e2e-aws-fips is very flakey with a 10-20% pass rate. This PR is effectively a NO-OP, it just adds new container files that are unused. I've [followed](https://redhat-internal.slack.com/archives/C01CQA76KMX/p1782329725173379) up with the TRT team before, they agreed that the nightly job is passing nicely, so they have no concerns for an override.

/override ci/prow/e2e-aws-fips

gcs278:
Actually, just one nit left
/hold

images/router/haproxy28/Dockerfile.ocp
```diff
@@ -0,0 +1,12 @@
+FROM registry.ci.openshift.org/ocp/4.23:base-rhel9
```
gcs278: Should these be 5.0, not 4.23? 

```suggestion
FROM registry.ci.openshift.org/ocp/5.0:base-rhel9
```

gcs278: 4.23 is not found.

```
REGISTRY_AUTH_FILE=/home/gspence/.secrets/pull-secret.txt oc image info registry.ci.openshift.org/ocp/5.0:base-rhel10
error: the image is a manifest list and contains multiple images - use --filter-by-os to select from:

  OS             DIGEST
  linux/amd64    sha256:7eb59cf389d6c111cc92876fccfb50ebb8501c97e0e3923310c1d50d33b871cc
  linux/arm64/v8 sha256:7fbfd0dea4d4a3f132bce6ede6ee15b8e17a043acb9d222de60842b1a0d7b8bf
  linux/s390x    sha256:925ef65b3df3fc0c528b46feb70eeaa6178d8c66007a43923dbaeec541294dff
  linux/ppc64le  sha256:98466ff11d17038ac5d211edb262567a0f867c6311aee9960df5240b4dc0187f

thinkpadp1gen3:~/src/github.com/openshift/router$ REGISTRY_AUTH_FILE=/home/gspence/.secrets/pull-secret.txt oc image info registry.ci.openshift.org/ocp/4.23:base-rhel10
error: image "registry.ci.openshift.org/ocp/4.23:base-rhel10" not found: manifest unknown: manifest unknown
```

jcmoraisjr: Yeah it should according to latest ART PRs. Just updated.

jcmoraisjr:
```
$ podman pull registry.ci.openshift.org/ocp/4.23:base-rhel9
Trying to pull registry.ci.openshift.org/ocp/4.23:base-rhel9...
Getting image source signatures
Copying blob c9b6d021f151 done   | 
Copying blob cd999a7deb91 done   | 
Copying blob a2b5995a0cf3 skipped: already exists  
Copying blob c831420d7a68 done   | 
Copying config 9abf3cb055 done   | 
Writing manifest to image destination
9abf3cb0559943a906e849c50968217ce8c884a32a3613b7d989479c0859d065
```

rhel9 does exist though, but it wasn't accurate anyway.

gcs278: Oh I see I accidentally tried rhel10. Sounds good.

gcs278:
Thanks for the last minute fix. I tried running these dockerfile locally, but was unable to get the yum installation to work. I think you can go ahead and make a openshift/release PR with the new container images, then try to pj-rehearse that to make sure these build.

/lgtm

gcs278:
@rhamini3 I think we will need another verified later

PR #782, opened 2026-06-03
OCPBUGS-74947: Adding escapeHAProxySingleQuotes for sanitize [release-4.14]

Thealisyed:
This function will allow users to include
certain characters and spacing to avoid
any config injection

Adapted router_test.go to use `mustCreate` struct (`mustCreateRoute` does not exist on this branch).

PR #783, opened 2026-06-03
OCPBUGS-74945: Adding escapeHAProxySingleQuotes for sanitize [release-4.13]

Thealisyed:
This function will allow users to include
certain characters and spacing to avoid
any config injection

Adapted router_test.go to use `mustCreate` struct (`mustCreateRoute` does not exist on this branch).

### DIFF
--- a/pkg/router/router_test.go
+++ b/pkg/router/router_test.go
@@ -408,6 +408,63 @@ func TestConfigTemplate(t *testing.T) {
 				},
 			},
 		},
+		"Rewrite target with root path": {
+			mustCreateWithConfig{
+				mustCreate: mustCreate{
+					name: "rewrite-root",
+					host: "rewrite-root.example.com",
+					path: "/app",
+					time: start,
+					annotations: map[string]string{
+						"haproxy.router.openshift.io/rewrite-target": "/",
+					},
+				},
+				mustMatchConfig: mustMatchConfig{
+					section:     "backend",
+					sectionName: insecureBackendName(h.namespace, "rewrite-root"),
+					attribute:   "http-request",
+					value:       `replace-path '^/app/?(.*)$' '/\1'`,
+				},
+			},
+		},
+		"Rewrite target with non-root path": {
+			mustCreateWithConfig{
+				mustCreate: mustCreate{
+					name: "rewrite-nonroot",
+					host: "rewrite-nonroot.example.com",
+					path: "/api/v1",
+					time: start,
+					annotations: map[string]string{
+						"haproxy.router.openshift.io/rewrite-target": "/v2",
+					},
+				},
+				mustMatchConfig: mustMatchConfig{
+					section:     "backend",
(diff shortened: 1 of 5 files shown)


Miciah:
/approve

This is a risky backport that could break routes that use special characters such as backslashes in `spec.path`.

/label backport-risk-assessed

@Thealisyed, could you reverse the order of the commits, or squash them, so that the first commit doesn't cause `make check` to fail?

Thealisyed:
infra issue

/retest

Thealisyed:
Prow's blobless clone failed with `Could not resolve host: github.com` during lazy blob fetch after merge. Unrelated to code changes (same diff passes on #782 and #785)

rhamini3:
Tested on a 4.13 cluster

```
% oc annotate route edg-route haproxy.router.openshift.io/rewrite-target="/it's" --overwrite
route.route.openshift.io/edg-route annotated
% oc get route edg-route -o yaml 
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  annotations:
    haproxy.router.openshift.io/rewrite-target: /it's
    openshift.io/host.generated: "true"
% oc -n openshift-ingress rsh router-default-654f74c94c-46lxr
sh-4.4$ cat haproxy.config | grep edg-route -A20
backend be_edge_http:default:edg-route
  mode http
  option redispatch
  option forwardfor
  balance random

  timeout check 5000ms
  # Path rewrite target
  http-request replace-path '^(.*)$' '/its\1'
  cookie 52eb704682406c0b9dc7ca0797a30db9 insert indirect nocache httponly secure attr SameSite=None
```
haproxy shows the single quotes as expected
/verified by @rhamini3

pkg/router/router_test.go
```diff
@@ -658,3 +715,7 @@ func createRouterDirs() {
 func edgeBackendName(ns, route string) string {
 	return "be_edge_http:" + ns + ":" + route
 }
+
+func insecureBackendName(ns, route string) string {
+	return "be_http:" + ns + ":" + route
+}
```
Miciah, line 721: This definition is an addition to the release-4.14 backport.  Ideally the commit message should mention this addition, but in the interest of time, we can leave it as is.

Miciah:
Thanks!  

This is a mostly clean backport.

/lgtm

[ci/prow/e2e-aws-serial](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_router/783/pull-ci-openshift-router-release-4.13-e2e-aws-serial/2063154569818411008) failed because the `[sig-storage] Pod Disks [Feature:StorageProvider] [Serial] attach on previously attached volumes should work` test failed with `error listing AWS instances: MissingRegion: could not find region configuration`.  This is a known issue that is being tracked with [OCPBUGS-77660](https://redhat.atlassian.net/browse/OCPBUGS-77660), and we are confident that the error is not caused by the changes in this PR, so I am overriding the failing job.

/override ci/prow/e2e-aws-serial

PR #787, opened 2026-06-04
OCPBUGS-87205: Strip X-SSL-Client-* headers for plain HTTP

bentito:
This prevents unauthenticated spoofing of mutual TLS client identities by deleting X-SSL-Client headers in the fe_http frontend before they can reach the backend.

It introduces a new environment variable `ROUTER_MUTUAL_TLS_AUTH_FILTER`
which defaults to `true`. Setting this to `false` allows external load balancers
to inject these headers.

Joint effort of the author and gemini-pro-3.1-preview

## Summary by CodeRabbit

* **Bug Fixes**
  * Strips mutual-TLS client identity headers from requests forwarded over plain HTTP.
  * Applies this header filtering consistently across edge-terminated, SNI, and non-TLS request flows.
  * Adds an environment toggle to enable/disable the header filtering (enabled by default).

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -275,7 +275,7 @@ frontend public
 
   # Strip off X-SSL-Client-* headers for plain HTTP if not explicitly disabled.
   # This prevents unauthenticated spoofing of mutual TLS client identities.
-  {{- if isTrue (env "ROUTER_MUTUAL_TLS_AUTH_FILTER" "true") }}
+  {{- if isTrue (env "ROUTER_MUTUAL_TLS_HEADER_FILTER" "true") }}
   http-request del-header X-SSL-Client-DN
   http-request del-header X-SSL-Client-DER
   http-request del-header X-SSL-Client-NotAfter
@@ -403,7 +403,7 @@ frontend fe_sni
 
   # Strip off X-SSL-Client-* headers for plain HTTP if not explicitly disabled.
   # This prevents unauthenticated spoofing of mutual TLS client identities.
-  {{- if isTrue (env "ROUTER_MUTUAL_TLS_AUTH_FILTER" "true") }}
+  {{- if isTrue (env "ROUTER_MUTUAL_TLS_HEADER_FILTER" "true") }}
   http-request del-header X-SSL-Client-DN
   http-request del-header X-SSL-Client-DER
   http-request del-header X-SSL-Client-NotAfter
@@ -529,7 +529,7 @@ frontend fe_no_sni
 
   # Strip off X-SSL-Client-* headers for plain HTTP if not explicitly disabled.
   # This prevents unauthenticated spoofing of mutual TLS client identities.
-  {{- if isTrue (env "ROUTER_MUTUAL_TLS_AUTH_FILTER" "true") }}
+  {{- if isTrue (env "ROUTER_MUTUAL_TLS_HEADER_FILTER" "true") }}
   http-request del-header X-SSL-Client-DN
   http-request del-header X-SSL-Client-DER
   http-request del-header X-SSL-Client-NotAfter
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -276,12 +276,18 @@ frontend public
   # Strip off X-SSL-Client-* headers for plain HTTP if not explicitly disabled.
   # This prevents unauthenticated spoofing of mutual TLS client identities.
   {{- if isTrue (env "ROUTER_MUTUAL_TLS_HEADER_FILTER" "true") }}
-  http-request del-header X-SSL-Client-DN
+  http-request del-header X-SSL
+  http-request del-header X-SSL-Client-CN
   http-request del-header X-SSL-Client-DER
+  http-request del-header X-SSL-Client-DN
   http-request del-header X-SSL-Client-NotAfter
   http-request del-header X-SSL-Client-NotBefore
   http-request del-header X-SSL-Client-SHA1
+  http-request del-header X-SSL-Client-Serial
   http-request del-header X-SSL-Client-Subject
+  http-request del-header X-SSL-Client-Verify
+  http-request del-header X-SSL-Client-Version
+  http-request del-header X-SSL-Issuer
   {{- end }}
 
   # DNS labels are case insensitive (RFC 4343), we need to convert the hostname into lowercase
@@ -404,12 +410,18 @@ frontend fe_sni
   # Strip off X-SSL-Client-* headers for plain HTTP if not explicitly disabled.
   # This prevents unauthenticated spoofing of mutual TLS client identities.
   {{- if isTrue (env "ROUTER_MUTUAL_TLS_HEADER_FILTER" "true") }}
-  http-request del-header X-SSL-Client-DN
+  http-request del-header X-SSL
+  http-request del-header X-SSL-Client-CN
   http-request del-header X-SSL-Client-DER
+  http-request del-header X-SSL-Client-DN
   http-request del-header X-SSL-Client-NotAfter
   http-request del-header X-SSL-Client-NotBefore
   http-request del-header X-SSL-Client-SHA1
+  http-request del-header X-SSL-Client-Serial
   http-request del-header X-SSL-Client-Subject
+  http-request del-header X-SSL-Client-Verify
+  http-request del-header X-SSL-Client-Version
+  http-request del-header X-SSL-Issuer
   {{- end }}
 
   # DNS labels are case insensitive (RFC 4343), we need to convert the hostname into lowercase
@@ -530,12 +542,18 @@ frontend fe_no_sni
   # Strip off X-SSL-Client-* headers for plain HTTP if not explicitly disabled.
   # This prevents unauthenticated spoofing of mutual TLS client identities.
   {{- if isTrue (env "ROUTER_MUTUAL_TLS_HEADER_FILTER" "true") }}
-  http-request del-header X-SSL-Client-DN
+  http-request del-header X-SSL
+  http-request del-header X-SSL-Client-CN
   http-request del-header X-SSL-Client-DER
+  http-request del-header X-SSL-Client-DN
   http-request del-header X-SSL-Client-NotAfter
   http-request del-header X-SSL-Client-NotBefore
   http-request del-header X-SSL-Client-SHA1
+  http-request del-header X-SSL-Client-Serial
   http-request del-header X-SSL-Client-Subject
+  http-request del-header X-SSL-Client-Verify
+  http-request del-header X-SSL-Client-Version
+  http-request del-header X-SSL-Issuer
   {{- end }}
 
   # DNS labels are case insensitive (RFC 4343), we need to convert the hostname into lowercase
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -273,6 +273,17 @@ frontend public
   # Strip off Proxy headers to prevent HTTpoxy (https://httpoxy.org/)
   http-request del-header Proxy
 
+  # Strip off X-SSL-Client-* headers for plain HTTP if not explicitly disabled.
+  # This prevents unauthenticated spoofing of mutual TLS client identities.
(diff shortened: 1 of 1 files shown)


images/router/haproxy/conf/haproxy-config.template
```diff
@@ -273,6 +273,17 @@ frontend public
...
+  # Strip off X-SSL-Client-* headers for plain HTTP if not explicitly disabled.
+  # This prevents unauthenticated spoofing of mutual TLS client identities.
+  {{- if isTrue (env "ROUTER_MUTUAL_TLS_AUTH_FILTER" "true") }}
+  http-request del-header X-SSL-Client-DN
+  http-request del-header X-SSL-Client-DER
+  http-request del-header X-SSL-Client-NotAfter
+  http-request del-header X-SSL-Client-NotBefore
+  http-request del-header X-SSL-Client-SHA1
+  http-request del-header X-SSL-Client-Subject
+  {{- end }}
```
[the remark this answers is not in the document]
bentito: Good catch! I've renamed the environment variable to `ROUTER_MUTUAL_TLS_HEADER_FILTER` in the latest commit to avoid the collision with the existing regex functionality. I didn't include the backward compatibility logic because there are no existing deployments of this specific boolean feature (it was just introduced in this PR).

bentito:
@Miciah in https://github.com/openshift/router/pull/787#pullrequestreview-4429940428 coderabbit seems to be correct in having found more mTLS header with the same problem, like X-SSL, X-SSL-Client-Verify, X-SSL-Client-Serial, X-SSL-Client-Version, X-SSL-Client-CN, and X-SSL-Issuer

what do you think, fix these in this PR too?

Miciah:
> coderabbit seems to be correct in having found more mTLS header with the same problem, like X-SSL, X-SSL-Client-Verify, X-SSL-Client-Serial, X-SSL-Client-Version, X-SSL-Client-CN, and X-SSL-Issuer
> 
> what do you think, fix these in this PR too?

Yes, we should delete them all.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -273,6 +273,17 @@ frontend public
   # Strip off Proxy headers to prevent HTTpoxy (https://httpoxy.org/)
   http-request del-header Proxy
 
+  # Strip off X-SSL-Client-* headers for plain HTTP if not explicitly disabled.
+  # This prevents unauthenticated spoofing of mutual TLS client identities.
+  {{- if isTrue (env "ROUTER_MUTUAL_TLS_AUTH_FILTER" "true") }}
```
Miciah: Are you planning to add an API option and logic in cluster-ingress-operator to set this environment variable, or are you adding the environment variable just in case we later need to add an API field or unsupported override?

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -390,6 +401,17 @@ frontend fe_sni
   # Strip off Proxy headers to prevent HTTpoxy (https://httpoxy.org/)
   http-request del-header Proxy
 
+  # Strip off X-SSL-Client-* headers for plain HTTP if not explicitly disabled.
+  # This prevents unauthenticated spoofing of mutual TLS client identities.
```
Miciah: Here in `fe_sni` and also in `fe_no_sni`, you should omit "plain HTTP" (which I interpret as cleartext HTTP) and note that these deletions are specifically for the case that mTLS is *not* enabled:
```suggestion
  # Strip off X-SSL-Client-* headers if not explicitly disabled.
  # This prevents unauthenticated spoofing of mutual TLS client identities
  # when mutual TLS is not enabled and so the headers are not set below.  
```

images/router/haproxy/conf/haproxy-config.template
bentito: Idea was in case a user has got a need for this, there will be a way. Is that sufficient for now?

rikatz: I think the flag will also need to come with a documentation that says "if you enable this, be sure you add a network policy on your router to trust HTTP traffic just from your loadbalancer"

bentito: @rikatz would you want to create a DOCS tracker story please

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -273,6 +273,23 @@ frontend public
...
+  {{- if isTrue (env "ROUTER_MUTUAL_TLS_HEADER_FILTER" "true") }}
```
Miciah, line 278: The indentation of these Go template stanzas is inconsistent with surrounding Go template stanzas, but the template already has similar inconsistencies elsewhere; we can fix them all up in a follow-up.

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -273,6 +273,23 @@ frontend public
...
+  http-request del-header X-SSL
+  http-request del-header X-SSL-Client-CN
+  http-request del-header X-SSL-Client-DER
+  http-request del-header X-SSL-Client-DN
+  http-request del-header X-SSL-Client-NotAfter
+  http-request del-header X-SSL-Client-NotBefore
+  http-request del-header X-SSL-Client-SHA1
+  http-request del-header X-SSL-Client-Serial
+  http-request del-header X-SSL-Client-Subject
```
Miciah, line 287: Where is `X-SSL-Client-Subject` added?  Why does it need to be deleted?

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -273,6 +273,23 @@ frontend public
   # Strip off Proxy headers to prevent HTTpoxy (https://httpoxy.org/)
   http-request del-header Proxy
 
+  # Strip off X-SSL-Client-* headers for plain HTTP if not explicitly disabled.
```
Miciah: With the latest changes, `X-SSL-Client-*` should be changed to `X-SSL*` here and below.
```suggestion
  # Strip off X-SSL* headers for plain HTTP if not explicitly disabled.
```

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -505,6 +540,23 @@ frontend fe_no_sni
...
+  # This prevents unauthenticated spoofing of mutual TLS client identities.
```
Miciah: The comment in `fe_no_sni` should match the one in `fe_sni`, not the one in `public`.

images/router/haproxy/conf/haproxy-config.template
rikatz: While we don't set them on router, I think it is a valid deletion. This header can also be used for authentication, in case some day we decide to add it to router passing to backend by a customer request we can run on a problem of forgetting to remove it

images/router/haproxy/conf/haproxy-config.template
rikatz: https://redhat.atlassian.net/browse/OSDOCS-20071

images/router/haproxy/conf/haproxy-config.template
Miciah: It seems arbitrary.  Why not also add `http-request del-header X-SSL-Client-Chain-DER` or `http-request del-header X-SSL-Client-Used` or other arbitrary things that an application could theoretically use or HAProxy could theoretically set?

Miciah:
Thanks!

/approve
/lgtm

By the way, do you want to squash the commits?

melvinjoseph86:
Check this PR with clusterbot

```
➜  oc get clusterversion
version   5.0.0-0-2026-06-09-165857-test-ci-ln-798yy8k-latest   True        False         3m59s   Cluster version is 5.0.0-0-2026-06-09-165857-test-ci-ln-798yy8k-latest
➜  oc get po -n openshift-ingress
router-default-7f5b6b4556-8bth7   1/1     Running   0          31m
router-default-7f5b6b4556-x496l   1/1     Running   0          31m
➜  oc rsh -n openshift-ingress router-default-7f5b6b4556-8bth7 
sh-5.1$ cat haproxy-config.template

<---snip----->

  # Strip off Proxy headers to prevent HTTpoxy (https://httpoxy.org/)
  http-request del-header Proxy

  # Strip off X-SSL* headers for plain HTTP if not explicitly disabled.
  # This prevents unauthenticated spoofing of mutual TLS client identities.
  {{- if isTrue (env "ROUTER_MUTUAL_TLS_HEADER_FILTER" "true") }}
  http-request del-header X-SSL
  http-request del-header X-SSL-Client-CN
  http-request del-header X-SSL-Client-DER
  http-request del-header X-SSL-Client-DN
  http-request del-header X-SSL-Client-NotAfter
  http-request del-header X-SSL-Client-NotBefore
  http-request del-header X-SSL-Client-SHA1
  http-request del-header X-SSL-Client-Serial
  http-request del-header X-SSL-Client-Subject
  http-request del-header X-SSL-Client-Verify
  http-request del-header X-SSL-Client-Version
  http-request del-header X-SSL-Issuer
  {{- end }}

  # DNS labels are case insensitive (RFC 4343), we need to convert the hostname into lowercase
  # before matching, or any requests containing uppercase characters will never match.
  http-request set-header Host %[req.hdr(Host),lower]

<---snip----->

  # Strip off Proxy headers to prevent HTTpoxy (https://httpoxy.org/)
  http-request del-header Proxy

  # Strip off X-SSL* headers if not explicitly disabled.
  # This prevents unauthenticated spoofing of mutual TLS client identities
  # when mutual TLS is not enabled and so the headers are not set below.  
  {{- if isTrue (env "ROUTER_MUTUAL_TLS_HEADER_FILTER" "true") }}
  http-request del-header X-SSL
  http-request del-header X-SSL-Client-CN
  http-request del-header X-SSL-Client-DER
  http-request del-header X-SSL-Client-DN
  http-request del-header X-SSL-Client-NotAfter
  http-request del-header X-SSL-Client-NotBefore
  http-request del-header X-SSL-Client-SHA1
  http-request del-header X-SSL-Client-Serial
  http-request del-header X-SSL-Client-Subject
  http-request del-header X-SSL-Client-Verify
  http-request del-header X-SSL-Client-Version
  http-request del-header X-SSL-Issuer
  {{- end }}

  # DNS labels are case insensitive (RFC 4343), we need to convert the hostname into lowercase
  # before matching, or any requests containing uppercase characters will never match.
  http-request set-header Host %[req.hdr(Host),lower]

<---snip----->

  # Strip off Proxy headers to prevent HTTpoxy (https://httpoxy.org/)
  http-request del-header Proxy

  # Strip off X-SSL* headers if not explicitly disabled.
  # This prevents unauthenticated spoofing of mutual TLS client identities
  # when mutual TLS is not enabled and so the headers are not set below.
  {{- if isTrue (env "ROUTER_MUTUAL_TLS_HEADER_FILTER" "true") }}
  http-request del-header X-SSL
  http-request del-header X-SSL-Client-CN
  http-request del-header X-SSL-Client-DER
  http-request del-header X-SSL-Client-DN
  http-request del-header X-SSL-Client-NotAfter
  http-request del-header X-SSL-Client-NotBefore
  http-request del-header X-SSL-Client-SHA1
  http-request del-header X-SSL-Client-Serial
  http-request del-header X-SSL-Client-Subject
  http-request del-header X-SSL-Client-Verify
  http-request del-header X-SSL-Client-Version
  http-request del-header X-SSL-Issuer
  {{- end }}

  # DNS labels are case insensitive (RFC 4343), we need to convert the hostname into lowercase
  # before matching, or any requests containing uppercase characters will never match.
  http-request set-header Host %[req.hdr(Host),lower]
```

/verified by @mjoseph

images/router/haproxy/conf/haproxy-config.template
```diff
@@ -273,6 +273,23 @@ frontend public
   # Strip off Proxy headers to prevent HTTpoxy (https://httpoxy.org/)
   http-request del-header Proxy
 
+  # Strip off X-SSL* headers for plain HTTP if not explicitly disabled.
+  # This prevents unauthenticated spoofing of mutual TLS client identities.
+  {{- if isTrue (env "ROUTER_MUTUAL_TLS_HEADER_FILTER" "true") }}
+  http-request del-header X-SSL
+  http-request del-header X-SSL-Client-CN
+  http-request del-header X-SSL-Client-DER
```
rikatz, line 281: is this case insensitive? I forgot to check

PR #788, opened 2026-06-04
OCPBUGS-87204: Prevent SSRF via FQDN-typed EndpointSlices

bentito:
The OpenShift Router previously did not validate backend destinations resolved from FQDN-typed EndpointSlices. This allowed an attacker with EndpointSlice write access to create a Service backed by an FQDN pointing to a hostname that resolves to restricted IPs (like the cloud metadata service at 169.254.169.254) and bypass SSRF protections.

This commit updates the ExtendedValidator to inspect all EndpointSlice addresses. If an address is an FQDN, it performs a DNS lookup and blocks the endpoint if it resolves to a restricted IP address.

Includes unit tests to verify malicious FQDNs are successfully blocked.

Joint effort of the author and gemini-pro-3.1-preview

## Summary by CodeRabbit

* **New Features**
  * Endpoint addresses are now validated and filtered to prevent invalid IPs from being processed
  * FQDN endpoint addresses are automatically resolved to IP addresses for better compatibility
  * Improved endpoint validation with individual error logging for rejected addresses

* **Tests**
  * Added comprehensive test coverage for endpoint validation and FQDN resolution scenarios

### DIFF
--- a/pkg/cmd/infra/router/template.go
+++ b/pkg/cmd/infra/router/template.go
@@ -827,9 +827,7 @@ func (o *TemplateRouterOptions) Run(stopCh <-chan struct{}) error {
 	if o.UpgradeValidation {
 		plugin = controller.NewUpgradeValidation(plugin, recorder, o.UpgradeValidationForceAddCondition, o.UpgradeValidationForceRemoveCondition)
 	}
-	if o.ExtendedValidation {
-		plugin = controller.NewExtendedValidator(plugin, recorder)
-	}
+	plugin = controller.NewExtendedValidator(plugin, recorder, o.ExtendedValidation)
 	if o.AllowExternalCertificates {
 		plugin = controller.NewRouteSecretManager(plugin, recorder, secretManager, o.RouterName, kc.CoreV1(), routeLister, authorizationClient.SubjectAccessReviews())
 	}
--- a/pkg/router/controller/extended_validator.go
+++ b/pkg/router/controller/extended_validator.go
@@ -2,6 +2,8 @@ package controller
 
 import (
 	"fmt"
+	"net/netip"
+	"slices"
 
 	kapi "k8s.io/api/core/v1"
 
@@ -21,15 +23,20 @@ type ExtendedValidator struct {
 
 	// recorder is an interface for indicating route status.
 	recorder RouteStatusRecorder
+
+	// extendedRouteValidation enables extended route validation checks.
+	extendedRouteValidation bool
 }
 
-// NewExtendedValidator creates a plugin wrapper that ensures only routes that
-// pass extended validation are relayed to the next plugin in the chain.
-// Recorder is an interface for indicating route status updates.
-func NewExtendedValidator(plugin router.Plugin, recorder RouteStatusRecorder) *ExtendedValidator {
+// NewExtendedValidator creates a plugin wrapper that ensures only routes and
+// endpoints that pass validation are relayed to the next plugin in the chain.
+// Endpoint address validation is always enabled. Route validation is
+// controlled by extendedRouteValidation.
+func NewExtendedValidator(plugin router.Plugin, recorder RouteStatusRecorder, extendedRouteValidation bool) *ExtendedValidator {
 	return &ExtendedValidator{
-		plugin:   plugin,
-		recorder: recorder,
+		plugin:                  plugin,
+		recorder:                recorder,
+		extendedRouteValidation: extendedRouteValidation,
 	}
 }
 
@@ -39,21 +46,136 @@ func (p *ExtendedValidator) HandleNode(eventType watch.EventType, node *kapi.Nod
 }
 
 // HandleEndpoints processes watch events on the Endpoints resource.
+// Addresses that fail IP validation are removed individually rather
+// than rejecting the entire endpoint set.
 func (p *ExtendedValidator) HandleEndpoints(eventType watch.EventType, endpoints *kapi.Endpoints) error {
+	if eventType == watch.Deleted {
+		return p.plugin.HandleEndpoints(eventType, endpoints)
+	}
+
+	// endpoints may be an object shared with (and owned by) an informer's
+	// cache, so it must be deep-copied before it is mutated below.
+	endpoints = endpoints.DeepCopy()
+	for i, subset := range endpoints.Subsets {
+		endpoints.Subsets[i].Addresses = filterValidAddresses(subset.Addresses)
+		endpoints.Subsets[i].NotReadyAddresses = filterValidAddresses(subset.NotReadyAddresses)
+	}
 	return p.plugin.HandleEndpoints(eventType, endpoints)
+}
+
+func filterValidAddresses(addrs []kapi.EndpointAddress) []kapi.EndpointAddress {
+	return slices.DeleteFunc(addrs, func(addr kapi.EndpointAddress) bool {
+		err := validateEndpointAddress(addr.IP)
+		if err != nil {
+			log.Error(err, "Skipping endpoint address with restricted or invalid IP", "address", addr.IP)
+		}
+		return err != nil
+	})
+}
+
+func validateEndpointAddress(address string) error {
+	addr, err := netip.ParseAddr(address)
+	if err != nil {
+		return fmt.Errorf("address %q is not a valid IP address", address)
+	}
+	// EndpointSlice addresses of type IPv4 or IPv6 are validated by the Kubernetes
+	// API, which requires canonical IP form and rejects hostnames, non-canonical
+	// encodings (such as ::127.0.0.1), and IPv4-mapped IPv6 literals.
+	return checkRestrictedIP(addr)
+}
+
+var (
+	azureMetadata = netip.MustParseAddr("168.63.129.16")
+	awsIPv6IMDS   = netip.MustParseAddr("fd00:ec2::254")
+)
+
+func ipv4CompatibleTo4(addr netip.Addr) (netip.Addr, bool) {
+	if !addr.Is6() || addr.Is4In6() {
+		return netip.Addr{}, false
+	}
+	b := addr.As16()
+	for i := 0; i < 12; i++ {
+		if b[i] != 0 {
+			return netip.Addr{}, false
+		}
+	}
+	return netip.AddrFrom4([4]byte{b[12], b[13], b[14], b[15]}), true
+}
+
+func normalizeEmbeddedIPv4(addr netip.Addr) (netip.Addr, bool) {
+	if addr.Is4In6() {
+		return addr.Unmap(), true
+	}
+	if v4, ok := ipv4CompatibleTo4(addr); ok {
+		return v4, true
+	}
+	return addr, false
+}
+
+func restrictedProperties(addr netip.Addr) error {
+	if addr.IsUnspecified() {
+		return fmt.Errorf("IP address %s is a restricted unspecified IP", addr)
+	}
+	if addr.IsLoopback() {
+		return fmt.Errorf("IP address %s is a restricted loopback IP", addr)
+	}
+	if addr.IsLinkLocalUnicast() {
+		return fmt.Errorf("IP address %s is a restricted link-local IP", addr)
+	}
+	if addr.IsMulticast() {
+		return fmt.Errorf("IP address %s is a restricted multicast IP", addr)
+	}
+	if addr == azureMetadata || addr == awsIPv6IMDS {
+		return fmt.Errorf("IP address %s is a restricted cloud metadata IP", addr)
+	}
+	return nil
+}
+
+// checkRestrictedIP rejects addresses that must not be used as router backends.
+//
+// Validation runs in two phases against a single rule set (restrictedProperties):
+//
+//  1. Check the address as parsed. This catches native forms directly — plain IPv4
+//     (10.0.0.1), native IPv6 loopback (::1), link-local (fe80::1), and cloud-specific
+//     IPv6 metadata (fd00:ec2::254). Most addresses, including essentially all plain
+//     IPv4 endpoints, are fully validated here and return immediately.
+//
+//  2. If the address embeds an IPv4 address inside an IPv6 encoding, extract it and
+//     check again. IPv4-mapped literals (::ffff:x.x.x.x) are unmapped; IPv4-compatible
+//     literals (::x.x.x.x / ::hh:hh:hh:hh, the deprecated ::/96 form) have their low
+//     32 bits decoded. The second pass catches restricted destinations that do not look
+//     restricted in IPv6 form, such as ::a9fe:a9fe (169.254.169.254 metadata) or
+//     ::127.0.0.1 (loopback).
+//
+// The native check must run before extraction: ::1 also matches the ::/96 layout but
+// is IPv6 loopback, not embedded 0.0.0.1; checking after extraction would misclassify
+// it and allow it through.
+//
+// Typical cost: one pass for plain IPv4 and legitimate IPv6; two passes only for
+// IPv4-mapped or IPv4-compatible encodings.
+func checkRestrictedIP(addr netip.Addr) error {
+	if err := restrictedProperties(addr); err != nil {
+		return err
+	}
+	if normalized, ok := normalizeEmbeddedIPv4(addr); ok {
+		return restrictedProperties(normalized)
+	}
+	return nil
 }
 
 // HandleRoute processes watch events on the Route resource.
 func (p *ExtendedValidator) HandleRoute(eventType watch.EventType, route *routev1.Route) error {
 	log.V(10).Info("HandleRoute: ExtendedValidator")
-	// Check if previously seen route and its Spec is unchanged.
-	routeName := routeNameKey(route)
-	if err := routeapihelpers.ExtendedValidateRoute(route).ToAggregate(); err != nil {
-		log.Error(err, "skipping route due to invalid configuration", "route", routeName)
 
-		p.recorder.RecordRouteRejection(route, "ExtendedValidationFailed", err.Error())
-		p.plugin.HandleRoute(watch.Deleted, route)
-		return fmt.Errorf("invalid route configuration")
+	if p.extendedRouteValidation {
+		routeName := routeNameKey(route)
+		if err := routeapihelpers.ExtendedValidateRoute(route).ToAggregate(); err != nil {
+			log.Error(err, "skipping route due to invalid configuration", "route", routeName)
+
+			p.recorder.RecordRouteRejection(route, "ExtendedValidationFailed", err.Error())
+			p.plugin.HandleRoute(watch.Deleted, route)
+			return fmt.Errorf("invalid route configuration")
+		}
 	}
 
 	return p.plugin.HandleRoute(eventType, route)
--- a/pkg/router/controller/extended_validator_test.go
+++ b/pkg/router/controller/extended_validator_test.go
@@ -1,0 +1,676 @@
+package controller
+
+import (
+	"fmt"
+	"net/netip"
+	"testing"
+
+	routev1 "github.com/openshift/api/route/v1"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	kapi "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/watch"
+
+	"github.com/openshift/router/pkg/router/controller/endpointsubset"
+)
+
+type fakeTestPlugin struct {
+	endpoints []*kapi.Endpoints
+}
+
+func (p *fakeTestPlugin) HandleRoute(watch.EventType, *routev1.Route) error { return nil }
+func (p *fakeTestPlugin) HandleEndpoints(eventType watch.EventType, endpoints *kapi.Endpoints) error {
+	p.endpoints = append(p.endpoints, endpoints)
+	return nil
+}
+func (p *fakeTestPlugin) HandleNamespaces(sets.String) error           { return nil }
+func (p *fakeTestPlugin) HandleNode(watch.EventType, *kapi.Node) error { return nil }
+func (p *fakeTestPlugin) Commit() error                                { return nil }
+
+type fakeTestRecorder struct {
+	rejections []string
+}
+
+func (r *fakeTestRecorder) RecordRouteRejection(route *routev1.Route, reason, message string) {
+	r.rejections = append(r.rejections, fmt.Sprintf("%s: %s", reason, message))
+}
+func (r *fakeTestRecorder) RecordRouteUpdate(route *routev1.Route, reason, message string) {}
+func (r *fakeTestRecorder) RecordRouteUnservableInFutureVersions(route *routev1.Route, reason, message string) {
+}
+func (r *fakeTestRecorder) RecordRouteUnservableInFutureVersionsClear(route *routev1.Route) {}
+
+func Test_checkRestrictedIP(t *testing.T) {
+	tests := []struct {
+		name        string
+		ip          string
+		expectError bool
+	}{
+		{
+			name:        "valid public IPv4",
+			ip:          "1.2.3.4",
+			expectError: false,
+		},
+		{
+			name:        "valid private IPv4",
+			ip:          "10.0.0.1",
+			expectError: false,
+		},
+		{
+			name:        "loopback IPv4",
+			ip:          "127.0.0.1",
+			expectError: true,
+		},
+		{
+			name:        "loopback IPv4 inside 127.0.0.0/8",
+			ip:          "127.127.127.222",
+			expectError: true,
+		},
+		{
+			name:        "loopback IPv6",
+			ip:          "::1",
+			expectError: true,
+		},
+		{
+			name:        "link-local IPv4 metadata",
+			ip:          "169.254.169.254",
+			expectError: true,
+		},
+		{
+			name:        "link-local IPv4 other",
+			ip:          "169.254.1.1",
+			expectError: true,
+		},
+		{
+			name:        "Azure metadata IP",
+			ip:          "168.63.129.16",
+			expectError: true,
+		},
+		{
+			name:        "valid IPv6",
+			ip:          "2001:db8::1",
+			expectError: false,
+		},
+		{
+			name:        "link-local IPv6",
+			ip:          "fe80::1",
+			expectError: true,
+		},
+		{
+			name:        "unspecified IPv4",
+			ip:          "0.0.0.0",
+			expectError: true,
+		},
+		{
+			name:        "unspecified IPv6",
+			ip:          "::",
+			expectError: true,
+		},
+		{
+			name:        "AWS IPv6 IMDS",
+			ip:          "fd00:ec2::254",
+			expectError: true,
+		},
+		{
+			name:        "link-local multicast IPv6",
+			ip:          "ff02::1",
+			expectError: true,
+		},
+		{
+			name:        "link-local multicast IPv4",
+			ip:          "224.0.0.1",
+			expectError: true,
+		},
+		{
+			name:        "non-link-local multicast IPv4",
+			ip:          "239.255.255.250",
+			expectError: true,
+		},
+		{
+			name:        "IPv4-mapped IPv6 loopback",
+			ip:          "::ffff:127.0.0.1",
+			expectError: true,
+		},
+		{
+			name:        "IPv4-mapped IPv6 metadata link-local",
+			ip:          "::ffff:169.254.169.254",
+			expectError: true,
+		},
+		{
+			name:        "IPv4-compatible IPv6 metadata (canonical)",
+			ip:          "::a9fe:a9fe",
+			expectError: true,
+		},
+		{
+			name:        "IPv4-compatible IPv6 metadata (dotted)",
+			ip:          "::169.254.169.254",
+			expectError: true,
+		},
+		{
+			name:        "IPv4-compatible IPv6 loopback",
+			ip:          "::127.0.0.1",
+			expectError: true,
+		},
+		{
+			name:        "IPv4-compatible IPv6 loopback (hex)",
+			ip:          "::7f00:1",
+			expectError: true,
+		},
+		{
+			name:        "loopback IPv4 from within 127/8",
+			ip:          "127.172.127.172",
+			expectError: true,
+		},
+		{
+			name:        "IPv4-compatible IPv6 loopback from within 127/8 (dotted)",
+			ip:          "::127.172.127.172",
+			expectError: true,
+		},
+		{
+			name:        "IPv4-compatible IPv6 loopback from within 127/8 (hex)",
+			ip:          "::7fac:7fac",
+			expectError: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			addr, err := netip.ParseAddr(tc.ip)
+			require.NoErrorf(t, err, "failed to parse IP address %s", tc.ip)
+			err = checkRestrictedIP(addr)
+			if tc.expectError {
+				require.Error(t, err)
+			}
+			if !tc.expectError {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func Test_validateEndpointAddress(t *testing.T) {
+	tests := []struct {
+		name        string
+		address     string
+		expectError bool
+	}{
+		{
+			name:        "valid private network IPv4 address",
+			address:     "10.0.0.1",
+			expectError: false,
+		},
+		{
+			name:        "valid private network IPv4 address",
+			address:     "23.206.60.92",
+			expectError: false,
+		},
+		{
+			name:        "valid IPv6",
+			address:     "2001:db8::1",
+			expectError: false,
+		},
+		{
+			name:        "restricted loopback IP",
+			address:     "127.0.0.1",
+			expectError: true,
+		},
+		{
+			name:        "restricted loopback IP from within the range",
+			address:     "127.0.127.1",
+			expectError: true,
+		}, {
+			name:        "restricted link-local IP",
+			address:     "169.254.169.254",
+			expectError: true,
+		},
+		{
+			name:        "restricted Azure metadata IP",
+			address:     "168.63.129.16",
+			expectError: true,
+		},
+		{
+			name:        "unspecified IPv4",
+			address:     "0.0.0.0",
+			expectError: true,
+		},
+		{
+			name:        "non-IP address rejected",
+			address:     "evil.example.com",
+			expectError: true,
+		},
+		{
+			name:        "empty string rejected",
+			address:     "",
+			expectError: true,
+		},
+		{
+			name:        "IPv4-mapped IPv6 loopback",
+			address:     "::ffff:127.0.0.1",
+			expectError: true,
+		},
+		{
+			name:        "IPv4-mapped IPv6 metadata link-local",
+			address:     "::ffff:169.254.169.254",
+			expectError: true,
+		},
+		{
+			name:        "IPv4-compatible IPv6 metadata (canonical)",
+			address:     "::a9fe:a9fe",
+			expectError: true,
+		},
+		{
+			name:        "IPv4-compatible IPv6 loopback",
+			address:     "::127.0.0.1",
+			expectError: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			err := validateEndpointAddress(tc.address)
+			if tc.expectError {
+				require.Error(t, err)
+			}
+			if !tc.expectError {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestExtendedValidator_HandleEndpoints(t *testing.T) {
+	tests := []struct {
+		name              string
+		endpoints         *kapi.Endpoints
+		expectedEndpoints *kapi.Endpoints
+	}{
+		{
+			name: "valid IP passes through",
+			endpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						Addresses: []kapi.EndpointAddress{{IP: "1.2.3.4"}},
+					},
+				},
+			},
+			expectedEndpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						Addresses: []kapi.EndpointAddress{{IP: "1.2.3.4"}},
+					},
+				},
+			},
+		},
+		{
+			name: "restricted link-local IP filtered out",
+			endpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						Addresses: []kapi.EndpointAddress{{IP: "169.254.169.254"}},
+					},
+				},
+			},
+			expectedEndpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						Addresses: []kapi.EndpointAddress{},
+					},
+				},
+			},
+		},
+		{
+			name: "restricted loopback IP filtered out",
+			endpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						Addresses: []kapi.EndpointAddress{{IP: "127.0.0.1"}},
+					},
+				},
+			},
+			expectedEndpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						Addresses: []kapi.EndpointAddress{},
+					},
+				},
+			},
+		},
+		{
+			name: "restricted IP in NotReadyAddresses filtered out",
+			endpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						NotReadyAddresses: []kapi.EndpointAddress{{IP: "169.254.169.254"}},
+					},
+				},
+			},
+			expectedEndpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						NotReadyAddresses: []kapi.EndpointAddress{},
+					},
+				},
+			},
+		},
+		{
+			name: "valid IP in NotReadyAddresses passes through",
+			endpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						NotReadyAddresses: []kapi.EndpointAddress{{IP: "10.0.0.5"}},
+					},
+				},
+			},
+			expectedEndpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						NotReadyAddresses: []kapi.EndpointAddress{{IP: "10.0.0.5"}},
+					},
+				},
+			},
+		},
+		{
+			name: "mixed valid and restricted keeps only valid",
+			endpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						Addresses: []kapi.EndpointAddress{{IP: "10.0.0.1"}},
+					},
+					{
+						Addresses: []kapi.EndpointAddress{{IP: "127.0.0.1"}},
+					},
+				},
+			},
+			expectedEndpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						Addresses: []kapi.EndpointAddress{{IP: "10.0.0.1"}},
+					},
+					{
+						Addresses: []kapi.EndpointAddress{},
+					},
+				},
+			},
+		},
+		{
+			name: "Azure metadata IP filtered out",
+			endpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						Addresses: []kapi.EndpointAddress{{IP: "168.63.129.16"}},
+					},
+				},
+			},
+			expectedEndpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						Addresses: []kapi.EndpointAddress{},
+					},
+				},
+			},
+		},
+		{
+			name: "mixed valid and restricted in same subset",
+			endpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						Addresses: []kapi.EndpointAddress{
+							{IP: "10.0.0.1"},
+							{IP: "127.0.0.1"},
+							{IP: "10.0.0.2"},
+						},
+					},
+				},
+			},
+			expectedEndpoints: &kapi.Endpoints{
+				Subsets: []kapi.EndpointSubset{
+					{
+						Addresses: []kapi.EndpointAddress{
+							{IP: "10.0.0.1"},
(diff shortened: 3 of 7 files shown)


pkg/router/controller/extended_validator.go
```diff
@@ -40,9 +41,69 @@ func (p *ExtendedValidator) HandleNode(eventType watch.EventType, node *kapi.Nod
...
+		// We'll allow transient resolution errors to just pass through, as the router itself
+		// would fail to resolve it if it's truly unresolvable. But if it's an SSRF, we MUST block.
+		// Actually, HAProxy resolves it at runtime. If we fail to resolve, HAProxy might succeed later.
+		// For CVE-2026-42965, the researcher's FQDN resolves. If they use a malicious DNS server that
+		// returns a timeout to us but 169.254 to HAProxy, they bypass this.
+		// To be safe, if we can't resolve an FQDN, we should probably fail. But let's check
+		// what existing validation does. Since there is none, let's just do best-effort or fail closed.
+		// For now, let's fail if we can't verify, or maybe just log?
+		// Actually, standard practice for SSRF is to fail closed if unresolvable.
+		return fmt.Errorf("failed to resolve FQDN %q: %v", address, err)
```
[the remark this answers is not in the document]
bentito: Good catch. I've removed the verbose deliberation and added a concise note explicitly calling out the TOCTOU limitation, along with a `TODO` to investigate mitigation via HAProxy configuration in the future.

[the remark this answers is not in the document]
bentito: Agreed. I have refactored `validateEndpointAddress` to take a `context.Context` and updated the implementation to use `net.DefaultResolver.LookupIPAddr` with a 5-second timeout. This prevents potential hangs from malicious DNS servers.

pkg/router/controller/extended_validator.go
```diff
@@ -40,9 +41,69 @@ func (p *ExtendedValidator) HandleNode(eventType watch.EventType, node *kapi.Nod
...
+func checkRestrictedIP(ip net.IP) error {
+	// Block cloud metadata and localhost
+	if ip.Equal(net.ParseIP("169.254.169.254")) {
+		return fmt.Errorf("IP address %s is a restricted cloud metadata IP", ip.String())
+	}
+	if ip.IsLoopback() {
+		return fmt.Errorf("IP address %s is a restricted loopback IP", ip.String())
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
bentito: Excellent suggestion. I've updated the logic to use `ip.IsLinkLocalUnicast()`, which covers both the `169.254.0.0/16` and `fe80::/10` ranges, and I explicitly added a block for the Azure Wireserver IP.

pkg/router/controller/extended_validator.go
```diff
@@ -40,9 +43,70 @@ func (p *ExtendedValidator) HandleNode(eventType watch.EventType, node *kapi.Nod
...
+
+var (
+	azureMetadata = net.ParseIP("168.63.129.16")
+)
+
+func checkRestrictedIP(ip net.IP) error {
+	if ip.IsLoopback() {
+		return fmt.Errorf("IP address %s is a restricted loopback IP", ip.String())
+	}
+	if ip.IsLinkLocalUnicast() {
```
bentito: To clarify for anyone reviewing this diff against the original PR description: we replaced the explicit literal check for 169.254.169.254 with ip.IsLinkLocalUnicast()

pkg/router/controller/extended_validator.go
```diff
@@ -40,9 +43,70 @@ func (p *ExtendedValidator) HandleNode(eventType watch.EventType, node *kapi.Nod
...
+		return checkRestrictedIP(ip)
+	}
+
+	// If not a valid IP, assume it is an FQDN and resolve it
+	resolver := net.DefaultResolver
+	lookupCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+	addrs, err := resolver.LookupIPAddr(lookupCtx, address)
+	if err != nil {
+		// TOCTOU Limitation: An attacker can bypass this check by returning a benign IP now and
```
rikatz: This is valid but a very edge case IMO. The attacker would need to either:
* Control the DNS of the cluster, which brings a different problem
* Be able to rename the "invalid" domain to a valid domain on a good TTL window (doable)

I think long term we can consider something like `external-check command /usr/local/bin/check_ip_blacklist.sh` where HAProxy can call a binary running on the same container of HAProxy passing the hostname or IP and leaving this binary to decide if it is safe or not.

pkg/router/controller/extended_validator.go
```diff
@@ -40,9 +43,70 @@ func (p *ExtendedValidator) HandleNode(eventType watch.EventType, node *kapi.Nod
...
 // HandleEndpoints processes watch events on the Endpoints resource.
 func (p *ExtendedValidator) HandleEndpoints(eventType watch.EventType, endpoints *kapi.Endpoints) error {
+	ctx := context.TODO()
+	for _, subset := range endpoints.Subsets {
+		for _, addr := range subset.Addresses {
+			if err := validateEndpointAddress(ctx, addr.IP); err != nil {
+				log.Error(err, "skipping endpoints due to invalid configuration", "endpoints", fmt.Sprintf("%s/%s", endpoints.Namespace, endpoints.Name))
+				// We don't have a recordEndpointRejection mechanism, so we log and drop the event.
+				// This prevents the router from proxying to the malicious backend.
+				return fmt.Errorf("invalid endpoint configuration: %v", err)
```
rikatz: I am curious and can't remember, what happens when we return an error? Does this block any further configuration of the router, or does this block just the specific route? (I think it is just the route, but I cannot remember).

Also, would it make sense to aggregate all of the errors on a []err and return at once?

pkg/router/controller/resolver.go
```diff
@@ -0,0 +1,76 @@
...
+	resolver *net.Resolver
+}
+
+func NewDNSEndpointResolver() *DNSEndpointResolver {
+	return &DNSEndpointResolver{
+		resolver: net.DefaultResolver,
+	}
+}
+
+func (r *DNSEndpointResolver) ResolveEndpointAddress(ctx context.Context, hostname string) ([]net.IP, error) {
```
rikatz: flagging for myself: add a better comment on what these timeouts are, make the timeout constants with a proper explanation

pkg/router/controller/resolver.go
```diff
@@ -0,0 +1,76 @@
...
+// DNSEndpointResolver resolves hostnames using the system DNS
+// resolver with HAProxy-compatible timeout and retry settings:
+// 1 retry, 5s per attempt, 10s overall timeout.
+type DNSEndpointResolver struct {
+	resolver *net.Resolver
+}
+
+func NewDNSEndpointResolver() *DNSEndpointResolver {
+	return &DNSEndpointResolver{
+		resolver: net.DefaultResolver,
```
rikatz: Is there a better way to define timeouts and retries? Need to verify

pkg/router/controller/resolver.go
```diff
@@ -0,0 +1,76 @@
...
+	ips := make([]net.IP, 0, len(addrs))
+	for _, addr := range addrs {
+		ips = append(ips, addr.IP)
+	}
+	sortIPsIPv6First(ips)
+	return ips, nil
+}
+
+// sortIPsIPv6First sorts IPs with IPv6 addresses before IPv4.
+func sortIPsIPv6First(ips []net.IP) {
```
rikatz: ask Claude to make this reflect exactly the behavior of HAProxy

pkg/router/controller/resolver_test.go
```diff
@@ -0,0 +1,56 @@
...
+func TestSortIPsIPv6First(t *testing.T) {
+	tests := []struct {
+		name     string
+		ips      []net.IP
+		expected []string
+	}{
+		{
+			name:     "IPv6 before IPv4",
+			ips:      []net.IP{net.ParseIP("10.0.0.1"), net.ParseIP("2001:db8::1")},
+			expected: []string{"2001:db8::1", "10.0.0.1"},
```
rikatz: what happens when we render IPv6? Will HAProxy break? Is this the right semantic? We probably need a better test here, or even some e2e test or some validation.

pkg/router/controller/router_controller.go
```diff
@@ -255,6 +262,41 @@ func (c *RouterController) HandleEndpointSlice(eventType watch.EventType, objMet
 	c.HandleEndpoints(eventType, endpoints)
 }
 
+// resolveEndpointAddresses resolves FQDN addresses in endpoint
```
rikatz: this seems like a flaw. needs to be verified. We must focus on any endpointslice of type FQDN, I need to check inside the conversion again if the endpointslice of type FQDN is converted to the IP field, but without validating it is an IP

rikatz: some checking confirmed this is correct. `ConvertEndpointSlice` will blindly convert EndpointSlices to EndpointSubset, without checking the Endpoint type and moving the Addresses to an IP.

So per the code here we are:
* resolving the address list, when it is not an IP type (resolveEndpointAddress comes first)
* Then call handleEndpoints that will filter out invalid addresses

pkg/router/controller/resolver.go
rikatz: done

pkg/router/controller/resolver.go
rikatz: fixed

pkg/router/controller/resolver.go
rikatz: fixed

pkg/router/controller/resolver_test.go
rikatz: the internal plugin is able to convert IPv6 to bracket address, this is fine

rikatz:
@coderabbitai rise and shine. I want you to give me reasons not to merge this PR. Find broken behavior. Find concerning issues. If you don't find any concerning issue, we are on a happy path here, otherwise let me know what should be fixed

rikatz:
@coderabbitai for issue 1, we do not set the FQDN anymore on the HAProxy configuration, but rather the address. The current approach specifically stops passing FQDNs, resolves the addresses locally and sets only endpoints for HAProxy, so is this real a concern or did you missed it?

rikatz:
Is 3 really concerning? If a DNS is not available, it doesn´t mean the operator won´t start, but rather that EndpointSlices with FQDN and addresses will not be resolved. Maybe we can lower the timeouts, but this doesn't mean we are breaking here right?

rikatz:
@coderabbitai: 

Finding 2: Synchronous DNS Blocks the Informer Event Loop — Valid, but nuanced, a user with access to create EndpointSlices is already somehow very privileged. It is a known limitation, but not a critical one

Finding 3: this is by design, we decided to never opt-out of the validation

Finding 5: This is wrong. The EndpointSlice still exists in Kubernetes — the service is real, it just has no resolvable backends right now. Sending Deleted would remove the entire route/backend from HAProxy, and the next reconciliation when DNS recovers would need to re-create it.

Finding 6: seems a low priority and low risk, given we prefer to fail by default here

Finding 7: speculative, AWS just documents this endpoint. Maybe add to your learnings to always check on AWS documentation if they changed it

Finding 8: this is by design, we are separating concerns here

pkg/router/controller/extended_validator.go
```diff
@@ -39,21 +45,81 @@ func (p *ExtendedValidator) HandleNode(eventType watch.EventType, node *kapi.Nod
...
 
 // HandleEndpoints processes watch events on the Endpoints resource.
+// Addresses that fail IP validation are removed individually rather
+// than rejecting the entire endpoint set.
 func (p *ExtendedValidator) HandleEndpoints(eventType watch.EventType, endpoints *kapi.Endpoints) error {
-	return p.plugin.HandleEndpoints(eventType, endpoints)
+	nsName := fmt.Sprintf("%s/%s", endpoints.Namespace, endpoints.Name)
+
+	// Build filtered subsets without mutating the original, which
+	// may be an informer cache object on the legacy Endpoints path.
```
Miciah: No, `endpoints` does not come from an informer; it is allocated here: https://github.com/openshift/router/blob/fa5382c64be012047cc6832297eef714487015bc/pkg/router/controller/endpointsubset/converter.go#L8-L63

pkg/router/controller/extended_validator.go
```diff
@@ -39,21 +45,81 @@ func (p *ExtendedValidator) HandleNode(eventType watch.EventType, node *kapi.Nod
...
+		ep.Subsets[i].NotReadyAddresses = filterValidAddresses(subset.NotReadyAddresses, nsName)
+	}
+	return p.plugin.HandleEndpoints(eventType, ep)
+}
+
+func filterValidAddresses(addrs []kapi.EndpointAddress, nsName string) []kapi.EndpointAddress {
+	valid := make([]kapi.EndpointAddress, 0, len(addrs))
+	for _, addr := range addrs {
+		if err := validateEndpointAddress(addr.IP); err != nil {
+			log.Error(err, "Skipping endpoint address with restricted IP", "endpoints", nsName, "address", addr.IP)
```
Miciah: Is the `nsName` allocation necessary only for a log message that typically will not be printed anyway?  How about changing the log message so that `nsname` (and therefore the string allocation) is not needed?  You can print the name and namespace as separate fields in the log message.

pkg/router/controller/extended_validator.go
```diff
@@ -39,21 +45,81 @@ func (p *ExtendedValidator) HandleNode(eventType watch.EventType, node *kapi.Nod
...
+			continue
+		}
+		valid = append(valid, addr)
+	}
```
Miciah: Is it fair to say that the output will be the same as the input for every route in very nearly all clusters?  Would it be worthwhile to scan first to check whether filtering is necessary, and then allocate a new array for `valid` only if it turns out to be necessary?

pkg/router/controller/extended_validator.go
Miciah: This log message should say "restricted or invalid" per the logic in `validateEndpointAddress`: https://github.com/openshift/router/blob/fa5382c64be012047cc6832297eef714487015bc/pkg/router/controller/extended_validator.go#L76-L78.

pkg/router/controller/extended_validator.go
```diff
@@ -39,21 +45,81 @@ func (p *ExtendedValidator) HandleNode(eventType watch.EventType, node *kapi.Nod
...
+	return valid
+}
+
+func validateEndpointAddress(address string) error {
+	ip := net.ParseIP(address)
+	if ip == nil {
+		return fmt.Errorf("address %q is not a valid IP", address)
```
Miciah:
```suggestion
		return fmt.Errorf("address %q is not a valid IP address", address)
```

pkg/router/controller/extended_validator.go
```diff
@@ -39,21 +45,81 @@ func (p *ExtendedValidator) HandleNode(eventType watch.EventType, node *kapi.Nod
...
+	}
+	return checkRestrictedIP(ip)
+}
+
+var (
+	azureMetadata = net.ParseIP("168.63.129.16")
+	awsIPv6IMDS   = net.ParseIP("fd00:ec2::254")
+)
```
Miciah, line 90: Is it worth using a set instead?  Of course, you would need to convert the `net.IP` value back to a `string` value.

pkg/router/controller/extended_validator.go
Miciah: It would be helpful to add code comments here, and also here: https://github.com/openshift/router/blob/fa5382c64be012047cc6832297eef714487015bc/pkg/router/controller/extended_validator.go#L83-L86

Something like this:
```go
	// The AWS IMDS endpoint 169.254.169.254 is covered by the ip.IsLinkLocalUnicast()
	// check in `checkRestrictedIP`.
```

pkg/cmd/infra/router/template.go
```diff
@@ -814,9 +814,7 @@ func (o *TemplateRouterOptions) Run(stopCh <-chan struct{}) error {
 	if o.UpgradeValidation {
 		plugin = controller.NewUpgradeValidation(plugin, recorder, o.UpgradeValidationForceAddCondition, o.UpgradeValidationForceRemoveCondition)
 	}
-	if o.ExtendedValidation {
-		plugin = controller.NewExtendedValidator(plugin, recorder)
-	}
+	plugin = controller.NewExtendedValidator(plugin, recorder, o.ExtendedValidation)
```
Miciah, line 830: Out of curiosity, did you consider making the endpoints validation a separate plugin, or implementing it in `createRouterEndpoints`?

pkg/router/controller/extended_validator_test.go
```diff
@@ -0,0 +1,368 @@
...
+		})
+	}
+}
+
+func TestExtendedValidator_HandleEndpoints(t *testing.T) {
+	tests := []struct {
+		name              string
+		endpoints         *kapi.Endpoints
+		expectedAddrCount int
+		expectedNRCount   int
```
Miciah: Any reason to check the count and not the actual addresses?

pkg/router/controller/extended_validator_test.go
```diff
@@ -0,0 +1,368 @@
...
+
+func (r *fakeTestRecorder) RecordRouteRejection(route *routev1.Route, reason, message string) {
+	r.rejections = append(r.rejections, fmt.Sprintf("%s: %s", reason, message))
+}
+func (r *fakeTestRecorder) RecordRouteUpdate(route *routev1.Route, reason, message string) {}
+func (r *fakeTestRecorder) RecordRouteUnservableInFutureVersions(route *routev1.Route, reason, message string) {
+}
+func (r *fakeTestRecorder) RecordRouteUnservableInFutureVersionsClear(route *routev1.Route) {}
+
+func TestCheckRestrictedIP(t *testing.T) {
```
Miciah:
```suggestion
func Test_checkRestrictedIP(t *testing.T) {
```

pkg/router/controller/extended_validator_test.go
```diff
@@ -0,0 +1,368 @@
...
+				t.Errorf("expected error for IP %s, got nil", tc.ip)
+			}
+			if !tc.expectError && err != nil {
+				t.Errorf("expected no error for IP %s, got %v", tc.ip, err)
+			}
+		})
+	}
+}
+
+func TestValidateEndpointAddress(t *testing.T) {
```
Miciah:
```suggestion
func Test_validateEndpointAddress(t *testing.T) {
```

pkg/router/controller/extended_validator_test.go
```diff
@@ -0,0 +1,368 @@
...
+		{
+			name:        "valid IPv6",
+			address:     "2001:db8::1",
+			expectError: false,
+		},
+		{
+			name:        "restricted loopback IP",
+			address:     "127.0.0.1",
+			expectError: true,
+		},
```
Miciah, line 218: Please add a test case for another loopback address.

pkg/router/controller/extended_validator_test.go
```diff
@@ -0,0 +1,368 @@
...
+func TestValidateEndpointAddress(t *testing.T) {
+	tests := []struct {
+		name        string
+		address     string
+		expectError bool
+	}{
+		{
+			name:        "valid public IPv4",
+			address:     "10.0.0.1",
+			expectError: false,
```
Miciah, line 202: Calling 10.0.0.1 a "public" address is confusing.  Maybe it makes sense in the context of cloud networking and distinguishing between infrastructure and non-infrastructure?  Still, I'd suggest saying "private network IPv4 address" or using some different term entirely to avoid confusion.  

Also, it might make sense to add a test case with a public IPv4 address.

pkg/router/controller/resolver.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return nil, fmt.Errorf("resolving %q: %w", hostname, lastErr)
+}
+
+func isRetryableDNSError(err error) bool {
+	var dnsErr *net.DNSError
+	if !errors.As(err, &dnsErr) {
+		return false
+	}
+	return dnsErr.IsTimeout && !dnsErr.IsNotFound
```
Miciah: This condition looks suspicious.  Can `dnsErr.IsTimeout` and `dnsErr.IsNotFound` both be true?  Should we check `dnsErr.IsTimeout || dnsErr.IsTemporary` instead?

pkg/router/controller/router_controller.go
```diff
@@ -242,6 +244,11 @@ func (c *RouterController) HandleEndpointSlice(eventType watch.EventType, objMet
 		Subsets: endpointsubset.ConvertEndpointSlice(items, endpointsubset.DefaultEndpointAddressOrderByFuncs(), endpointsubset.DefaultEndpointPortOrderByFuncs()),
 	}
 
+	nsName := fmt.Sprintf("%s/%s", objMeta.Namespace, objMeta.Name)
+	if c.Resolver != nil {
+		c.resolveEndpointAddresses(endpoints, nsName)
+	}
```
Miciah: Related to my question about adding validation to the existing plugin, did you consider adding the name resolution to `createRouterEndpoints`?  I figured that we would have it both resolve and validate addresses.  I see now that you separated those concerns between the router controller and the extended validation, and maybe it is better design to have those concerns separated, but I'm curious about your design choices.

pkg/router/controller/router_controller_test.go
```diff
@@ -0,0 +1,323 @@
...
+			items: []discoveryv1.EndpointSlice{{
+				ObjectMeta:  metav1.ObjectMeta{Name: "slice-1", Namespace: "ns"},
+				AddressType: ipv4Type,
+				Endpoints: []discoveryv1.Endpoint{{
+					Addresses: []string{"10.0.0.1"},
+				}},
+			}},
+			expectedEventType: watch.Modified,
+			expectedAddrs:     []string{"10.0.0.1"},
+		},
```
Miciah: If the test case omits `resolver`, does the test actually cover what it is intended to cover?  It seems that the controller doesn't actually attempt name resolution if it doesn't have a resolver: https://github.com/openshift/router/blob/fa5382c64be012047cc6832297eef714487015bc/pkg/router/controller/router_controller.go#L248-L249

pkg/router/controller/router_controller_test.go
```diff
@@ -0,0 +1,323 @@
...
+			},
+			items: []discoveryv1.EndpointSlice{{
+				ObjectMeta:  metav1.ObjectMeta{Name: "slice-1", Namespace: "ns"},
+				AddressType: fqdnType,
+				Endpoints: []discoveryv1.Endpoint{{
+					Addresses: []string{"multi.example.com"},
+				}},
+			}},
+			expectedEventType: watch.Modified,
+			expectedAddrs:     []string{"10.0.0.1", "2001:db8::1"},
```
Miciah: I thought the intention was that IPv6 addresses would come first?

pkg/router/controller/router_controller.go
```diff
@@ -255,6 +262,44 @@ func (c *RouterController) HandleEndpointSlice(eventType watch.EventType, objMet
...
+func (c *RouterController) resolveEndpointAddresses(endpoints *kapi.Endpoints, nsName string) {
+	ctx, cancel := context.WithTimeout(context.Background(), endpointResolutionTimeout)
+	defer cancel()
+
+	for i := range endpoints.Subsets {
+		endpoints.Subsets[i].Addresses = c.resolveAddressList(ctx, endpoints.Subsets[i].Addresses, nsName)
+		endpoints.Subsets[i].NotReadyAddresses = c.resolveAddressList(ctx, endpoints.Subsets[i].NotReadyAddresses, nsName)
+
+		endpointsubset.SortAddresses(endpoints.Subsets[i].Addresses, endpointsubset.DefaultEndpointAddressOrderByFuncs())
+		endpointsubset.SortAddresses(endpoints.Subsets[i].NotReadyAddresses, endpointsubset.DefaultEndpointAddressOrderByFuncs())
```
Miciah: `resolveAddressList` sorts IPv6 addresses first, but then `endpointsubset.SortAddresses` puts IPv4 addresses first — right?

Miciah: Related question:  Is the correct behavior to sort IPv6 addresses before IPv4 addresses, or is the correct behavior to drop IPv4 addresses if the the name resolves to both IPv6 addresses and IPv4 addresses?

pkg/cmd/infra/router/template.go
rikatz: nope, mostly claude took this approach and I though was fine

pkg/router/controller/extended_validator.go
rikatz: removed this leading comment, didn't made sense anyway

pkg/router/controller/extended_validator.go
rikatz: removed.

rikatz: it does, below on checkRestrictedIP. In this specific case the message here is right: this is not a valid IP.

the checkRestrictedIP will return messages saying it is restricted

pkg/router/controller/router_controller_test.go
```diff
@@ -0,0 +1,323 @@
...
+func TestHandleEndpointSlice_FQDNResolution(t *testing.T) {
+	ipv4Type := discoveryv1.AddressTypeIPv4
+	ipv6Type := discoveryv1.AddressTypeIPv6
+	fqdnType := discoveryv1.AddressTypeFQDN
+
+	tests := []struct {
+		name              string
+		resolver          EndpointResolver
+		items             []discoveryv1.EndpointSlice
+		expectedEventType watch.EventType
```
Miciah: Does it make sense to have `expectedEventType` as a test input, or should the test runner unconditionally assert that it is `watch.Modified`?

pkg/router/controller/router_controller_test.go
```diff
@@ -0,0 +1,323 @@
...
+			}
+		})
+	}
+}
+
+// TestFQDNToRestrictedIP_DefenseInDepth verifies that when a FQDN
+// resolves to a restricted IP, the address is resolved by the
+// RouterController but then filtered by the ExtendedValidator before
+// reaching the inner plugin.
+func TestFQDNToRestrictedIP_DefenseInDepth(t *testing.T) {
```
Miciah: Why is this test named "DefenseInDepth"?  Wouldn't that imply that we had two checks for restricted addresses rather than the one check that we actually have?

Miciah:
It looks like there has been some drift between the commit message and the current changes.  The commit message should be updated:

> This commit updates the Endpoints validation and addition, first
> trying to resolve any non-IP endpoint address to a proper address,
> and then validating against a forbidden list of IPs on Endpoints,
> like multicast, link local, localhost and cloud metadata server

- The router controller now resolves addresses, so this commit has two main parts: (1) updates to the extended validation and (2) updates to the router controller logic.  
- Is "addition" a typo for "admission", or do you mean it in the sense of handling `watch.Added` (and also `watch.Modified`) events?  A more general term like "handling" might be less confusing.
- What do you mean by "proper address"?  Since the resolution and validation are now separate concerns, maybe it is more accurate just to say the controller resolves a hostname to "an _IP_ address" (and the validation makes sure that it is a "proper"/valid/allowed address).  
- I'm not sure what the phrase "on Endpoints" is intended to convey.  Can it be omitted?
- You are missing a period at the end of the sentence (or at least I can only assume it is the end of the sentence!).

pkg/router/controller/extended_validator.go
rikatz: ah the log message, fixed

pkg/router/controller/extended_validator.go
rikatz: fixed

pkg/router/controller/extended_validator.go
rikatz: what is the advantage? I mean, it is a 2 line check, I think we can make it not a set for now and make a set in case we get more IPs

pkg/router/controller/extended_validator.go
Miciah: `filterValidAddresses` will log a message that says "restricted" with the error message that says not "valid" appended, right?  

Edit: Disregard, I didn't see your second reply when I made this comment.

pkg/router/controller/extended_validator.go
Miciah: Are you removing the deep-copy too?  Better to avoid doing a (large) allocation for every single endpoints if we can avoid it.

pkg/router/controller/extended_validator.go
rikatz: fixed

pkg/router/controller/extended_validator_test.go
rikatz: fixed

pkg/router/controller/extended_validator_test.go
rikatz: fixed

pkg/router/controller/extended_validator_test.go
rikatz: fixed

pkg/router/controller/extended_validator_test.go
rikatz: added

pkg/router/controller/extended_validator_test.go
rikatz: no reason, added inner subset assertion/equality

pkg/router/controller/extended_validator.go
Miciah: Fair enough.  No need to use a set if you prefer the current approach.

pkg/router/controller/extended_validator.go
rikatz: nope 🤦 removing now

pkg/router/controller/resolver.go
rikatz: indeed, I have fixed here as suggested

pkg/router/controller/router_controller.go
rikatz: original approach was from @bentito and I agreed that the current design is good enough. It is part of extendedValidation but yet always executed (it is not behind the route validation flag)

pkg/router/controller/router_controller.go
rikatz: from my understanding, HAProxy prioritizes IPv6 and then IPv4 but does not drop it, so I have followed the same approach

rikatz: apparently and per unit tests I am wrong, let me check and fix here

pkg/router/controller/router_controller_test.go
rikatz: yes, I have removed this assertion and added an assertion for always being watch.Modified

pkg/router/controller/router_controller_test.go
rikatz: so, this is something that while I was doing self review got me but I was "meh", but yes, we should never allow null resolver. I will add it to my fix list

pkg/router/controller/router_controller_test.go
rikatz: because the validation is deep on the code, so the test here is more like a broader unit test that verifies the whole workflow. It is not calling the validation functions, but rather the whole ExtendedValidation from route controller. I can change for a better name if you prefer

pkg/router/controller/router_controller.go
rikatz: fixed

pkg/router/controller/router_controller_test.go
rikatz: added a default resolver always even if it is null, and removed this check

pkg/router/controller/router_controller_test.go
rikatz: yes, fixed in a next commit

pkg/router/controller/extended_validator.go
rikatz: this will make the code much more complex than it should IMO. Given this is called per EndpointSlice event, and the slices are not huge, I would say that we don't need to optimize it now

Miciah: Would it?  All I mean is something like this:
```go
	// Return quickly without any allocs if nothing needs to be filtered.
	if !slices.ContainsFunc(addrs, func(addr kapi.EndpointAddress) bool {
		return validateEndpointAddress(addr.IP) != nil
	}) {
		return addrs
	}

	// Proceed with the current filter logic.
	valid := make([]kapi.EndpointAddress, 0, len(addrs))
	for _, addr := range addrs {
		if err := validateEndpointAddress(addr.IP); err != nil {
			log.Error(err, "Skipping endpoint address with restricted or invalid IP", "address", addr.IP)
			continue
		}
		valid = append(valid, addr)
	}
	return valid
}
```
(This will make your unit tests fail because they sometimes have nil `Subsets.Addresses` or nil `Subsets.NotReadyAddresses`, and the `return addrs` will result in an empty `[]EndpointAddress{}` value instead of `nil`, but that is an easily fixed defect of the unit tests.)

pkg/router/controller/router_controller.go
Miciah: > from my understanding, HAProxy prioritizes IPv6 and then IPv4 but does not drop it, so I have followed the same approach

I mean, if the controller still adds `server` lines for IPv4 addresses in the HAProxy configuration file, then HAProxy will still use them, right?  Even if the IPv6 lines come first, HAProxy will use the IPv4 addresses when using roundrobin balancing, or with leastconn or random if the number of concurrent connections is greater than the number of IPv6 addresses.   In contrast — given the description of the `httpclient.resolvers.prefer` option:
```text
httpclient.resolvers.prefer <ipv4|ipv6>
  This option allows to chose which family of IP you want when resolving,
  which is convenient when IPv6 is not available on your network. Default
  option is "ipv6".
```
— this means that HAProxy _only_ uses the IPv6 addresses if you give it a `server` line that has a DNS name that resolves to a mix of IPv4 and IPv6 addresses, right?

pkg/router/controller/router_controller_test.go
Miciah: > because the validation is deep on the code, so the test here is more like a broader unit test that verifies the whole workflow. It is not calling the validation functions, but rather the whole ExtendedValidation from route controller.

The phrase "defense in depth" refers to having redundant security layers — for example, using SELinux labels _plus_ Linux namespaces to contain processes, or using NetworkPolicy _plus_ TLS to restrict access to network endpoints — so if one layer fails, the next layer provides protection.

> I can change for a better name if you prefer

It isn't very creative, but maybe `TestRouterController_EndpointsValidation`?  The name isn't terribly important though; I find the current name a little confusing, but it isn't worth spending much time on if you are happy with it.

pkg/router/controller/extended_validator_test.go
```diff
@@ -123,28 +124,26 @@ func TestCheckRestrictedIP(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			ip := net.ParseIP(tc.ip)
-			if ip == nil {
-				t.Fatalf("failed to parse IP %q", tc.ip)
-			}
+			require.NotNil(t, ip, "failed to parse IP")
```
Miciah:
```suggestion
			require.NotNil(t, ip, "failed to parse IP address %s", tc.ip)
```

pkg/router/controller/extended_validator_test.go
```diff
@@ -123,28 +124,26 @@ func TestCheckRestrictedIP(t *testing.T) {
...
 			err := checkRestrictedIP(ip)
 			if tc.expectError && err == nil {
-				t.Errorf("expected error for IP %s, got nil", tc.ip)
+				require.Error(t, err)
```
Miciah: ` && err == nil` is redundant.
```suggestion
			if tc.expectError {
				require.Error(t, err, "expected error for IP %s, got nil", tc.ip)
```

pkg/router/controller/extended_validator_test.go
```diff
@@ -123,28 +124,26 @@ func TestCheckRestrictedIP(t *testing.T) {
...
 			}
 			if !tc.expectError && err != nil {
-				t.Errorf("expected no error for IP %s, got %v", tc.ip, err)
+				require.NoError(t, err)
```
Miciah:
```suggestion
			if !tc.expectError {
				require.NoError(t, err)
```

pkg/router/controller/extended_validator_test.go
```diff
@@ -189,10 +192,10 @@ func TestValidateEndpointAddress(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			err := validateEndpointAddress(tc.address)
 			if tc.expectError && err == nil {
-				t.Errorf("expected error for address %q, got nil", tc.address)
+				require.Error(t, err)
```
Miciah:
```suggestion
			if tc.expectError {
				require.Error(t, err, "expected error for address %q, got nil", tc.address)
```

pkg/router/controller/extended_validator_test.go
```diff
@@ -189,10 +192,10 @@ func TestValidateEndpointAddress(t *testing.T) {
...
 			}
 			if !tc.expectError && err != nil {
-				t.Errorf("expected no error for address %q, got %v", tc.address, err)
+				require.NoError(t, err)
```
Miciah:
```suggestion
			if !tc.expectError {
				require.NoError(t, err)
```

pkg/router/controller/extended_validator_test.go
Miciah: I see that "public" was changed, but I don't see a new test case with an actual public address.  It was just a suggestion, but maybe it was overlooked.

pkg/router/controller/router_controller.go
```diff
@@ -271,9 +273,6 @@ func (c *RouterController) resolveEndpointAddresses(endpoints *kapi.Endpoints, n
 	for i := range endpoints.Subsets {
 		endpoints.Subsets[i].Addresses = c.resolveAddressList(ctx, endpoints.Subsets[i].Addresses, nsName)
 		endpoints.Subsets[i].NotReadyAddresses = c.resolveAddressList(ctx, endpoints.Subsets[i].NotReadyAddresses, nsName)
-
-		endpointsubset.SortAddresses(endpoints.Subsets[i].Addresses, endpointsubset.DefaultEndpointAddressOrderByFuncs())
-		endpointsubset.SortAddresses(endpoints.Subsets[i].NotReadyAddresses, endpointsubset.DefaultEndpointAddressOrderByFuncs())
```
Miciah: I'm not sure that removing `SortAddresses` entirely is the right solution.  As <https://github.com/openshift/enhancements/blob/master/enhancements/ingress/power-of-two-random-choices.md#alternatives-not-implemented> notes, deterministic ordering has advantages.  In particular, the source balancing algorithm depends on server ordering consistency to preserve session affinity across configuration updates and among router pods belonging to the same deployment.  

How difficult would it be both to sort using the default sort funcs _and_ to sort IPv6 addresses first?  I think you could straightforwardly convert the logic in `sortIPsIPv6First` into an `EndpointAddressLessFunc` value and pass that as an additional func to `SortAddresses`.  Then `DNSEndpointResolver` itself wouldn't need any sorting logic.

pkg/router/controller/extended_validator_test.go
rikatz: Changed to use also NotNilf

pkg/router/controller/extended_validator_test.go
rikatz: yeah, I think my brain was melted when I wrote this on friday...

pkg/router/controller/extended_validator_test.go
rikatz: fixed

pkg/router/controller/extended_validator_test.go
rikatz: yeah I've read the first paragraph, not the second. Fixed here

pkg/router/controller/extended_validator_test.go
rikatz: fixed

pkg/router/controller/extended_validator_test.go
rikatz: fixed

pkg/router/controller/router_controller_test.go
rikatz: no, sounds good, I am already making changes and changing this name is fine

pkg/router/controller/extended_validator.go
rikatz: I can buy it, considering that 99% of the clusters will not have bad addresses or bad FQDNs. The 1% concerns me because we are now running the validation twice.

Maybe I can join your approach with DeleteFunc:

```
return slices.DeleteFunc(addrs, func(addr kapi.EndpointAddress) bool {
	err := validateEndpointAddress(addr.IP); err != nil {
          log.Error(err, "Skipping endpoint address with restricted or invalid IP", "address", addr.IP)
    }
    return err != nil
}
```

rikatz: fixed, the approach with DeleteFunc is the clean one. It will run validation on every IP just once, and reduce the array removing invalid addresses without the need of a new allocation.

gcs278:
I've started to review this. 
/assign

gcs278:
I've take a first glance at this. Nice work, it's solidly written and through through. But, I feel quite conflicted about it. I understand why you have arrived at the conclusion that we need to move DNS resolution into the router, but it feels risky as the probability of breaking a customer using ExternalName DNS resolution feels quite high to me.

Here are my initial concerns:

- **DNS freshness degrades from TTL-driven to ~30min.** HAProxy re-resolved when the DNS record's TTL expired — could be seconds for short-TTL records used in failover scenarios. Now the router resolves once and hands HAProxy a static IP. Re-resolution only happens on the next informer resync (30 minutes). Customers relying on short-TTL DNS for fast failover will see stale backends.

- **Failure handling is worse.** Before: HAProxy marked the server DOWN, kept the config line, retried DNS [every 30s](https://github.com/haproxy/haproxy/blob/master/src/resolvers.c#L3646-L3652) on failure, auto-recovered when DNS came back. After: the address is silently dropped from the config, no server line exists, no retry until next informer resync (30 minutes). A transient DNS blip during a resync can remove a working backend for 30 minutes with no user-visible signal except router pod logs.

- **Switching from HAProxy's C resolver to Go's net.Resolver is risky.** We've been burned by DNS resolver differences before — [OCPBUGS-27397](https://issues.redhat.com/browse/OCPBUGS-27397) (SRV lookups broke when CoreDNS changed EDNS behavior), [OCPBUGS-240](https://issues.redhat.com/browse/OCPBUGS-240) (UDP payload size change from 512 to 2048 broke customer F5 firewalls), [OCPBUGS-35104](https://issues.redhat.com/browse/OCPBUGS-35104) (TCP protocol strategy degraded operators). Go's resolver handles EDNS, TCP fallback, ndots, and search domains differently than HAProxy's. Subtle differences in how these two resolvers interact with CoreDNS could surface as hard-to-debug failures. DNS is a bit messy.

- **DNS resolution blocks the informer event loop.** Resolution is synchronous on the event handler goroutine. A single slow FQDN can block for 25s (10s timeout + 5s retry + 10s timeout), capped at 30s per EndpointSlice. A namespace admin with `edit` permissions can create FQDN EndpointSlices pointing at unresolvable hostnames and stall all endpoint processing cluster-wide.

So to me, it feels like we are giving up on a well-tested DNS solution and moving to new homegrown DNS solution that requires us to fight through the same bugs that HAProxy has with worse performance.

You might say, well if you are just listing the reasons why this won't work, what's the better solution then? Hear me out....this could be an easy HAProxy upstream solution. What if we contributed something like `resolve-deny` to upstream haproxy:

```
resolvers ingress_dns
    parse-resolv-conf
    resolve_retries 1
    timeout retry 5s
    timeout resolve 10s
    resolve-deny 127.0.0.0/8
    resolve-deny 169.254.0.0/16
    resolve-deny 168.63.129.16/32
    resolve-deny fd00:ec2::254/128
    resolve-deny ::1/128
```

It's actually quite easy to implement, I think we could even implement it and possibly carry patch it if upstream denies. We'd just put a skip [here](https://github.com/haproxy/haproxy/blob/146015e58b142246b529390abd8886219460c344/src/resolvers.c#L1687) like:
```
  /* Check against deny list */
  for (j = 0; j < resolvers->deny_net_nb; j++) {
      if (ip_type == resolvers->deny_net[j].family &&
          ((ip_type == AF_INET &&
            in_net_ipv4(ip, &resolvers->deny_net[j].mask.in4,
                        &resolvers->deny_net[j].addr.in4)) ||
           (ip_type == AF_INET6 &&
            in_net_ipv6(ip, &resolvers->deny_net[j].mask.in6,
                        &resolvers->deny_net[j].addr.in6))))
          goto next;  // skip this IP
```

Has that been considered? HAProxy may be very interested in this because it's security related, but I don't know. Carrying this patch even if upstream doesn't accept, is definitely a headache, but it might be the lesser of two evils. Overall, less code and behavior changes while maintaining these features that HAProxy offers us.

Did anyone ever discuss this approach? Has it been ruled out? Let me know, I might be missing context or discussion. CC: @jcmoraisjr @Miciah @rikatz

gcs278:
I've opened https://github.com/openshift/router/pull/788 which is a carry-patch of the HAProxy fix, as a proof of concept. I've also suggested the feature in the HAProxy upstream repo https://github.com/haproxy/haproxy/issues/3437. Early feedback from the HAProxy community would help us decide whether this is an acceptable approach to the problem.

Open to suggestions here too. I'm not a fan of either solution (upstream with potential carry vs. rewriting our DNS logic), but trying to be pragmatic. A carrypatch & pursue upstream at least an alternative to consider.

gcs278:
Update. I realized that since PR #677 hasn't shipped in any release, the comparison for behavioral regressions should be against the current released behavior (4.22) where HAProxy resolves FQDNs once at startup with no runtime re-resolution. Here's an updated list:

- ~~**DNS freshness degrades from ~10s to ~30min.**~~ Both the current released behavior and PR #788 resolve DNS once — neither has runtime re-resolution. Not a regression from what's shipped.

- ~~**Failure handling is worse.**~~ On released versions, HAProxy crashes entirely if a hostname can't be resolved at startup ([OCPBUGS-55506](https://issues.redhat.com/browse/OCPBUGS-55506)). PR #788 silently drops the address, which is arguably better than crashing the whole router.

- **Still Applies: Switching from HAProxy's C resolver to Go's net.Resolver is risky.** We've been burned by DNS resolver differences before — [OCPBUGS-27397](https://issues.redhat.com/browse/OCPBUGS-27397) (SRV lookups broke when CoreDNS changed EDNS behavior), [OCPBUGS-240](https://issues.redhat.com/browse/OCPBUGS-240) (UDP payload size change from 512 to 2048 broke customer F5 firewalls), [OCPBUGS-35104](https://issues.redhat.com/browse/OCPBUGS-35104) (TCP protocol strategy degraded operators). Go's resolver handles EDNS, TCP fallback, ndots, and search domains differently than HAProxy's. Subtle differences in how these two resolvers interact with CoreDNS could surface as hard-to-debug failures.

- **Still Applies: DNS resolution blocks the informer event loop.** Resolution is synchronous on the event handler goroutine. A single slow FQDN can block for 25s (10s timeout + 5s retry + 10s timeout), capped at 30s per EndpointSlice. A namespace admin with `edit` permissions can create FQDN EndpointSlices pointing at unresolvable hostnames and stall all endpoint processing cluster-wide.

The first two regressions were deal breakers for me, but they are not a concern now. While the last two are definitely not great, but probably acceptable until HAProxy upstream decides if the `resolve-deny` is a reasonable approach. We can adjust later.

I'll follow up with a review of this PR soon.

rikatz:
/close
Moved to #812

PR #794, opened 2026-06-10
OCPBUGS-87082: Reload HAProxy when adding the first backend server

jcmoraisjr:
The following scenario is happening on HAProxy 2.8:

* Configure a new route resource, which references a service whose deployment does not have any replica;
* Scale out to 3 replicas;
* Check the balance of the route: only 2 of the 3 replicas are added to the load balance;
* If scaling one by one, the issue happens adding the third replica, which makes the second one to be removed from the load balance.

This only happens with HAProxy 2.8, even latest (2.8.24). HAProxy 3.2 does not have this issue.

This update circumvents this behavior by requesting a reload in case a backend is being scaled from zero to one or more replicas, this way the first backend servers are added via configuration files. All the other new backend servers can be added dynamically.

https://redhat.atlassian.net/browse/OCPBUGS-87082

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced compatibility with HAProxy 2.8 by adding safeguards for dynamically added endpoints in empty backends.
  * Improved router behavior to ensure proper reloads when backend services transition from zero to non-zero endpoints.

### DIFF
--- a/pkg/router/template/router.go
+++ b/pkg/router/template/router.go
@@ -886,6 +886,32 @@ func (r *templateRouter) dynamicallyReplaceEndpoints(id ServiceUnitKey, service 
 		return false
 	}
 
+	// HAProxy 2.8 is dropping the second added server from the balance when the third one is added,
+	// and all of them were dynamically added into an empty backend. HAProxy 3.2 does not have this issue.
+	// This extra check ensures that the first dynamically added server happens by reloading HAProxy.
+	// This can be removed after dropping support for HAProxy 2.8.
+	//
+	// TODO: make this conditional to HAProxy 2.8 when adding support to HAProxy 3.2.
+	var oldServerCount, newServerCount int
+	for backendKey := range service.ServiceAliasAssociations {
+		if cfg, found := r.state[backendKey]; found {
+			oldEndpoints := cfg.EndpointTable[id]
+			newEndpoints := endpointsForAlias(cfg, service)
+			oldServerCount += len(oldEndpoints)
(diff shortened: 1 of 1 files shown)


pkg/router/template/router.go
```diff
@@ -886,6 +886,32 @@ func (r *templateRouter) dynamicallyReplaceEndpoints(id ServiceUnitKey, service
...
+	// TODO: make this conditional to HAProxy 2.8 when adding support to HAProxy 3.2.
+	var oldServerCount, newServerCount int
+	for backendKey := range service.ServiceAliasAssociations {
+		if cfg, found := r.state[backendKey]; found {
+			oldEndpoints := cfg.EndpointTable[id]
+			newEndpoints := endpointsForAlias(cfg, service)
+			oldServerCount += len(oldEndpoints)
+			newServerCount += len(newEndpoints)
+		}
+	}
```
jcmoraisjr, line 903: This is counting the endpoints of all the backends backing this service, but ignoring the backends having more than one service. This should cause an unneeded reload for backends having backend server being served by other service.

Maybe a better approach is to iterate over all the backends, and from each of them count all the endpoints from all their services, and short-circuit as soon as the first one is found. cfg.EndpointTable should always be updated.

gcs278:
@jcmoraisjr should this be closed now that we decided to go with fixing it with a backport port in the HAProxy source code?

jcmoraisjr:
Yeah let me close it, we're going to patch upstream instead.

/close

PR #796, opened 2026-06-11
OCPBUGS-87082: Use ID when adding backend server dynamically

jcmoraisjr:
The following scenario is happening on HAProxy 2.8:

* Configure a new route resource, which references a service whose deployment does not have any replica;
* Scale out to 3 replicas;
* Check the balance of the route: only 2 of the 3 replicas are added to the load balance pool.

If scaling one by one, the issue happens adding the third replica, which makes the second one to be removed from the load balance pool.

This is a HAProxy 2.8 only issue, 3.0+ does not have this behavior. It happens due to the missing of the backend server ID, which is being assigned late in this scenario.

This patch declares the backend server ID whenever a new backend server is added dynamically, which circumvents the misbehavior from the HAProxy side. A new state field controls the last backend server ID used, so new ones use newer IDs. The field starts as 2^24 to avoid colision with the automatically generated ones from the backend servers declared in the configuration file. It is also reinitialized after a successful reload, since all the dynamically added servers became part of the configuration file.

Although a DCM related fix, this only impacts 5.0 since this only happens in the add server API call, only added on this version.

https://redhat.atlassian.net/browse/OCPBUGS-87082

## Summary by CodeRabbit

* **Bug Fixes**
  * HAProxy backend servers now receive explicit, tracked IDs during dynamic endpoint updates, improving reliability and avoiding ID collisions during server add/remove operations.

* **Tests**
  * Test suite updated to validate explicit server ID assignment and ID allocation behavior across backend server operations.

### DIFF
--- a/pkg/router/template/configmanager/haproxy/backend_test.go
+++ b/pkg/router/template/configmanager/haproxy/backend_test.go
@@ -26,6 +26,7 @@ func TestBackendDynamicUpdate(t *testing.T) {
 		endpointID      *string                               // default: "server1"
 		ip              *string                               // default: "10.0.1.11"
 		port            *string                               // default: "9000"
+		backendServerID *int32                                // default: 1
 		weight          *int32                                // default: 1
 		workingDir      *string                               // default: "tmp"
 		publicHostname  string
@@ -47,7 +48,15 @@ func TestBackendDynamicUpdate(t *testing.T) {
 		"should add insecure server": {
 			cmd: cmdAdd,
 			cmdExpected: []string{
-				"add server route1/server1 10.0.1.11:9000 weight 1 check inter 5000ms",
+				"add server route1/server1 10.0.1.11:9000 weight 1 id 1 check inter 5000ms",
+				"set server route1/server1 state ready",
+			},
+		},
+		"should add insecure server with custom backend server ID": {
+			cmd:             cmdAdd,
+			backendServerID: ptr.To[int32](1023),
+			cmdExpected: []string{
+				"add server route1/server1 10.0.1.11:9000 weight 1 id 1023 check inter 5000ms",
 				"set server route1/server1 state ready",
 			},
 		},
@@ -55,7 +64,7 @@ func TestBackendDynamicUpdate(t *testing.T) {
 			cmd:    cmdAdd,
 			weight: ptr.To[int32](0),
 			cmdExpected: []string{
-				"add server route1/server1 10.0.1.11:9000 weight 0 check inter 5000ms",
+				"add server route1/server1 10.0.1.11:9000 weight 0 id 1 check inter 5000ms",
 				"set server route1/server1 state drain",
 			},
 		},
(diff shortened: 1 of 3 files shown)


pkg/router/template/configmanager/haproxy/backend_test.go
```diff
@@ -47,15 +48,24 @@ func TestBackendDynamicUpdate(t *testing.T) {
 		"should add insecure server": {
 			cmd: cmdAdd,
 			cmdExpected: []string{
-				"add server route1/server1 10.0.1.11:9000 weight 1 check inter 5000ms",
+				"add server route1/server1 10.0.1.11:9000 weight 1 id 1 check inter 5000ms",
 				"set server route1/server1 state ready",
 			},
 		},
+		"should add insecure server with custom puid": {
```
Miciah: I guess "puid" stands for "proxy-unique server ID"? https://github.com/haproxy/haproxy/blob/ae90be6491cfe0cb0a16bfd5b03babfaf312b0bf/include/haproxy/server-t.h#L361

It would probably be better just to write "id" since we don't use the term "puid" anywhere in the router code.

jcmoraisjr: Proxy unique server ID is exactly what this server ID is, but in fact the terminology is not the same from the rest of the code. Coderabbit also suggested this change, let me fix.

pkg/router/template/configmanager/haproxy/manager.go
```diff
@@ -732,6 +742,14 @@ func (cm *haproxyConfigManager) reset() {
...
 	// Reset the client - clear its caches.
 	cm.client.Reset()
+
+	// Start/reset backend server ID as 2^24, giving the first 16.7mm to the
+	// automatically assigned ones via configuration file.
+	//
+	// It is loaded behind !r.synced and cm.isReloading(), so it is initialized
+	// before first use, not used during reload, and also it is only reset after
+	// the reload finished successfully.
+	cm.currentServerID.Store(1 << 24)
```
Miciah, line 752: Can large id values affect memory usage?  Have you done any performance analysis?  

I ask because we did observe that large weight values caused significant increases in memory usage; see <https://bugzilla.redhat.com/show_bug.cgi?id=2007581#c40> and [NE-709](https://redhat.atlassian.net/browse/NE-709).  

Hm, I see in https://github.com/haproxy/haproxy/blob/ae90be6491cfe0cb0a16bfd5b03babfaf312b0bf/src/lb_chash.c#L500-L510 that the weight value is a factor in `lb_nodes_tot`, which is used in `calloc(srv->lb_nodes_tot, sizeof(*srv->lb_nodes))` and therefore directly impacts memory usage, but `puid` is only a factor in a value that is hashed to generate a key.  Based on that, it seems that the id value has no impact on memory usage.  

@coderabbitai, can you confirm the preceding paragraph?

jcmoraisjr:
We're going to patch upstream instead, closing this PR.

/close

PR #803, opened 2026-06-24
[WIP] OCPBUGS-88729: Fix HSTS annotation regex to reject newlines and accept trailing semicolons

Thealisyed:
## Summary

- Replace `\s*` with `[ \t]*` in the HSTS regex pattern to reject newline characters (`\n`, `\r`, `\f`) that break HAProxy config on reload and can cause cluster-wide router unavailability
- Allow an optional trailing semicolon in HSTS annotation values, which were previously silently rejected causing the `Strict-Transport-Security` header to not be set
- Add 6 new unit tests covering newline rejection (3 cases) and trailing semicolon acceptance (3 cases)

## The Issue

The HSTS regex in `haproxy-config.template:31` used `\s*` to match whitespace between directives. In Go, `\s` matches `[\t\n\f\r ]` — including newlines. An annotation like `max-age=31536000\n;\npreload` would pass the regex, render literal newlines into the HAProxy config, and cause HAProxy to reject the entire config on reload — taking down all routes cluster-wide.

## Solution

Two changes to the same regex:
1. `\s*` → `[ \t]*` — restricts whitespace matching to horizontal whitespace only (spaces and tabs)
2. Append `(?:[ \t]*;)?` — allows an optional trailing semicolon, fixing a silent failure that has existed since 2018 ([openshift/origin#19860](https://github.com/openshift/origin/issues/19860))

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -28,7 +28,7 @@
 {{- /* hsts header in response: */}}
 {{- /* Not fully compliant to RFC6797#6.1 yet: has to accept not conformant directives */}}
 {{- $hstsOptionalTokenPattern := `(?:includeSubDomains|preload)` }}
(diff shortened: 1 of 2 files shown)


Thealisyed:
input images failure

/retest

PR #811, opened 2026-07-07
NE-2741: Implement TLS Curves Support for Metrics Endpoints

bentito:
This pull request implements TLS curve support for the router's metrics server, aligning with the `TLSGroupPreferences` API for PQC readiness.

**Changes:**
- Update `openshift/api` to integrate `TLSGroupPreferences` definitions.
- Vendor the updated API dependencies.
- Update `makeTLSConfig` to read the `ROUTER_METRICS_TLS_CURVES` environment variable.
- Implement `tlsGroupsToCurveIDs` helper to map `TLSGroup` strings to `tls.CurveID`.
- Add unit tests for the curve mapping logic.

**Jira:** [NE-2741](https://redhat.atlassian.net/browse/NE-2741)

***

**Test Infrastructure Fix: UNIX Socket Path Lengths & Deadlocks**
This PR includes a fix for a consistent test deadlock in `configmanager/haproxy` experienced on macOS (and intermittently causing flakes in CI).

The fake HAProxy test server (`StartFakeServerForTest`) previously used the full test name to construct a UNIX domain socket path (e.g., `fake-haproxy-TestClientRunInfoCommandConverter...`). When combined with the OS temporary directory path, this string frequently exceeded the strict maximum path length for UNIX domain sockets (104 bytes on macOS, 108 bytes on Linux). We proved this locally by observing the OS actively rejecting the 105-character path with `bind: invalid argument`.

Previously, when `net.Listen` failed due to this length limit, the mock server's background goroutine would silently exit `return err`, leaving the main test thread permanently deadlocked waiting for a `<-started` signal until it hit a 10-minute timeout.

**Resolution:**
1. **Shortened the socket prefix** to simply `fake-haproxy-` (dropping the verbose test name). This saves ~35 characters, ensuring the socket path safely fits within the OS limits across all environments.
2. **Propagated the error:** Updated the `Start()` method to explicitly pass `net.Listen` errors back to the main thread via a channel. This guarantees that if a bind failure ever happens again (due to permissions, resource exhaustion, etc.), the test suite will instantly panic with a clear error message rather than silently deadlocking for 10 minutes.

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring TLS curve preferences for the router metrics endpoint.
  - Administrators can specify supported TLS groups using comma- or colon-separated values.
  - Supports X25519, P-256, P-384, P-521, and X25519MLKEM768 curves.

- **Bug Fixes**
  - Invalid or unsupported TLS curve settings now produce clear configuration errors instead of being silently ignored.

### DIFF
--- a/pkg/cmd/infra/router/template.go
+++ b/pkg/cmd/infra/router/template.go
@@ -974,25 +974,26 @@ func makeTLSConfig(reloadPeriod time.Duration) (*tls.Config, error) {
 		secureTLSConfig.MinVersion = crypto.TLSVersionOrDie(versionName)
 	}
 	if curvesStr := env("ROUTER_METRICS_TLS_CURVES", ""); len(curvesStr) > 0 {
-		var curveNames []string
-		if strings.Contains(curvesStr, ",") {
-			curveNames = strings.Split(curvesStr, ",")
-		} else {
-			curveNames = strings.Split(curvesStr, ":")
-		}
+		curveNames := strings.FieldsFunc(curvesStr, func(c rune) bool {
+			return c == ',' || c == ':'
+		})
 
 		var groups []configv1.TLSGroup
 		for _, name := range curveNames {
-			groups = append(groups, configv1.TLSGroup(strings.TrimSpace(name)))
+			trimmed := strings.TrimSpace(name)
+			if len(trimmed) > 0 {
+				groups = append(groups, configv1.TLSGroup(trimmed))
+			}
 		}
 
 		curvePrefs, unsupportedGroups := tlsGroupsToCurveIDs(groups)
 		if len(unsupportedGroups) > 0 {
-			log.V(0).Info("some curves from ROUTER_METRICS_TLS_CURVES are not supported", "unsupportedCurves", unsupportedGroups)
+			return nil, fmt.Errorf("some curves from ROUTER_METRICS_TLS_CURVES are not supported: %v", unsupportedGroups)
 		}
-		if len(curvePrefs) > 0 {
-			secureTLSConfig.CurvePreferences = curvePrefs
+		if len(curvePrefs) == 0 {
+			return nil, errors.New("no valid curves found in ROUTER_METRICS_TLS_CURVES")
 		}
+		secureTLSConfig.CurvePreferences = curvePrefs
 	}
 
 	return secureTLSConfig, nil
--- a/pkg/cmd/infra/router/template.go
+++ b/pkg/cmd/infra/router/template.go
@@ -30,6 +30,7 @@ import (
 	authenticationclient "k8s.io/client-go/kubernetes/typed/authentication/v1"
 	authorizationclient "k8s.io/client-go/kubernetes/typed/authorization/v1"
 
+	configv1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	projectclient "github.com/openshift/client-go/project/clientset/versioned"
 	routeclientset "github.com/openshift/client-go/route/clientset/versioned"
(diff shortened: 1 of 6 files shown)


pkg/cmd/infra/router/template.go
```diff
@@ -974,10 +975,54 @@ func makeTLSConfig(reloadPeriod time.Duration) (*tls.Config, error) {
...
+		}
+		secureTLSConfig.CurvePreferences = curvePrefs
+	}
 
 	return secureTLSConfig, nil
 }
 
+// tlsGroupsToCurveIDs converts openshift/api TLSGroup identifiers to Go crypto/tls CurveID values.
+// Returns the mapped curve preferences and any group names not supported by this Go version.
+func tlsGroupsToCurveIDs(groups []configv1.TLSGroup) ([]tls.CurveID, []string) {
```
davidesalerno, line 1015: I'll put a TODO here to replace this method with the `library-go` equivalent (`CurveIDsForTLSGroups`) when this [PR](https://github.com/openshift/library-go/pull/2347/) will be merged

pkg/cmd/infra/router/template.go
```diff
@@ -974,10 +975,54 @@ func makeTLSConfig(reloadPeriod time.Duration) (*tls.Config, error) {
 	if versionName := env("ROUTER_METRICS_TLS_MIN_VERSION", ""); len(versionName) > 0 {
 		secureTLSConfig.MinVersion = crypto.TLSVersionOrDie(versionName)
 	}
+	if curvesStr := env("ROUTER_METRICS_TLS_CURVES", ""); len(curvesStr) > 0 {
+		curveNames := strings.FieldsFunc(curvesStr, func(c rune) bool {
```
davidesalerno: We could extract the code till line 998 in a dedicated function (i.e. `parseCurvePreferences`) for testability of delimiter handling and fail-closed validation

pkg/router/template/configmanager/haproxy/testing/haproxy.go
```diff
@@ -129,8 +131,14 @@ func (p *fakeHAProxy) Start() {
 		}
 	}()
 
-	// wait for server to indicate it started up.
-	<-started
+	// wait for server to indicate it started up or failed.
+	select {
+	case <-started:
+	case err := <-listenErr:
+		if err != nil {
```
davidesalerno: I think that we can probably remove the nil check since if we are in this case the error is not null

pkg/cmd/infra/router/template_test.go
```diff
@@ -133,3 +135,43 @@ func TestParseHeadersToBeSetOrDeleted(t *testing.T) {
 		})
 	}
 }
+
+func TestTlsGroupsToCurveIDs(t *testing.T) {
```
davidesalerno: I'll rename it to 'TestTLSGroupsToCurveIDs'

pkg/cmd/infra/router/template_test.go
```diff
@@ -133,3 +135,43 @@ func TestParseHeadersToBeSetOrDeleted(t *testing.T) {
...
+	testCases := []struct {
```
davidesalerno, line 141: It seems to me that we can add a test case for all-unsupported input to TestTlsGroupsToCurveIDs to confirms the len(curvePrefs) == 0 error path works correctly

davidesalerno (commented):
Just some minor feedbacks

pkg/router/template/configmanager/haproxy/testing/haproxy.go
bentito: Yep, addressed in [0d8c9863b88c](https://github.com/openshift/router/pull/811/commits/0d8c9863b88ce0b56f1dc138b1f77b9146b3f028)

pkg/cmd/infra/router/template_test.go
bentito: Yep, addressed in [0d8c9863b88c](https://github.com/openshift/router/pull/811/commits/0d8c9863b88ce0b56f1dc138b1f77b9146b3f028)

pkg/cmd/infra/router/template.go
bentito: Yep, addressed in 0d8c9863b88c

pkg/cmd/infra/router/template_test.go
bentito: Yep, addressed in [0d8c9863b88c](https://github.com/openshift/router/pull/811/commits/0d8c9863b88ce0b56f1dc138b1f77b9146b3f028)

pkg/cmd/infra/router/template.go
bentito: Yep, addressed in 0d8c9863b88c

PR #823, opened 2026-07-23
RFE-2924: Return HTTP 429 when route HTTP rate limit is exceeded

sanjaytripathi97:
## Summary
- When `haproxy.router.openshift.io/rate-limit-connections.rate-http` is exceeded, HAProxy previously used `tcp-request content reject`, which dropped the connection with no HTTP status (clients saw "Empty reply from server").
- Change that path to `http-request deny deny_status 429` so rate-limited HTTP clients receive **HTTP 429 Too Many Requests**.
- Leave `concurrent-tcp` and `rate-tcp` as `tcp-request content reject` (TCP-layer limits; no HTTP response is possible). Passthrough backends are unchanged (no `rate-http` there).
Fixes: https://issues.redhat.com/browse/RFE-2924
## Changes
- `images/router/haproxy/conf/haproxy-config.template`: use `http-request deny deny_status 429` for `rate-http`.
- `pkg/router/router_test.go`: add `TestConfigTemplate` cases for edge and insecure backends.
## Test plan
- [ ] `CGO_ENABLED=1 go test -mod=vendor -count=1 -timeout 120s -v ./pkg/router/ -run 'TestConfigTemplate'`
- [ ] Manual: annotate a route with rate-limit + low `rate-http`, exceed the limit, confirm response is `429` (not empty reply)
- [ ] Confirm `concurrent-tcp` / `rate-tcp` still reject without an HTTP status
- [ ] Confirm passthrough routes with TCP rate limits are unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * HTTP requests exceeding configured per-route rate limits now receive a clear **429 Too Many Requests** response instead of a generic rejection.
  * TCP connection limits continue to use connection-level rejection behavior.

* **Tests**
  * Added coverage for HTTP rate limiting on secure and non-secure routes.

### DIFF
--- a/images/router/haproxy/conf/haproxy-config.template
+++ b/images/router/haproxy/conf/haproxy-config.template
@@ -708,7 +708,9 @@ backend {{ genBackendNamePrefix $cfg.TLSTermination }}:{{ $cfgIdx }}
           {{- end }}
 
           {{- if (isInteger (index $cfg.Annotations "haproxy.router.openshift.io/rate-limit-connections.rate-http")) }}
-  tcp-request content reject if { src_http_req_rate ge {{ index $cfg.Annotations "haproxy.router.openshift.io/rate-limit-connections.rate-http" }} }
+  # Return HTTP 429 when the per-source HTTP request rate limit is exceeded (RFE-2924).
+  # TCP connection limits above still use tcp-request reject because no HTTP response is possible there.
+  http-request deny deny_status 429 if { src_http_req_rate ge {{ index $cfg.Annotations "haproxy.router.openshift.io/rate-limit-connections.rate-http" }} }
           {{- else }}
   #HTTP request rate not restricted
           {{- end }}
--- a/pkg/router/router_test.go
+++ b/pkg/router/router_test.go
@@ -976,6 +976,47 @@ func TestConfigTemplate(t *testing.T) {
 				},
 			},
 		},
+		"HTTP rate limit returns 429": {
+			mustCreateWithConfig{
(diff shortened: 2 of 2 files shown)


sanjaytripathi97:
@candita will you be able to review this PR,  any reviews please ?

gcs278:
@sanjaytripathi97 thanks for the contribution!

There are a couple things about this PR:
- This PR is adding an annotation, and while we support existing annotations, we do all new configurations as proper API fields. Annotations are not validated, not well documented/discoverable, and lack other benefits of an API.
- We are unable to accept new features tied directly to RFEs without an enhancement proposal. This [doc](https://github.com/openshift/enhancements/blob/master/dev-guide/feature-zero-to-hero.md) is a good place to start with the process.

My suggestion - if you'd like to drive this forward, please work with our team's stakeholders to get a feature created and open an enhancement PR.

sanjaytripathi97:
Thanks for the feedback @gcs278 !!

I was not adding a new annotation, I try to change the behavior of the existing
`haproxy.router.openshift.io/rate-limit-connections.rate-http` annotation so that, when the
HTTP request rate limit is exceeded, HAProxy returns HTTP 429 instead of using
`tcp-request content reject` (which drops the connection without an HTTP status).

The change is limited to HTTP-mode route backends (edge/reencrypt/plain HTTP). TCP limits
(`concurrent-tcp`, `rate-tcp`) are unchanged and still use TCP-level reject.

I'll work on an enhancement proposal covering what you have suggested.

PR #745, opened 2026-03-03
OCPBUGS-77056: Make external cert validation asynchronous

bentito:
This PR implements asynchronous and cached `SubjectAccessReview` checks for `spec.tls.externalCertificate` during router startup.

Previously, each external certificate triggered synchronous API validations that blocked the main thread. When scaling to thousands of external cert routes, the router hit `O(N * API_latency)` delays and would hit `CrashLoopBackOff` from failing liveness probes.

This fix maintains full RBAC security checks but executes them asynchronously in the background. A global `sync.Map` acts as an `asyncSARCache` to drastically speed up repeated checks for the same underlying secret.

Fixes: OCPBUGS-77056

## Summary by CodeRabbit

* **New Features**
  * Cache-backed, semaphore-throttled validation for external certificates with configurable concurrent SAR checks.
  * Configurable parallelism for route status writer lease workers.

* **Improvements**
  * Hybrid secret informer strategy (faster where allowed; more selective when listing is forbidden).
  * Higher Kubernetes client request throughput for router startup.
  * Architecture-aware debug image builds and more robust route condition/status comparisons.

* **Bug Fixes**
  * Preserve “ignored” admission condition details and record completion of external-certificate access checks.

* **Tests**
  * Updated and added unit tests for secret/SAR caching behavior and status logic.

PR #828, opened 2026-07-29
OCPBUGS-77056: Make external cert validation asynchronous (v3 — fix re-admission after secret deletion)

bentito:
~> **Draft — waiting on revert of #825 (v2) to land before rebase.**~

## Background

This is the fourth attempt at landing asynchronous external certificate validation. PR #825 (v2) fixed the `x509: ECDSA verification failure` race conditions from #822 but introduced a new regression: 5 out of 10 payload-aggregate runs failed on `RouteExternalCertificate` tests.

### v2 payload results (5/10 — not 10/10)

The `/payload-aggregate` run for #825 ([results](https://pr-payload-tests.ci.openshift.org/runs/ci/b2ed0490-8ad9-11f1-901b-0cff623bf83c-0)) showed **5 successes, 5 failures**. The failing runs had two blocking `RouteExternalCertificate` test failures:

1. **"the secret is deleted then routes are not reachable"** — timed out at 900s
2. **"the secret is updated but RBAC permissions are dropped then routes are not reachable"** — timed out at 900s

Notably, **zero `x509` or `ECDSA` errors** appeared anywhere in the logs — the v2 race condition fixes (DeepCopy, atomic PEM write) eliminated those. The new failure mode is a **status transition timeout**: the route never reaches `Admitted=False` and the test polls until the 15-minute timeout.

### Root cause: SARCompleted re-admits a deleted-secret route

After `DeleteFunc` rejects a route (`Admitted=False, ValidationFailed`), the status update triggers a re-enqueue. The subsequent `HandleRoute(Modified)` call:

1. DeepCopies the route (capturing `Admitted=False` in the local copy)
2. Calls `validate()` — SAR check passes (RBAC is still valid, only the secret was deleted)
3. Calls `populateRouteTLSFromSecret()` — `GetSecret` **succeeds** because the informer cache hasn't propagated the deletion yet
4. Plugin chain accepts the route
5. The `SARCompleted` guard checks the DeepCopied route, sees `Admitted=False` (no ext-cert reason), and writes `SARCompleted` — **flipping the route back to `Admitted=True`**

The route bounces between `Admitted=False` (from DeleteFunc) and `Admitted=True` (from SARCompleted), and the E2E test polls for `Admitted=False` until timeout.

## What's new in this PR (net new vs #825)

One commit on top of the full v2 (#825) changeset:

### Fix: check `deletedSecrets` before writing SARCompleted

Before emitting `RecordRouteUpdate(SARCompleted)`, check `p.deletedSecrets.Load(key)`. If the secret has been marked as deleted by the `DeleteFunc`, skip the write. This prevents re-admission during the informer cache propagation window.

### Test: `TestDeletedSecretDoesNotGetReadmitted`

Reproduces the exact failure:
1. Admits a route with an external cert (SARCompleted written)
2. Fires `DeleteFunc` (route rejected with `ValidationFailed`)
3. Calls `HandleRoute(Modified)` while `GetSecret` still succeeds (cache race)
4. Asserts the route is NOT re-admitted

**Fails on v2 code** (2 SARCompleted writes = re-admission), **passes after the fix**.

## Update: root cause found and fixed — synchronous certificate refresh

Post-review, payload-aggregate kept showing the `RouteExternalCertificate` "secret is updated but RBAC permissions are dropped" test failing on roughly half of the runs, despite passing PR-time tests every time. We iterated through several mitigations narrowing when `SARCompleted` fires and adding a delayed RBAC re-check after a secret update — these reduced but didn't eliminate the flake.

**Root cause:** `UpdateFunc` only wrote a status condition on secret change. The actual certificate refresh (`populateRouteTLSFromSecret`) only ran when the router's own route-watch happened to redeliver the route as `Modified` — an indirect round trip through its own status write. Found by locally injecting API-server latency and packet loss into the router's connection (emulating HyperShift's separated control plane), since this never reproduced under CI's PR-time tests or low-latency local runs. Under injected latency/loss, that round trip could be delayed or dropped, leaving a route serving a stale certificate indefinitely until an unrelated later event (e.g. the next secret rotation) happened to reprocess it. This is also the direct answer to @jcmoraisjr's review question about whether the 3s delayed re-check was "time enough" — that goroutine only ever re-ran the SAR check, never the cert refresh, so it provided no protection against this specific failure mode.

**Fix:** `UpdateFunc` now calls `validate()` and `populateRouteTLSFromSecret()` synchronously and pushes the result through the plugin chain directly, instead of depending on the router observing its own write. The delayed re-check goroutine is unchanged and now purely serves its original purpose: catching RBAC revocations that haven't propagated yet.

**Simplification:** also investigated increasing `writerlease`'s worker count (1→N), reasoning a single worker could head-of-line-block status writes under high latency. Built a local repro harness (6 routes sharing one secret, matching CI's topology, with `tc netem` injecting 1s±300ms latency, 10% loss, 2% duplication into the router's API-server connection) and empirically disproved this — increasing workers made no measurable difference, and the failure signature (all 6 routes in a namespace failing or succeeding together) ruled out a per-worker queuing bottleneck. Kept `writerlease` at its original single worker rather than adding concurrency that wasn't fixing anything.

**Validation** — same local harness, before vs. after the synchronous-refresh fix, under identical injected latency/loss:
- Before: 137/360 checks timed out (38%)
- After: 3/360 timed out (0.8%) — residual failures show a distinct, much rarer race between this fix's synchronous path and the router's independent periodic re-validation running concurrently on a different goroutine; tracked separately as a follow-up, not blocking.

CI: `/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10` — 10/10 clean, zero `RouteExternalCertificate` failures (only pre-existing, unrelated non-blocking flakes elsewhere in the suite).

## Test plan

- [x] `go test -race ./pkg/router/controller/` — all pass, including `TestDeletedSecretDoesNotGetReadmitted`
- [x] `make verify` and `make check` — pass
- [x] `/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10` — 10/10, zero RouteExternalCertificate failures
- [x] Scale test: 2000 ext-cert routes admitted in ~120s (verified: 2000/2000 in 138s, 0 rejections, on a fresh 4.21 cluster with the synchronous-refresh fix)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bentito:
/payload-aggregate-with-prs periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 20

pkg/router/template/certmanager.go
```diff
@@ -178,15 +178,41 @@ func newSimpleCertificateWriter() certificateWriter {
...
+	if err := tmpFile.Close(); err != nil {
+		os.Remove(tmpName)
+		log.Error(err, "error closing temp certificate file", "file", tmpName)
+		return err
+	}
+	if err := os.Chmod(tmpName, 0644); err != nil {
+		os.Remove(tmpName)
+		log.Error(err, "error setting permissions on temp certificate file", "file", tmpName)
+		return err
+	}
```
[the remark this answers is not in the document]
bentito: we'll address this in a separate PR as it could have impacts on HAProxy

hack/Makefile.debug
```diff
@@ -11,8 +12,8 @@ REMOTE_IMAGE ?= $(REGISTRY)/$(IMAGE):$(TAG)
 OPENSHIFT_ENDPOINT ?= $(shell oc config view --minify --template '{{(index .clusters 0).cluster.server}}' | grep -o '//[^ :]*' | sed 's/^..//')
 
 new-openshift-router-image:
-	GO111MODULE=on CGO_ENABLED=0 GOFLAGS=-mod=vendor go build -o openshift-router -gcflags=all="-N -l" ./cmd/openshift-router
-	$(IMAGEBUILDER) build -t $(LOCAL_IMAGE) -f hack/Dockerfile.debug .
+	GO111MODULE=on CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) GOFLAGS=-mod=vendor go build -o openshift-router -gcflags=all="-N -l" ./cmd/openshift-router
+	$(IMAGEBUILDER) build --arch $(GOARCH) -t $(LOCAL_IMAGE) -f hack/Dockerfile.debug .
```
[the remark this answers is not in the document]
bentito: I think for building debug (dev) we're okay with being podman specific

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -85,60 +56,55 @@ func fakeSecret(namespace, name string, secretType corev1.SecretType, data map[s
...
-
-var _ router.Plugin = &fakePluginDone{}
-
 type statusRecorder struct {
+	sync.Mutex
 	rejections                 []string
 	updates                    []string
 	unservableInFutureVersions map[string]string
-	doneCh                     chan struct{}
 }
```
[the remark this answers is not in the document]
bentito: None of these tests call `RecordRouteUnservableInFutureVersions` — the `nil` map is never hit. The `statusRecorder` here is a minimal test double for the `RouteSecretManager` tests only

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -1157,6 +1139,286 @@ func TestRouteSecretManager(t *testing.T) {
...
+// not reachable": after the DeleteFunc rejects the route (Admitted=False),
+// the status update triggers a re-enqueue. If the subsequent HandleRoute
+// call succeeds (because GetSecret still returns the secret from cache
+// during the deletion propagation window), the SARCompleted write must
+// NOT flip the route back to Admitted=True. Otherwise the route bounces
+// between admitted and rejected, and the E2E test polls for Admitted=False
+// until timeout.
+//
+// Expected: FAILS on unfixed code (route is re-admitted via SARCompleted).
+func TestDeletedSecretDoesNotGetReadmitted(t *testing.T) {
```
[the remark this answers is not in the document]
bentito: Fixed in 6566c837 I updated the comment and tightened the assertion from sarCount > 1 to sarCount != 1

pkg/router/controller/shared_secret_manager.go
```diff
@@ -0,0 +1,300 @@
...
+				klog.V(2).Infof("namespace %s is restricted (RBAC), falling back to per-secret informers for routes", namespace)
+			}
+		}
+	}
+
+	m.registeredRoutes[key] = referencedSecret{
+		secretName: secretName,
+		handler:    handler,
+		restricted: restricted,
+	}
```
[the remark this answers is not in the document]
bentito: `RegisterRoute` is only called from the single-threaded controller loop, so concurrent registration of the same route cannot happen in practice. This is pre-existing from original fix PR #745.

pkg/router/controller/shared_secret_manager.go
```diff
@@ -0,0 +1,300 @@
...
+		m.lock.Lock()
+
+		// Re-check if someone else probed while we were unlocked.
+		if restricted, known = m.restrictedNamespaces[namespace]; !known {
+			restricted = (err != nil && apierrors.IsForbidden(err))
+			m.restrictedNamespaces[namespace] = restricted
+			if restricted {
+				klog.V(2).Infof("namespace %s is restricted (RBAC), falling back to per-secret informers for routes", namespace)
+			}
+		}
```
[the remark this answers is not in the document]
bentito: Valid observation, this is pre-existing from the orig fix PR #745. Practical risk is low since the probe runs at registration time when the API server is typically healthy. We can address in a follow-up.

pkg/router/controller/status.go
```diff
@@ -305,10 +302,11 @@ func handleRouteStatusUpdate(ctx context.Context, action string, oc client.Route
...
 		return writerlease.Release, false
 	case errors.IsConflict(err):
-		// just follow the normal process, and retry when we receive the update notification due to
-		// the other entity updating the route.
-		log.V(4).Info("updating route status failed due to write conflict", "action", action, "namespace", route.Namespace, "name", route.Name)
-		return writerlease.Release, true
+		// A write conflict is expected under high concurrency.
+		// Return None instead of Release to prevent this router from dropping its leader lease,
+		// which would otherwise cause a 60-second delay for subsequent routes.
+		log.V(4).Info("updating route status failed due to write conflict, retrying", "action", action, "namespace", route.Namespace, "name", route.Name)
```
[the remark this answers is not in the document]
bentito: Every other log line in this file and across the router logs namespace/name at V(4). This matches the existing pattern. Redacting route identifiers in router logs would make debugging impossible.

pkg/router/template/configmanager/haproxy/testing/haproxy.go
```diff
@@ -121,22 +118,18 @@ func (p *fakeHAProxy) Start() {
...
 			if shutdown {
-				return nil
+				return
 			}
 			conn, err := listener.Accept()
 			if err != nil {
-				return err
+				return
 			}
 			go p.process(conn)
```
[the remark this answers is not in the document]
bentito: This is pre-existing test helper behavior from orig PR #745. Surfacing connection errors in a fake server would add noise, errors here are typically "connection reset by peer" when tests complete

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

jcmoraisjr:
I manually compared the previous and already approved branch with the current one. It differs only in the deletedSecrets hashmap being checked before emitting `RecordRouteUpdate()`, along with a concise and well documented unit test exploring the scenario.

/approve
/lgtm

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

melvinjoseph86:
Marking this as verified based on https://github.com/openshift/router/pull/825#issuecomment-5114216858
/verified by @mjoseph

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

pkg/router/controller/route_secret_manager.go
```diff
@@ -289,43 +331,75 @@ func (p *RouteSecretManager) generateSecretHandler(namespace, routeName string)
...
+				time.Sleep(3 * time.Second)
+				routeapihelpers.InvalidateAsyncSARCache(namespace, secretNew.Name)
+				route, err := p.routelister.Routes(namespace).Get(routeName)
+				if err != nil {
+					return
+				}
+				route = route.DeepCopy()
+				retryMsg := fmt.Sprintf("secret %q re-validated for route %q", secretNew.Name, key)
+				p.recorder.RecordRouteUpdate(route, ExtCrtStatusReasonSecretUpdated, retryMsg)
+			}()
```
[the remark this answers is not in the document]
bentito: Valid observations. The goroutine is bounded (3s sleep, one API call, idempotent) and shutdown is benign since the writerlease queue rejects work after shutdown. Coalescing is a nice-to-have but adds complexity to a PR already at its fourth iteration. We'll consider it in a follow-up if secret update storms become a practical concern

melvinjoseph86:
Marking this as verified based on https://github.com/openshift/router/pull/825#issuecomment-5114216858
/verified by @mjoseph

pkg/router/controller/route_secret_manager.go
jcmoraisjr: How do we know 3s is time enough, what if env is slower than that? What if `Get(route)` fails? Is it expected to have two `RecordRouteUpdate()` calls, one inside and another outside the goroutine?

jcmoraisjr (commented):
Mostly lgtm, just a few comments in the goroutine. I however don't want to block the PR because of that, they can be handled as followups in case they make sense, along with the coderabbit suggestion.

/lgtm

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

pkg/router/controller/route_secret_manager.go
bentito: Thanks for flagging this @jcmoraisjr — it turned out to be directly connected to the root cause we were chasing on this PR.

Your question about whether 3s was "time enough" pointed right at it: that delayed goroutine only ever re-ran the SAR check, never the actual certificate refresh. We were relying on the router observing its own status write as a `Modified` watch event to trigger the real `populateRouteTLSFromSecret()` call — an indirect round trip. 

I setup a dev cluster to inject API-server latency/packet loss into the router(1) (to emulate HyperShift's separated control plane, since this only reproduced under real network conditions on Hypershift tests) showed that round trip can be delayed or dropped, leaving a route serving a stale certificate indefinitely until an unrelated later event happened to reprocess it. So the delayed re-check wasn't insufficiently timed — it just wasn't protecting against this failure mode at all.

Latest commit (b0361b8a) makes `UpdateFunc` call validate + populate + HandleRoute synchronously instead of depending on that round trip. The delayed goroutine you flagged now serves purely its original purpose: catching RBAC revocations that haven't propagated yet. That also answers your "two RecordRouteUpdate calls" question — one is the immediate status write to keep the route admitted, the other only fires from the RBAC-revocation path, they're no longer doing double duty for the cert-refresh case.

Confirmed with that dev cluster harness (injected 1s±300ms latency, 10% loss, 2% duplication into the router's API-server connection): failure rate dropped from 38% to 0.8% under the same conditions that previously reproduced the CI flake. Payload-aggregate came back 10/10 clean with zero RouteExternalCertificate failures.

---

(1) With Claude I was inspired to make great use of a dev cluster to replicate the origin payload aggregate tests but with much better debug cycle time, but it needed some way to simulate the HCP situation too. The solution used `tc` (Linux traffic control) with the `netem` (network emulation) queueing discipline — a tool for injecting artificial latency, jitter, packet loss, duplication, and reordering onto a network interface.

How it was used here specifically:

  1. Scoping to the right traffic: netem applies to a whole qdisc, so to avoid corrupting HAProxy's client-facing traffic (which would confound the test), I set up a prio qdisc with 4 bands as the root, then attached netem only to one band, then added a u32 filter matching dst 172.30.0.1/32 (the API server's ClusterIP) to steer only router→API-server traffic into that band. Everything else (client→router HTTPS) passed through unaffected.
  2. The netem parameters, tuned iteratively until they reproduced the CI failure signature:
```bash
  tc qdisc add dev eth0 root handle 1: prio bands 4 priomap 1 2 2 2 1 2 0 0 1 1 1 1 1 1 1 1
  tc qdisc add dev eth0 parent 1:4 handle 40: netem delay 1s 300ms loss 10% duplicate 2%
  tc filter add dev eth0 parent 1:0 protocol ip prio 1 u32 match ip dst 172.30.0.1/32 flowid 1:4
```
    That's 1s ± 300ms jitter, 10% packet loss, 2% duplication — landed there after starting at 150ms (no repro) and escalating.
  3. Applying it to the right network namespace: since the router pod itself doesn't have NET_ADMIN, I used ```oc debug node/<node>``` to get a privileged shell on the host, found the container's PID via `crictl inspect`, and used ```nsenter -t <PID> -n -- tc ... ``` to run the tc commands inside the pod's network namespace without modifying the pod spec.

pkg/router/controller/route_secret_manager.go
```diff
@@ -25,8 +27,21 @@ const (
...
+// before re-checking SAR permissions after a secret update, to catch RBAC
+// revocations that haven't propagated to the API server's authorizer yet.
+// Overridable in tests (via atomic Store/Load, since a prior test's spawned
+// goroutine may still be sleeping on this value when a later test runs) to
+// avoid real sleeps.
+var secretUpdateRecheckDelay atomic.Int64
+
+func init() {
+	secretUpdateRecheckDelay.Store(int64(3 * time.Second))
+}
```
jcmoraisjr, line 50: Making this var as a field of `RouteSecretManager{}` would make it simpler? It does not need to be a new constructor param, but changed instead on tests that rely on it just after its creation.

jcmoraisjr (commented):
The latest updates make sense to me. There's just a minor comment, consider it just as a nit, addressable only if it looks valid to you and if you also have other changes to push.

/lgtm

pkg/router/controller/route_secret_manager_test.go
```diff
@@ -1312,6 +1529,106 @@ func TestSecretUpdate(t *testing.T) {
...
+			for time.Now().Before(deadline) {
+				if len(recorder.GetRejections()) == len(s.expectedRejections) {
+					break
+				}
+				time.Sleep(time.Millisecond)
+			}
+
+			if !reflect.DeepEqual(s.expectedRejections, recorder.GetRejections()) {
+				t.Fatalf("expected rejections %v, but got %v", s.expectedRejections, recorder.GetRejections())
+			}
```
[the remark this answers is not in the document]
bentito: going to note this for later work

pkg/router/controller/route_secret_manager.go
bentito: agreed, but I'd rather not push a commit for a minor item and have to run all the e2e again. Will note this for later work

melvinjoseph86:
Marking this as verified based on https://github.com/openshift/router/pull/825#issuecomment-5114216858
/verified by @mjoseph

neisw:
/hold
@bentito let's run some more validation.

https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-router-828-periodics-e2e-aws-ovn-conformance/2084025823903354880
https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-router-828-periodics-e2e-aws-ovn-conformance/2084025827963441152

look like they have failures for `[sig-network][OCPFeatureGate:RouteExternalCertificate][Feature:Router][apigroup:route.openshift.io] with valid setup the router should support external certificate and the secret is deleted then routes are not reachable [Suite:openshift/conformance/parallel]`

I think we saw failures on other jobs as well.  May as well take some time to validate them before merging.

neisw:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4 periodic-ci-openshift-release-main-ci-5.0-e2e-aws-ovn-techpreview

neisw:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

neisw:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

bentito:
good call @neisw

melvinjoseph86:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4 periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10 periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

bentito:
@neisw yeah I thought I had 10/10 clean, but I missed 2 fails just in the 10 I ran. I think I see what happened. I have more work to do. If you could post a list of job triggers you'd like to see run and do well before I take the hold off again (and after an upcoming fix commit) I'll trigger them all before hold cancel. Thanks!

neisw:
Looks like we see failures in
https://prow.ci.openshift.org/view/gs/test-platform-results/logs/aggregator-periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance/2084727478915436544

Expand the Spyglass Suite Information to see the failed test run (click on the red square to open)

image


and
https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-router-828-periodics-e2e-aws-ovn-conformance/2084903396241838080

neisw:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

bentito:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

bentito:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4 periodic-ci-openshift-release-main-ci-5.0-e2e-aws-ovn-techpreview

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 20

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

bentito:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4 periodic-ci-openshift-release-main-ci-5.0-e2e-aws-ovn-techpreview

bentito:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

bentito:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4 periodic-ci-openshift-release-main-ci-5.0-e2e-aws-ovn-techpreview

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

bentito:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4 periodic-ci-openshift-release-main-ci-5.0-e2e-aws-ovn-techpreview

bentito:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

neisw:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

bentito:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4 periodic-ci-openshift-release-main-ci-5.0-e2e-aws-ovn-techpreview

bentito:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

neisw:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

neisw:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4 periodic-ci-openshift-release-main-ci-5.0-e2e-aws-ovn-techpreview

neisw:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

bentito:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

bentito:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4 periodic-ci-openshift-release-main-ci-5.0-e2e-aws-ovn-techpreview

bentito:
@neisw this seems to be doing quite well on the aggregate payload tests now. going to ask for human review and verify again

neisw:
/unhold
I agree @bentito - thanks for the extra effort on this!

pkg/router/controller/route_secret_manager.go
jcmoraisjr: nit: this func is not being used anymore.

pkg/router/controller/route_secret_manager.go
```diff
@@ -46,6 +67,40 @@ type RouteSecretManager struct {
...
+
+// lockRoute acquires the per-route lock for key, creating it on first use,
+// and returns a function to release it.
+func (p *RouteSecretManager) lockRoute(key string) func() {
+	value, _ := p.routeLocks.LoadOrStore(key, &sync.Mutex{})
+	mu := value.(*sync.Mutex)
+	mu.Lock()
+	return mu.Unlock
 }
```
jcmoraisjr, line 105: nit: a key as a `types.NamespacedName{}` avoids a string concat in the hotpath

pkg/router/controller/shared_secret_manager.go
```diff
@@ -0,0 +1,301 @@
...
+	for _, h := range handlers {
+		switch eventType {
+		case "Add":
+			if h.AddFunc != nil {
+				h.AddFunc(obj)
+			}
+		case "Update":
+			if h.UpdateFunc != nil {
+				h.UpdateFunc(oldObj, obj)
+			}
```
jcmoraisjr, line 206: nit: is handling hundreds of routes pointing to the same secret a valid scenario? If so maybe parallelizing the `UpdateFunc()` is a valid followup update?

pkg/router/template/router_test.go
```diff
@@ -1516,6 +1516,88 @@ func Test_configsAreEqual(t *testing.T) {
...
+func TestAddRouteStalenessGuard(t *testing.T) {
+	router := NewFakeTemplateRouter()
+
+	route := &routev1.Route{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "ns",
+			Name:      "extcert-route",
+			Annotations: map[string]string{
+				"router.openshift.io/cert-resource-version": "100",
+			},
```
jcmoraisjr, line 1528: nit: maybe use the `certResourceVersionAnnotation` const?

pkg/router/template/router.go
```diff
@@ -1105,6 +1105,7 @@ func (r *templateRouter) createServiceAliasConfig(route *routev1.Route, backendK
 		ActiveServiceUnits:    activeServiceUnits,
 		HTTPResponseHeaders:   httpResponseHeadersList,
 		HTTPRequestHeaders:    httpRequestHeadersList,
+		CertResourceVersion:   route.Annotations["router.openshift.io/cert-resource-version"],
```
jcmoraisjr: nit: same here, what about reusing `certResourceVersionAnnotation` const.

pkg/router/template/router.go
```diff
@@ -1209,6 +1210,14 @@ func (r *templateRouter) AddRoute(route *routev1.Route) {
 			return
 		}
 
+		if isStaleExtCert(newConfig, &existingConfig) {
+			log.V(4).Info("dropping stale external certificate update",
+				"namespace", route.Namespace, "name", route.Name,
+				"existingVersion", existingConfig.CertResourceVersion,
+				"incomingVersion", newConfig.CertResourceVersion)
+			return
+		}
```
jcmoraisjr, line 1221: What is the chance for this guard being dropping legit changes to routes?

pkg/router/template/router.go
```diff
@@ -1583,6 +1592,23 @@ func configsAreEqual(config1, config2 *ServiceAliasConfig) bool {
...
+	if incoming.CertResourceVersion == "" || existing.CertResourceVersion == "" {
+		return false
+	}
+	inV, errIn := strconv.ParseInt(incoming.CertResourceVersion, 10, 64)
+	exV, errEx := strconv.ParseInt(existing.CertResourceVersion, 10, 64)
+	if errIn != nil || errEx != nil {
+		return false
+	}
+	return inV <= exV
+}
```
jcmoraisjr, line 1614: `resourceVersion` is considered an opaque attribute. TIL it can be [safely compared](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions) when using the same api. Wondering however if using `resourceversion.CompareResourceVersion()` from `"k8s.io/apimachinery/pkg/util/resourceversion"` would be a better approach.

pkg/router/controller/route_secret_manager.go
```diff
@@ -289,43 +431,102 @@ func (p *RouteSecretManager) generateSecretHandler(namespace, routeName string)
...
+			// API-server load (e.g. HyperShift CI).
+			unlock := p.lockRoute(key)
+			if err := p.populateRouteTLSFromSecret(route); err != nil {
+				unlock()
+				return
 			}
+			if err := p.plugin.HandleRoute(watch.Modified, route); err != nil {
+				log.Error(err, "failed to propagate route after secret update", "namespace", namespace, "route", routeName)
+			}
+			unlock()
```
jcmoraisjr: If `HandleRoute()` panics and the caller recovers and don't let the process to crash, the route should continue locked indefinitely, potentially deadlocking the next goroutine trying to acquire a lock from that route. Wondering if it'd be safer to defer unlock inside an inline func().

jcmoraisjr (commented):
Changes look good. I focused on the latest ones after the last review, maybe I'm repeating myself or other previous comments, so sorry if this is happening.

melvinjoseph86:
Hi @bentito i tested this PR with 2000 route, is that much external cert routed required. 
The ingress pods came up
```
➜  oc wait --for=condition=Ready pod \oc logs -n openshift-ingress \
  -l ingresscontroller.operator.openshift.io/deployment-ingresscontroller=default \
  -c router --tail=200
oc wait --for=condition=Ready pod \oc logs -n openshift-ingress -l  -c router  0.07s user 0.02s system 123% cpu 0.076 total
```
Then i tried Secret deletion and updation and also RBAC revocation. All working as expected
Checked  No x509/ECDSA Errors Under Concurrent Secret Updates
```
➜   oc logs -n openshift-ingress \
    -l ingresscontroller.operator.openshift.io/deployment-ingresscontroller=default \
    -c router --tail=500 | grep -iE 'x509|ECDSA|verification failure'
➜    oc logs -n openshift-ingress \
    -l ingresscontroller.operator.openshift.io/deployment-ingresscontroller=default \
    -c router --tail=500 | grep -i panic
➜  
```
Checked Multiple Routes Sharing One Secret (SharedSecretManager) 
```
➜  for i in $(seq 1 5); do
  tatus=$(oc get route shared-route-${i} -n extcert-shared-test \
    -o jsonpath='{.status.ingress[0].conditions[0].status}' 2>/dev/null)
  reason=$(oc get route shared-route-${i} -n extcert-shared-test \
    -o jsonpath='{.status.ingress[0].conditions[0].reason}' 2>/dev/null)
  echo "shared-route-${i}: Admitted=${tatus} Reason=${reason}"
done
shared-route-1: Admitted=False Reason=ExternalCertificateValidationFailed
shared-route-2: Admitted=False Reason=ExternalCertificateValidationFailed
shared-route-3: Admitted=False Reason=ExternalCertificateValidationFailed
shared-route-4: Admitted=False Reason=ExternalCertificateValidationFailed
shared-route-5: Admitted=False Reason=ExternalCertificateValidationFailed
```

Marking as verified
/verified by @mjoseph

pkg/router/controller/route_secret_manager.go
bentito: Agreed, removed in upcoming commit

pkg/router/controller/route_secret_manager.go
bentito: Agreed, fixing in upcoming commit. Renamed `generateKey` → `routeKey` returning `types.NamespacedName`

pkg/router/controller/shared_secret_manager.go
bentito: It's a valid scenario we've tested with 2000 external cert routes (see Melvin's verification comment for instance). The current serial iteration works because each UpdateFunc is fast (one secret read + one HandleRoute call per route).

Parallelizing would need careful coordination with the per-route locks and the single-threaded controller loop's state expectations. Worth considering as a follow-up if profiling shows it's a bottleneck, but the current approach has passed HyperShift conformance aggregates cleanly, which has been tough for this PR as the whole thing is a move to more parallel operation.

pkg/router/template/router_test.go
bentito: Agreed, done in upcoming commit

pkg/router/template/router.go
bentito: agreed, done in upcoming commit

pkg/router/template/router.go
bentito: Chances are zero for non-external-cert routes as they never carry a `CertResourceVersion`, so `isStaleExtCert` returns false immediately and the guard is invisible. For external cert routes, `CertResourceVersion` is the secret's `ResourceVersion` which the API server guarantees to be monotonically increasing per object. A legit secret update always produces a strictly higher value, so it passes the inV <= exV check. The guard only drops deliveries where an older secret read arrives after a newer one has already been applied — exactly the out-of-order delivery scenario it's designed to prevent. Added a clarifying comment in the next commit.

pkg/router/template/router.go
bentito: The `resourceversion.CompareResourceVersion()` utility from `k8s.io/apimachinery/pkg/util/resourceversion` isn't in our vendor tree, and the only `CompareResourceVersion` available (in client-go's mutation_cache) operates on `runtime.Object` and panics on error, different API entirely. Vendoring a new package just for this would be scope creep, IMO. The inline `strconv.ParseInt` comparison does the same thing and fails safe (returns false on parse error, letting the update through). I'd like to leave this one as-is if you agree?

jcmoraisjr: My main concern is about dealing with an opaque attribute while there's an utility for that, which ensures it should continue to work in the future. A while ago resourceVersion should be compared for equality only.

Maybe I'm underestimating the new package addition so I'm fine if you're more comfortable letting it the way it is :+1:

pkg/router/controller/route_secret_manager.go
bentito: It is better code and safer so I've changed it in a further commit (to push soon). Wrapped the lock/unlock in an inline func with deferred unlock so the mutex is always released even if `HandleRoute` or `populateRouteTLSFromSecret` panics.

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

bentito:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

bentito:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4 periodic-ci-openshift-release-main-ci-5.0-e2e-aws-ovn-techpreview

melvinjoseph86:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10 periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10  periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4

melvinjoseph86:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

melvinjoseph86:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

melvinjoseph86:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4

melvinjoseph86:
Again check the PR using clusterbot and hence marking as verified
/verified by @mjoseph

bentito:
/payload-aggregate periodic-ci-openshift-hypershift-release-5.0-periodics-e2e-aws-ovn-conformance 10

bentito:
/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade 10

bentito:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-ipv4 periodic-ci-openshift-release-main-ci-5.0-e2e-aws-ovn-techpreview

PR #838, opened 2026-08-27
NE-2816: Remove HAProxy installation from the router image

jcmoraisjr:
HAProxy now runs on a sidecar container, using a distinct image from the router one. This update is removing the HAProxy installation from the router image. The reload script is also being removed, along with any configuration pointing to it, like envvar in the router image and dev documentation.

The --reload command-line option is still supported, so this does not change the expectation of the nginx plugin. Providing a script however is optional if the HAProxy admin unix socket is provided.

Hacking documentation for running the router locally is also updated.

Jira: https://redhat.atlassian.net/browse/NE-2816


jcmoraisjr:
This update needs to wait at least MicroShift supporting haproxy version selection.

/hold

images/router/haproxy/Dockerfile.ocp
```diff
@@ -18,5 +17,5 @@ USER 1001
 EXPOSE 80 443
 WORKDIR /var/lib/haproxy/conf
 ENV TEMPLATE_FILE=/var/lib/haproxy/conf/haproxy-config.template \
-    RELOAD_SCRIPT=/var/lib/haproxy/reload-haproxy
+    ROUTER_HAPROXY_ADMIN_UNIX_SOCKET=/var/lib/haproxy/run/admin.sock
```
jcmoraisjr, line 20: This is already being configured via operator, but some e2e tests deploy the router manually and expect either the reload script or the haproxy socket already configured. We'll probably need to revisit them, since some of them should be deploying the router as a single container pod.

jcmoraisjr: https://github.com/openshift/origin/pull/31589 configures HAProxy as sidecar in the tests configuring the router pod manually.

jcmoraisjr: Updated https://github.com/openshift/origin/pull/31589 to provide the envvar, this can be removed altogether from the container spec.

pkg/cmd/infra/router/template.go
```diff
@@ -611,12 +613,11 @@ func (o *TemplateRouterOptions) Run(stopCh <-chan struct{}) error {
...
 		}
 
-		adminUnixSocket := os.Getenv("ROUTER_HAPROXY_ADMIN_UNIX_SOCKET")
-		hasHAProxySidecar := len(adminUnixSocket) > 0
+		hasHAProxySidecar := len(o.HAProxyAdminUnixSocket) > 0
 
 		pidFn := haproxyPidEmbedded
 		if hasHAProxySidecar {
-			pidFn = haproxyPidSidecar(ctx, adminUnixSocket)
+			pidFn = haproxyPidSidecar(ctx, o.HAProxyAdminUnixSocket)
```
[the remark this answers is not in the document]
jcmoraisjr: I'm wondering if I understood you correctly. Just to confirm: there are two distinct sockets - one of them comes from the current haproxy instance and we use it to configure HAProxy via API, the other one comes from the master process and we use it to issue a reload command.

hack/local-router.sh
```diff
@@ -28,19 +28,24 @@ router_prepare() {
...
+    echo "$haproxydir prepared, starting HAProxy in foreground -- ^C to stop"
+
+    # starting the "sidecar" haproxy
+    haproxy -W -db -S "${haproxydir}/run/admin.sock,mode,600" -f "${haproxydir}/conf/haproxy.config"
 }
 
 router_run() {
     # changing listening ports, we usually don't have permission to bind to the default ones 80/443
-    ROUTER_SERVICE_HTTP_PORT=9090 ROUTER_SERVICE_HTTPS_PORT=9443 STATS_USERNAME=admin STATS_PASSWORD=admin STATS_PORT=1936 \
+    ROUTER_SERVICE_HTTP_PORT=9090 ROUTER_SERVICE_HTTPS_PORT=9443 STATS_USERNAME=admin STATS_PASSWORD=admin STATS_PORT=1936 ROUTER_GRACEFUL_SHUTDOWN_DELAY=2s \
```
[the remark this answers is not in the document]
jcmoraisjr: This is a script to run the router locally for debugging purpose. Having an easy/insecure password being easily configured is part of the plan - we can clearly see it and can easily remember its value.

hack/Dockerfile.debug
davidesalerno, line 9: Shouldn't we remove this line?

hack/local-router.sh
```diff
@@ -28,19 +28,24 @@ router_prepare() {
...
 
-    echo "$haproxydir prepared"
+    echo
+    echo "$haproxydir prepared, starting HAProxy in foreground -- ^C to stop"
+
+    # starting the "sidecar" haproxy
+    haproxy -W -db -S "${haproxydir}/run/admin.sock,mode,600" -f "${haproxydir}/conf/haproxy.config"
 }
 
 router_run() {
```
davidesalerno, line 43: We could consider making the new two-step developer workflow even harder to misuse by detecting a missing `/var/lib/haproxy/run/admin.sock` in this function and failing with a message like “run make local-prepare in another terminal first.”

hack/Dockerfile.debug
gcs278: Yea, when I tried `make -f hack/Makefile.debug` I got:

```
26 files removed
Failed to set capabilities on file `/usr/sbin/haproxy' (No such file or directory)
```

hack/local-router.sh
```diff
@@ -28,19 +28,24 @@ router_prepare() {
         sudo mkdir -p "$haproxydir"
         sudo chown "${USER}:${USER}" "$haproxydir"
     fi
-    rm -rf "${haproxydir}/{conf,run}"
-    mkdir -p "${haproxydir}/{conf,run}"
+    rm -rf "${haproxydir}/conf" "${haproxydir}/run"
+    mkdir -p "${haproxydir}/conf" "${haproxydir}/run" "${haproxydir}/router/certs" "${haproxydir}/router/cacerts" "${haproxydir}/router/allowlists"
     cp images/router/haproxy/conf/* "${haproxydir}/conf/"
+    touch "${haproxydir}/conf/haproxy.config"
```
gcs278, line 34: i'm a little confused here, will haproxy fail with a blank `haproxy.config` file?

images/router/haproxy/Dockerfile
gcs278, line 1: *nit* can we rename `images/router/haproxy/...` to `images/router/router`? Having `haproxy`, `haproxy28`, `haproxy32` directories is pretty confusing.

But I realize you likely need to make other ocp-build-type changes, so will likely need a multi-PR transition (add new directory --> Update build steps --> remove old directory).

Something we can follow up later on, not a requirement.

Edit: Ah but I guess we can't change the name of the router image easily? It's still [`ose-haproxy-router`](https://github.com/openshift-eng/ocp-build-data/blob/openshift-5.1/images/openshift-enterprise-haproxy-router.yml). Hmm. Something to consider

images/router/haproxy/Dockerfile.ocp
```diff
@@ -18,5 +17,5 @@ USER 1001
 EXPOSE 80 443
```
gcs278, line 17: should this also be removed? router is not passing traffic on 80/443 anymore, haproxy is.

images/router/haproxy/Dockerfile.ocp
```diff
@@ -1,12 +1,11 @@
 FROM registry.ci.openshift.org/ocp/5.0:haproxy-router-base
-RUN INSTALL_PKGS="socat haproxy32 rsyslog procps-ng util-linux" && \
+RUN INSTALL_PKGS="socat rsyslog procps-ng util-linux" && \
```
gcs278, line 2: I haven't dug deeply into any of these, but may be worth examining if each one is still required in router container. For example, `rsyslog` may no longer be required - I'm not sure if the router container needs it, it's also in the haproxy container (full disclosure: I'm not in tune with what syslogging features we offer, it may be needed still).

gcs278 (approved):
I took a very quick look from a "secondary reviewer" standpoint. Approach Looks good.

Seems like both microshift & hypershift have transitioned to the sidecar model if I am looking at the right things. 

I will apply an initial approve and @davidesalerno can follow up with the more detailed review and the LGTM.

PR #839, opened 2026-08-27
Add retry-on and retries route annotations

bobbymehra:
## Summary
- Expose HAProxy `retry-on` and `retries` as per-route annotations so intermittent backend failures (connection resets, HTTP 503, empty responses) can be retried in the router instead of by every client.
- `haproxy.router.openshift.io/retry-on` maps to HAProxy `retry-on` on HTTP backends (edge/reencrypt/cleartext). Invalid values are ignored. Passthrough routes ignore `retry-on`.
- `haproxy.router.openshift.io/retries` maps to HAProxy `retries` (0-99) on HTTP and passthrough backends. Optional router-wide defaults: `ROUTER_DEFAULT_RETRY_ON` and `ROUTER_DEFAULT_RETRIES`.

Fixes: https://issues.redhat.com/browse/RFE-3876

## Test plan
- [ ] `go test ./pkg/router -run TestConfigTemplate`
- [ ] Annotate an edge route with `haproxy.router.openshift.io/retry-on=conn-failure 503` and `haproxy.router.openshift.io/retries=3`; confirm `retry-on` and `retries` appear in the route's HAProxy backend
- [ ] Confirm an invalid `retry-on` value (e.g. `conn-failure; bind :80`) is not written to `haproxy.config`
- [ ] Confirm `retry-on` is ignored on passthrough routes while `retries` still applies
- [ ] Confirm the same annotations on a Kubernetes Ingress are copied to the generated Route and take effect

Made with [Cursor](https://cursor.com)


gcs278:
@bobbymehra thanks for the contribution!

There are a couple things about this PR:

- This PR is adding an annotation, and while we support existing annotations, we do all new configurations as proper API fields. Annotations are not validated, not well documented/discoverable, and lack other benefits of an API.
- We are unable to accept new features tied directly to RFEs without an enhancement proposal. This [doc](https://github.com/openshift/enhancements/blob/master/dev-guide/feature-zero-to-hero.md) is a good place to start with the process.

My suggestion - if you'd like to drive this forward, please work with our team's stakeholders to get a feature created and open an enhancement PR.

